PR #73, opened 2026-07-13
#0 feat: suggest a task for idle agents via local LLM

0xGosu:
## Summary

When an agent goes **idle** with **no task source** (no declared `[[task_sources]]` matches it and nothing is inferable from the pane) and the operator has configured the new `llm.generate_task_command`, hap invokes a **one-shot local LLM** that **suggests a task**. The suggestion surfaces as a normal **escalation**:

- **confirm+send** (`enter`/`y`) → hap writes a per-agent `tasks.md` (single in-progress `[-]` item) under the state dir, registers it as a `[[task_sources]]` entry, and sends the task to the agent;
- **dismiss** (`x`) → nothing happens;
- if generation **failed**, the escalation shows the failure rationale and the operator **retries with `l`** (same key as the existing LLM-consult retry).

This relaxes the FR-011 *"never synthesize a prompt without a task source"* invariant **only behind the new opt-in command** — unconfigured idle still escalates `no_task_source` and synthesizes nothing.

Once `tasks.md` exists, the **normal declared-task flow** takes over on later idles (the completed-list template steers the agent), so hap never re-generates for that agent.

## Design notes

- The generate-task CLI is a **one-shot whose stdout is the task text** — mirrors `llm.rewrite_command`, not the `submit_decision` MCP flow (this path always escalates, never auto-acts, so no safety re-gate is needed).
- **No tasklist parser change:** `[-]` already parses as done/checked, so a single-`[-]` file resolves via the existing "completed sentinel" path.
- Safety gates (kill switch, never-auto, variance guard, rate limit, irreversible heuristic) all precede the generation branch in `Decide`.
- The async generation runs off the main loop; a pending `llm_requests` row guards against concurrent generations from bursty idle events and lets `l`-retry reuse the in-flight check.

## Config

```toml
[llm]
generate_task_command = ["claude", "-p", "Suggest ONE concrete next task ...\n\nAgent: {agent_name}\nCwd: {cwd}\n\nScreen:\n{pane_excerpt}", "--model", "haiku"]
# generate_task_command_start = [ ... ]   # optional first-generation variant
# generate_task_timeout_seconds = 60      # omitted: inherits timeout_seconds
```

Placeholders: `{self}`, `{agent_name}`, `{agent_type}`, `{pane_excerpt}`, `{cwd}`.

## Files

| Area | Change |
|---|---|
| `internal/config` | new `[llm]` keys + `GenerateTaskTimeout()` |
| `internal/domain` | `ActionGenerateTask`, `ReasonTaskGenFailed`, `SuggestGenerateTask`/`SuggestTaskPrefix`, `TaskGenRequest`, idle relaxation, retryability |
| `internal/ports` | `TaskGeneratorPort` |
| `internal/llm` | `taskgen.go` one-shot subprocess (mirrors `rewrite.go`) |
| `cmd/hap` | wire adapter fields |
| `internal/daemon` | dispatch, `generateTask`, `handleTaskGenOutcome`, idle retry re-drive, stale-cutoff = max(consult, gentask) timeout |
| `internal/frontend` | `SuggestedAction` prefix, `Resolve` branch, `acceptGeneratedTask` |
| `sample/config.toml` | documented the new keys |

## Testing

- Full unit suite: **675 tests, 0 failures** (`go test -tags "vectors cpu" ./...`); build + `go vet` + `gofmt` clean.
- New coverage: config parse/timeout, decide idle relaxation + retryability, one-shot adapter (success/empty/timeout/preflight/start-template), daemon end-to-end (generate→escalation, failure→retryable, retry re-drive), frontend confirm (writes file + task source + correction + send).
- Code review (subagent): rated Excellent; the one minor finding (stale-cutoff ignoring a long `generate_task_timeout_seconds`) is fixed in `expireStaleLLMWork`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmd/hap/main.go
+++ b/cmd/hap/main.go
@@ -253,6 +253,9 @@ func runDaemon(ctx context.Context, paths config.Paths, args []string) error {
 			RewriteTemplate:      cfg.LLM.RewriteCommand,
 			RewriteStartTemplate: cfg.LLM.RewriteCommandStart,
 			RewriteTimeout:       cfg.RewriteTimeout(),
+			TaskGenTemplate:      cfg.LLM.GenerateTaskCommand,
+			TaskGenStartTemplate: cfg.LLM.GenerateTaskCommandStart,
+			TaskGenTimeout:       cfg.GenerateTaskTimeout(),
 		}
 	}
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -115,6 +115,23 @@ type LLM struct {
 	// fails (placeholders {original_text}, {agent_name}). Empty uses the
 	// built-in default; a rewrite failure never blocks the send.
 	RewriteFallbackTemplate string `toml:"rewrite_fallback_template"`
+	// GenerateTaskCommand is the argv template for the one-shot task
+	// suggestion an idle agent gets when it has NO task source (no declared
+	// [[task_sources]] and nothing inferable from the pane). Placeholders:
+	// {self}, {agent_name}, {agent_type}, {pane_excerpt}, {cwd}. The suggested
+	// task is read from the CLI's stdout and surfaced as an escalation the
+	// operator confirms (writing a per-agent tasks.md) or dismisses. Empty
+	// keeps today's behavior: idle with no task source escalates as
+	// no_task_source and the plugin never synthesizes a prompt (FR-011).
+	GenerateTaskCommand []string `toml:"generate_task_command"`
+	// GenerateTaskCommandStart is the argv template used on the FIRST task
+	// generation per agent (same first-interaction boundary as CommandStart).
+	// Same placeholders as GenerateTaskCommand. Empty inherits
+	// GenerateTaskCommand; tracked independently of the consult "first".
+	GenerateTaskCommandStart []string `toml:"generate_task_command_start,omitempty"`
+	// GenerateTaskTimeoutSeconds bounds one task-generation run; zero or
+	// omitted inherits timeout_seconds.
+	GenerateTaskTimeoutSeconds int `toml:"generate_task_timeout_seconds,omitempty"`
 }
 
 // Embedding configures semantic signature matching: situations are matched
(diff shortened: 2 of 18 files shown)


0xGosu:
Thanks for the review — all three findings are addressed (also rebased onto latest `main`, resolving the `applyLLMRetry` conflict by adopting main's live-transition re-drive, which carries the idle task-gen retry correctly).

**P1 — idempotent confirmation.** `acceptGeneratedTask` now atomically CLAIMs the escalation via a new `Store.ResolveEscalation(id) (bool, error)` (`UPDATE … WHERE status='escalated'`); only the claimer proceeds. The idempotent writes (file + `addTaskSourceIfAbsent`, which de-dupes on agent+path under `UpdateConfig`'s advisory lock) run **before** the claim so a failure there leaves the escalation pending/retriable, and only the non-idempotent send is gated by the claim. The correction is now best-effort. A double-submit can't re-send or stack duplicate sources.

**P2 — staleness.** A suggestion is now invalidated against the current idle state both **at surfacing** (daemon `handleTaskGenOutcome` drops it if the agent's live herdr status is no longer cleanly idle — `domain.AgentBusy` via `ListAgents` — or a task source now matches) and **at accepting** (frontend confirm refuses when the agent is no longer idle). Both fail **open** on unknown status (list error / agent absent).

**P3 — multiple tasks + safe format.** The LLM may now return multiple tasks (plain or Markdown). `domain.NormalizeGeneratedTasks` strips any bullet/number/checkbox/**code-fence** markup and drops punctuation-only lines (one clean task per line — can't inject a second checklist item); `RenderGeneratedTaskList` writes the first task in-progress `[-]` and the rest pending `[ ]`, and only the **first** task is sent to the agent.

New test coverage: idempotent double-confirm, working-agent refusal (escalation stays pending), multi-task checklist + single send, code-fence/punctuation normalization, and the daemon stale-drop. Full suite green (`go test -tags "vectors cpu" ./...`).

PR #89, opened 2026-07-13
feat: add live-event dedup to prevent duplicate escalations

0xGosu:
## 📌 Background

The daemon's escalation pipeline can receive duplicate events — the same situation (same agent, type, and pane content) arriving while an escalation is already pending user action. Currently, these duplicates are re-escalated, creating redundant asks and potential escalation storms from async rejection paths (LLM, rewrite, taskgen).

## 🔧 Reason for Changes

Implement content-level deduplication at the escalation choke point to suppress duplicate escalations while preserving legitimate new situations on the same agent (which may have different pane content). This prevents:
- Redundant user prompts for identical situations
- Escalation storms from async rejection paths
- Noise in the audit log

The dedup is content-level (agent + type + pane excerpt), not agent-level, so a genuinely new situation on a pane that already has an open escalation still escalates.

## ✅ Changes Implemented

- **`internal/store/store.go`**: Added `DuplicatePendingEscalation()` method to efficiently check if an unresolved escalation exists for the same agent + type + situation type + pane excerpt (uses `EXISTS` query, never materializes rows).
- **`internal/ports/ports.go`**: Added `DuplicatePendingEscalation()` to the `ReadStore` interface.
- **`internal/daemon/daemon.go`**: 
  - Added `duplicatePendingEscalation()` helper that wraps the store call and fails open (returns false) on error per the "never a silent drop" architecture rule.
  - Added dedup gate in `escalate()` before any escalation operations; if a duplicate is detected, the event is audited with rationale "duplicated event" and returns early.
- **`internal/daemon/daemon_test.go`**: 
  - Added `otherApprovalPane` constant for testing content-level dedup.
  - Added `ignoredRows()` helper to fetch audit records marked as ignored.
  - Added `TestPipelineIgnoresDuplicateEvent()` covering the full dedup flow: identical events are dropped with audit trail, genuinely new situations on the same agent still escalate.
- **`internal/store/store_test.go`**: Added `TestDuplicatePendingEscalation()` covering the store-level dedup query with field-level mutation tests.

## 🔍 Testing Results

- ✅ Unit tests added and passing:
  - `TestDuplicatePendingEscalation` validates the store query with exact-match and field-mutation cases
  - `TestPipelineIgnoresDuplicateEvent` validates the full daemon pipeline: duplicate suppression, audit trail, and content-level (not agent-level) dedup semantics
- ✅ Dedup gate is the single choke point in `escalate()`, so it caps storms from all rejection paths (LLM, rewrite, taskgen)
- ✅ Fails open on store errors (returns false, allowing event to proceed) per architecture rule

No database migrations or environment changes required. The dedup uses existing `audit_log` table schema (status, agent_id, agent_type, situation_type, pane_excerpt columns).

---

### 📝 Additional Notes

The dedup is gated at `escalate()` rather than earlier in the pipeline so that situations capable of auto-answering (e.g., after a kill-switch resume) still act even if a duplicate escalation exists. Only would-be escalations are suppressed.

https://claude.ai/code/session_01Ry7t9v5rExaDtfEAnkAXMP

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -679,6 +679,28 @@ func (d *Daemon) hasOpenEscalation(ctx context.Context, agentID string) bool {
 	return false
 }
 
+// duplicatePendingEscalation reports whether the captured situation exactly
+// matches an escalation the user has not yet handled. It is the content-level
+// sibling of hasOpenEscalation (which is pane/agent-level and used by the
+// reconcile sweep): finer, so a genuinely new situation on a pane that already
+// has an open escalation still gets processed. The pane excerpt is truncated
+// exactly as escalate() stores it so the comparison lines up.
+//
+// Fails OPEN (returns false) on a store error — the opposite of
+// hasOpenEscalation. Dropping a real event silently is worse than
+// re-processing one, so on doubt the event proceeds to escalate/act rather
+// than being ignored (the "never a silent drop" architecture rule).
+func (d *Daemon) duplicatePendingEscalation(ctx context.Context, s domain.Situation) bool {
+	dup, err := d.opt.Store.DuplicatePendingEscalation(ctx, s.AgentID, s.AgentType,
+		s.Type, truncateRunes(s.Content, snapshotMaxRunes))
+	if err != nil {
+		slog.Warn("duplicate-escalation check failed; processing event",
+			"agent", s.AgentID, "pane", s.PaneID, "error", err)
+		return false
+	}
+	return dup
+}
+
 // handleAttention is the post-delay half of the pipeline: the classification
 // pane read and everything after it. It runs on the main loop (via
 // delayedTr) and re-derives its inputs at fire time, so every gate — kill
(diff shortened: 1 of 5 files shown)


0xGosu:
Thanks — the blocking point is correct and now fixed in `9b9bc67`.

**Blocking (fixed):** The pane-read-failure path in `handleAttention` wrote its `escalated` audit row directly, bypassing `escalate()` and its dedup gate, so a persistent Herdr/read outage could pile up one identical pending escalation per event. That path now runs the same dedup inline before writing. Since the pane is unreadable, the key collapses to `agent_id + agent_type + unclassifiable` (empty excerpt), which exactly matches the row the path writes. Extracted a shared `ignoreDuplicate` audit helper so both routes record duplicates identically, and added `TestPaneReadFailureDedupsDuplicateEvents` as a regression test. Reworded the `escalate()` comment that overclaimed "single choke point."

**Non-blocking (index):** Skipping for now. `pending` escalations are operator-facing queue items that get resolved/dismissed, so the `status='escalated'` set the existing `idx_audit_status` narrows to is small (bounded well under the 200-row `PendingEscalations` cap in practice); the `EXISTS` residual excerpt comparison runs over a handful of rows. A composite index would be premature until we observe pending rows actually accumulating. Happy to add it if you'd prefer it pre-emptively.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #92, opened 2026-07-13
feat: add pre-send LLM review gate for declared tasks

0xGosu:
## 📌 Background

This change implements a safety gate for declared tasks: when an LLM command is configured, each task determined from a `[[task_sources]]` entry for an idle agent is now reviewed by the LLM before being sent to the agent. This allows the operator's LLM to validate that a queued task is appropriate given the live pane state, pick a different pending task if the current one is already done, or decline the task entirely if the agent is still busy.

## 🔧 Reason for Changes

Previously, declared tasks were sent directly to idle agents without LLM review, even when an LLM command was configured. This meant:
- Tasks could be sent when the agent was actually still working (false idle detection)
- Tasks already completed on the pane would still be sent
- No opportunity for the LLM to adapt the task list based on live context

The pre-send review gate closes this gap by consulting the LLM with the proposed task, the full pending task list, and the live pane context. The LLM can then:
1. **Approve** (recommend_action with the task) — sent if confident, escalated if low-confidence
2. **Decline** (@noop) — escalated if low-confidence, silently skipped if confident
3. **Adapt** (recommend_action with a different pending task) — when the current task is already done

## ✅ Changes Implemented

- **New test suite** (`internal/daemon/taskreview_test.go`): comprehensive coverage of the pre-send review flow, including approval, decline (sub-threshold and confident), duplicate-event suppression, and opt-out behavior.

- **Core review logic** (`internal/daemon/daemon.go`):
  - Added `consultDeclaredTask()` to run the LLM review in a goroutine (same pattern as `consultLLM`)
  - Added `taskReviewContext` struct to carry review-specific context fields: proposed task, task list path, current task, and pending tasks
  - Extended `consultContext()` to accept an optional `taskReviewContext` parameter and populate review-specific fields in the JSON context
  - Added `paneCwds` cache to avoid repeated `pane get` calls for the `{cwd}` placeholder resolution
  - Integrated review gate into `decideAndAct()`: when idle with a determined declared task, LLM configured, and no safety escalations pending, consult the LLM before sending

- **Domain types** (`internal/domain/tasklist.go`, `internal/domain/types.go`):
  - Extended `DeclaredTask` with `Cwd` field (for `{cwd}` template substitution) and `LLMReview` flag (opt-out per source)
  - Added `PendingDeclaredTasks()` function to extract all unchecked items from a task file (used by review context)
  - Updated `Prompt()` to substitute `{cwd}` placeholder

- **Config & frontend** (`internal/config/config.go`, `internal/frontend/frontend.go`):
  - Added `llm_review` field to `TaskSource` (opt-out: `nil` = on, explicit `false` = off)
  - Added test coverage for `llm_review` parsing

- **LLM outcome handling** (`internal/daemon/daemon.go`):
  - Updated `handleLLMOutcome()` to skip the variance guard for task reviews (each task is novel, not a repeated answer)
  - Enhanced escalation rationale to include the LLM's reasoning and the original proposed task for detail view

- **Documentation & config**:
  - Updated README with "Reviewing tasks before they are sent" section explaining the feature
  - Updated sample config with task-review examples
  - Renamed `generate_task_command*` keys to `task_generate_command*` for consistency (old keys no longer load)
  - Updated all references and comments throughout the codebase

## 🔍 Testing Results

- **Unit tests passed**: New `taskreview_test.go` covers:
  - `TestIdleDeclaredTaskCwdTemplate`: `{cwd}` placeholder substitution in templates
  - `Test

https://claude.ai/code/session_011KomYzGYgT6L4DxPXQnZZG


0xGosu:
All four review findings are addressed in `0e0bbf1` (declared-task review path):

1. **Racy in-flight guard** — the review's `llm_requests` row is now staged **synchronously** before the goroutine (mirroring `generateTask`), so two idle events can't both pass the `HasPendingLLMConsult` guard. The `context_json` (which needs off-loop pane reads) is filled afterward via a new `UpdateLLMRequestContext` store method, before the CLI reads `get_context`.
2. **Silent drop on persistence error** — the `HasPendingLLMConsult` error (and the synchronous-staging error) now route through `d.escalate(ReasonPersistenceFailed)` instead of log-and-return.
3. **Main-loop stall on `pane get`** — `paneCwd` is now non-blocking: it returns the cached (or empty) value immediately and refreshes off the main loop, so a `{cwd}` source can never pause the daemon loop. The first render for a cold pane is empty and self-heals once the refresh lands (the review path still gets the real cwd via `get_context`).
4. **Stale task after a long review** — the request now carries the source path + reviewed task; after the pane staleness check, `handleLLMOutcome` re-reads the source and escalates ("task list changed during review") instead of injecting the task if its next item changed during the review.

New tests cover the pending-check-error escalation, the source-changed escalation, and the async cwd resolution. Full suite + `gofmt`/`go vet`/`golangci-lint` all pass locally.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #95, opened 2026-07-13
Multi-tab MCQ sweep fails on Submit tab — Submit screen has no nav footer, MultiTabForm rejects it


0xGosu:
## Summary

The multi-tab MCQ sweep (`internal/daemon/sweep.go`) reliably fails on the **final Submit tab** and escalates with:

```
[herdr_unreachable] sweep failed: tab 3/3 no longer shows the 3-tab form; partial capture, answer in pane
```

Root cause: Claude Code's **Submit confirmation screen has no navigation footer**, and `domain.MultiTabForm` (`internal/domain/mcqform.go:54`) *requires* a footer. So when the sweep arrows onto the last tab, the same-form re-check at `sweep.go:134` returns `false`, the whole sweep is marked degraded, and the form escalates to the operator instead of being auto-answered. Every multi-tab AskUserQuestion form hits this — it is not intermittent.

This is **not** a duplicate of #50 (that was footer *wording* drift on the question tabs). Here the question tabs match fine; the Submit tab has **no footer at all**.

## Environment

- hap `v0.3.38` (daemon also observed at `v0.3.37`)
- Claude Code `2.1.207`
- herdr 0.7 (running)

## Symptom (from the live instance)

`hap audit`:
```
#74	07-13 16:54:23	escalated	choice	escalated	conf=0.00	rule=-	[herdr_unreachable] sweep failed: tab 3/3 no longer shows the 3-tab form; partial capture, answer in pane
```

Daemon log (`herd-auto-prompter.log`) — recurring:
```
{"time":"2026-07-13T03:12:16Z","level":"WARN","msg":"multi-tab sweep failed; degrading to single-frame capture","agent":"w3:p1","error":"tab 3/3 no longer shows the 3-tab form"}
{"time":"2026-07-13T16:54:23Z","level":"WARN","msg":"multi-tab sweep failed; degrading to single-frame capture","agent":"w3:p1","error":"tab 3/3 no longer shows the 3-tab form"}
```

## Captured evidence (`herdr pane read w3:p1 --source visible`)

This is a real 3-tab form (2 questions + Submit). Tabs 1 and 2 render the header **and** footer, so they pass. The Submit tab does not.

**Tab 1/3 — `☐ Agent identity` (passes):**
```
←  ☐ Agent identity  ☐ Stats to show  ✔ Submit  →

hap identifies each agent by its herdr pane id ...

❯ 1. Yes, new pane id
  2. Pane ids can be reused
[...]

PR #103, opened 2026-07-14
#95 feat: multi-select questions in Claude multi-tab MCQ forms (+ bundled concurrent work)

0xGosu:
## Summary

Adds support for **multi-select questions** in Claude's multi-tab `AskUserQuestion` MCQ forms, and bundles concurrent in-flight work from other agents (TUI/config/docs, LLM task-generation frontend) so it all ships together.

### The problem (multi-select)
A Claude multi-tab form can render a **multi-select** question: its options show per-option `[ ]` checkboxes, and the UX is *toggle several options with digit keys, then press Tab/Right to advance* — a multi-select tab does **not** auto-advance on a digit press. hap treated every tab as single-select, so the LLM could only pick one option and the digit-per-tab delivery could neither express nor type multiple toggles. Observed live on a real plan-mode form (question 2, "Stats to show").

### The fix
- **Detect** per-tab multi-select from the `[ ]` markers (`domain.MultiSelectTab`, `OptionCheckStates`); carried on `Situation.TabMultiSelect`, captured during the sweep.
- **Answer format**: MCP `submit_decision.select_options` items are now **int OR array-of-int** — `[1, [1,3], 2]` (tab 2 toggles options 1 and 3). Fully backward compatible with the flat `[1,2,1]` shape. Internally a per-tab token series with comma-grouped toggles (`"1 1,3 2"`), parsed by `ParseTabSelections`; the `len == TabCount` guards still hold (one token per tab).
- **Delivery**: a pure `MultiTabKeys` planner emits the toggle digits plus an explicit advance for multi-select tabs (single-select auto-advances). Shared by the daemon (`sendTabSelections`) and the operator-confirm frontend.
- **Consult context** advertises `tab_select_kinds` so the LLM knows which tabs take an array.
- **Full parity**: learned rules, LLM, and operator paths all deliver multi-select, re-gated by the existing safety controls.

### Safety (toggling is relative)
Toggling a checkbox is a relative flip, so it's only safe from an all-unchecked baseline:
- The sweep **refuses** a multi-select tab that already has a selection (escalates instead of blind-toggling).
- The daemon **re-verifies** the baseline immediately before autonomous delivery (`reverifyMultiSelect` re-runs the capture sweep as a guard), closing the window where an operator toggles a *middle* tab during a long LLM consult (the tab-1 staleness check can't see middle tabs).
- The operator-confirm path is **all-or-nothing**: a read-only baseline pre-pass over every tab runs before any answer keystroke, so a refusal never leaves the form half-answered.
- The advance key + reset count are unified into `domain.MCQAdvanceKey` / `MCQResetKeys`, referenced by both delivery paths so they can't diverge.

## Commits
- `#95 feat` — the multi-select feature (domain, daemon, MCP, frontend).
- `#95 fix` — code-review remediation (delivery-time re-verify, all-or-nothing operator path, shared constants).
- `#1 chore` / `#1 feat` — **bundled concurrent-agent work**: LLM task-generation frontend rendering (`DeclaredTask.Prompt`), TUI content-height config (`config.TUI.MaxContentHeight`) + list rendering, `enhance-tui-ux` plan docs, README/CLAUDE/sample-config/skill docs. Included per operator request so everything ships in one PR.

## Testing
- New unit tests: `MultiSelectTab`, `OptionCheckStates`, `ParseTabSelections`, `MultiTabKeys`, extended `ParseDigitSeries`; MCP mixed int-or-array resolution; daemon multi-select delivery keystroke assertion + a toggle-state-stale (pre-checked) escalation test; frontend all-or-nothing delivery.
- **Full suite green**: `go test -tags "vectors cpu" ./... -count=1` — all 18 test packages pass (including the bundled concurrent work). `gofmt` + `go vet` clean.

## Follow-up
- The real advance key (`"right"` vs `"tab"`) and the MCP `oneOf` schema shape should be confirmed against a live Claude form via the gated integration test (`HAP_ITEST_CLAUDE=1`); the advance key is const-ized for an easy switch.

### DIFF
--- a/.claude/skills/hap-development-local/SKILL.md
+++ b/.claude/skills/hap-development-local/SKILL.md
@@ -1,0 +1,238 @@
+---
+name: hap-development-local
+description: "Develop and debug the Herd Auto Prompter (hap) plugin against a live herdr — link the working tree, rebuild the binary, and hot-swap the daemon so code changes take effect. Use when iterating on hap source inside a running herdr (HERDR_ENV=1) and you want the local build (not the installed release) to run."
+---
+
+# hap-development-local — local dev loop for the hap plugin
+
+use this skill when you are **editing hap's Go source** and want the running
+herdr to exercise your **working-tree build** instead of the installed release.
+it is the developer counterpart to the `hap` skill (which *operates* an
+already-installed plugin) and the `herdr` skill (which drives herdr itself).
+
+before using this skill, confirm you are inside herdr (`HERDR_ENV=1`) and in the
+repo (`/workspaces/herdr-auto-pilot`, or wherever the working tree lives). if
+`HERDR_ENV` is not `1`, there is no live herdr to link against — say so and stop.
+
+## the mental model (read this first)
+
+herdr can run a plugin from a **local path** instead of a GitHub install. once
+linked, herdr invokes the plugin's binary as **`./bin/hap` relative to the
+plugin root** for every manifest command (daemon, tui, event hooks). so the
+single source of truth for what herdr runs is the file at
+`<plugin_root>/bin/hap`. rebuild that file, restart the daemon, done.
+
+three moving parts, keep them straight:
+
+- **`<plugin_root>/bin/hap`** — what herdr actually executes. Rebuild this.
+- **the running daemon** — a long-lived process spawned from whatever binary
+  last ran `hap daemon --ensure`. It does *not* pick up a new binary on its
+  own; you must re-run `--ensure` to make it self-replace.
+- **`/usr/local/bin/hap` (PATH symlink)** — a convenience symlink for *your*
+  shell. **`herdr plugin link` does NOT repoint it** — see the gotcha below.
+
+a local build reports version **`dev`** (`internal/buildinfo.Version`, stamped
+only by release ldflags). that string is how you tell local-vs-release apart.
+
+## one-time setup: link the working tree
+
+```bash
+herdr plugin link /workspaces/herdr-auto-pilot
+```
+
+- switches the plugin from `github:…@<sha>` to `local:/workspaces/herdr-auto-pilot`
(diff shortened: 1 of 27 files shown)


0xGosu:
Thanks @charliecreates — both blockers addressed in 9b90ba0, plus the lint failure and a bug the live verification surfaced.

**1. MCP resolver accepting arrays on single-select/Submit tabs** (`mcpserver.go`)
The daemon already emits per-tab `tab_select_kinds` ("single"/"multi") in the consult context; the resolver now reads it (`consultContextFields.TabSelectKinds`) and rejects any entry with more than one number on a non-`multi` tab (Submit included) via `tabIsMulti`. A scalar/single tab or Submit takes exactly one option, so the "extra digits with no advance shift onto later tabs" case can't be submitted. New test case covers an array on a single-select tab → rejected.

**2. Re-verification only checking for an error** (`sweep.go` `reverifyMultiSelect`)
It now captures the re-swept situation and fails **closed** unless the re-swept **aggregate content** *and* **per-tab select kinds** both equal the situation being delivered — not just tab-count/unchecked-baseline. A same-count replacement or a changed middle tab changes the aggregate content and is refused before any keystroke. New `TestReverifyMultiSelectRejectsChangedForm` proves a same-count, still-unchecked, changed-middle-tab form is rejected.

**Lint:** removed the now-unused `sweptSituation` helper (the `unused` failure).

**Bonus — found via live verification against a real blocked agent's multi-select form:** an *answered* tab renders `☒` (not `☐`), which `mcqTabEntryRE` didn't count, so a partially-answered 3-tab form read as **2 tabs**. Added `☒` to the tab-entry/header regexes; the live frame now reports `tabs=3`, `MultiSelectTab=true`, and the delivery plan `[1 1 3 right 1]` — all correct.

Full suite green (`go test -tags "vectors cpu" ./...`), gofmt clean.

PR #136, opened 2026-07-16
TUI Tasks tab: item CRUD, manual send, multi-line tasks, and guarded task-source retirement

0xGosu:
Builds out the TUI's task management so the operator can run a herd's checklists without leaving the pane, and tightens the safety rules around task sources. 8 commits, 37 files.

## What's in it

**A new Tasks tab** aggregating every configured `[[task_sources]]` entry and its checklist items into one searchable list — a header per source (with the live agent it feeds) and its items underneath. Full item CRUD in place: `a` add, `e` edit, `d` done/undone, `x` delete, `space` to mark a run for bulk actions, `f` to focus the agent in herdr, `v` for a task detail view.

**Manual task send** — `enter`/`y` sends the pending `[ ]` task under the cursor to the cleanly idle agent its source feeds, rendered through the source's next-task template, behind a `y/n` confirmation, marking it `[-]` on success so the daemon's own idle flow stays in step. CLI twin: `hap task <agent> send <n> [--yes]`.

**Multi-line task text** — Shift+Enter inserts a line break (Ctrl+J on terminals that can't report it) and the input box expands; Enter submits. A task always stays one checklist line: breaks are stored as the literal `\n` escape and decoded when sent.

**Retiring a task source** — `x` on a source's header row removes its config entry behind a confirmation, but only once the source can no longer be serving anyone: no live agent matches it, or every task in it is finished. The checklist file is deliberately kept on disk (sources are often hand-written docs hap never created). `hap task-source remove` and the Config tab's `x` stay unguarded as the deliberate force path.

**`max_tasks` cap** (per source, default 20) on both daemon refill and manual creation, so a hand-added list can't grow past what the daemon would then refuse to refill.

**Smaller items** — a terminal-bell setting, a clear-task-source action, and `killed` renamed to `daemon_paused` (the token is operator-facing: it reaches the escalation rationale, and should say the operator paused automation rather than imply a crash).

## Safety notes worth a reviewer's eye

Two subtleties drove the design of the source-removal guard, both easy to get wrong:

- **`ChecklistItem.Done` is a pending/not-pending flag, not "completed".** `ParseChecklist` sets `Done` for every mark except `" "`, so an in-progress `[-]` item reads as done and `PendingTasks` counts it as zero work left. A guard built on `PendingTasks` would retire a source out from under an agent that is actively mid-task. Hence `frontend.UnfinishedTasks`, which counts `[-]` explicitly.
- **An empty agent list is not the same as "no agents".** `GetStatus` previously swallowed `ListAgents` errors, returning an empty `MonitoredAgents` with a nil error — which the guard would read as "nothing matches this source". `Status.AgentsKnown` now distinguishes the two and the guard fails closed when the list is unknown. The "all tasks done" arm sits ahead of that gate, since a finished list is removable whoever it feeds.

Removability is also re-validated on confirm, not just when the question is asked: the 2s poll can land in between, and a finished list is exactly the state that both makes a source removable and triggers the daemon to regenerate tasks into that same file.

Task mutations are guarded against a checklist that changed underneath the operator — an action captured against a row aborts rather than mutating the wrong line if the task's text no longer matches at write time.

## Testing

- Full suite green (22 packages) and `golangci-lint` clean, both with the mandatory `vectors cpu` tags.
- The two guard fixes above were each verified by reverting the fix and confirming its regression test fails.
- Live-tested against real herdr agents: the same source refused removal while feeding a working agent with an unfinished task, then offered the confirmation once its list finished; on accept it removed only that entry while its duplicate-path twin survived and the checklist file stayed on disk.

## Release

`herdr-plugin.toml` is intentionally left at `0.3.62`, so merging takes the default automatic patch-bump path. Say the word if this should go out as a minor instead.

https://claude.ai/code/session_01LxMdSdreijPgYXN62Yvebi


0xGosu:
Thanks — all four were real. Fixed in 2f52c0c, each pinned by a test I verified fails without its fix.

**1. Send TOCTOU.** Fixed in the shared path, as suggested: `SendTaskToAgent` now re-resolves the agent by pane id and requires a still-idle state immediately before `SendToAgent`, so both the TUI's open confirmation and the CLI's `--yes` window are covered at the choke point. An unreadable agent list fails **closed** — "we couldn't ask" isn't "it's idle", the same boundary this PR already applies to `AgentsKnown`.

**2. `{cwd}` dropped.** Fixed via the inspector port, preferring `ForegroundCwd` then `Cwd` to mirror `daemon.refreshPaneCwd` exactly. Resolved only when the *effective* template references `{cwd}` — `Prompt()` falls back to `DefaultNextTaskTemplate`, so reading `Template` directly would miss it. That fallback is now `domain.TemplateOrDefault`, and the daemon's own `{cwd}` gate routes through it too, so the two paths can't drift apart later.

**3. Unreadable source treated as retirable.** Agreed, and it was the one asymmetry left in the guard. Reordered: an unreadable checklist now fails closed while an agent still matches the source. A source **no agent matches** stays retirable whatever its file says — that keeps a broken entry cleanable from the tab that shows it as broken, which was the case the exception existed for.

**4. Stale-row guard after delivery.** Inverted to reserve-then-send. `reserveTask` verifies text + pending and marks `[-]` as one locked read-modify-write, then delivers. A failed delivery rolls back to `[ ]`; a failed rollback leaves `[-]`, which merely parks the task (the daemon only ever sends `[ ]` items) rather than re-arming it — strictly the safer direction. The rollback is also claim-scoped: it only reopens an item still carrying this send's `[-]`, so a completion that lands in the window isn't discarded and re-armed.

Two residuals I want to name rather than quietly leave:

- `SendToAgent` is `agent send` then `pane send-keys enter`. If the text lands but the Enter fails, the rollback to `[ ]` re-arms the daemon while the input box already holds the text. The pre-existing mark-after-delivery code left `[ ]` on that same failure, so this isn't a regression — but it isn't closed either.
- The daemon reads the checklist without taking the task-file lock, so it can act on a pre-reservation snapshot. Reserve-then-send narrows that window; closing it needs the daemon to reserve under the same lock, which is beyond this PR.

Happy to take either as a follow-up if you'd rather they land here.

PR #140, opened 2026-07-17
#1 feat: keep learning honest — confirmable variance escalations, live confidence, and rules that start clean (0.4.0)

0xGosu:
Learning and its operator-facing surfaces drifted apart. Each fix here uncovered
the next, so they ship together as `0.4.0` (behavioural learning changes take the
manual minor path — the manifest is bumped in-PR and the merge commit gets tagged).

## A variance escalation could not be accepted

The variance guard escalated with an empty `Suggestion`, so `hap confirm <id>`
failed with *"carries no suggestion to confirm"* — the operator could see the
suggested action was still right but had no way to accept it. The guard ran
*before* `resolveSituation`, so it hardcoded `""`. The rate guard directly below
it already carried the same fix, with a comment describing this exact symptom.

The guard withholds autonomy, not information: it now surfaces the resolved
action, and names the irreversible indicator or the resolve failure (e.g.
`no_task_source`) so a line that genuinely *cannot* be confirmed says why.

This had a hidden second-order cost. `isConfirmation` requires a non-empty
suggestion, so every response to a variance escalation was recorded as a
*correction* — which is excluded from the `confirmation_weight` boost. The bug
wasn't just "I can't click accept"; it silently denied the operator's input its
full learning weight and stalled graduation.

## Every rule displayed `confidence 1.00`

Operator views rendered `SignatureState.CachedConfidence`, a snapshot refreshed
only on confirm/correct and stamped to a fake `1.0` by a reset. It drifted from
the score the decision core actually gates on: a rule scored `0.45` displayed as
`1.00` next to its own *"contradictory history"* escalation. On a live install, 9
of 43 rules showed a false `1.00` — two of them were really `0.24` and `0.31`.

`domain.LiveConfidence` is now the single source of truth, shared by `Decide` and
all eight render sites. Follow-ons the fix exposed:

- **`--min-conf` filtered on the stale column in SQL** while displaying the live
  score, so `--min-conf 0.9` returned rules visibly reading `0.24`. It cannot be
  a SQL predicate (the table only holds the snapshot, which drifts both ways), so
  it moved to where history is loaded.
- **`Decisions` became post-floor**, but it also fed the delete-confirmation
  prompt — and `DeleteSignature` erases every row regardless of the floor. A
  reset rule would have prompted *"delete … and its 0 decision(s)"* then destroyed
  N. Split into `Decisions` (post-floor, for confidence) and `TotalDecisions`.

## `CONF` showed `0.00`, which can never happen

Agreement is `topWeight/total`, so any real history scores strictly above zero —
`0.00` only ever meant "never scored". Stored values jump from `0.000000` straight
to `0.240828`; nothing lies between. It now renders `-`, from a pure function of
the row: an audit row is an immutable snapshot, so deleting a rule must never
rewrite the `CONF` of rows it once backed.

## A rule inherited the LLM's guesses

`applyCorrection` is the only place a rule is born, and only the operator triggers
it — yet the new rule inherited every decision the LLM had already made. Those
aren't agreement: one live signature held 15 LLM decisions spanning **6 distinct
actions**, scoring `0.45`, so a brand-new rule was pinned under the variance floor
despite the operator having agreed once and never disagreed. Worse, it could never
recover — only `applyCorrection` advances the streak, so LLM activity alone can
never graduate a rule, while every inconsistent answer drags the score further down.

A rule now starts clean: the floor is stamped at the newest pre-existing decision,
so it scores `1.00` over its single operator decision. The rows are **kept** — the
floor only hides them from confidence and graduation, exactly as a reset does.
Before a rule exists every decision is `SourceLLM` (`SourceRule` requires a
graduated rule), so this can only ever discard the LLM's guesses.

**`wasAutonomous` went with it.** Keyed off `audit.Status == "auto"`, it did not
mean "a graduated rule acted": it also matched an LLM auto-act on a signature with
no rule, and an auto row from *before* a reset — both stranded the streak at 0 so
the rule could never (re-)graduate. It was redundant anyway: `ObserveConfirmation`
freezes a graduated rule itself, so permanent graduation (FR-007) is unaffected
and is now pinned by a dedicated test.

## Also

Answering Claude preview-layout MCQ forms, where a digit only moves the caret and
Enter commits (`internal/mcqdeliver`).

## Verification

Full suite green (23 packages), `gofmt`/`vet`/`golangci-lint` clean. Verified live
against a running herdr: `0.00` is gone from every operator surface, 197
never-scored rows read `-`, real scores are untouched, and a fresh variance
escalation now carries a confirmable suggestion.

https://claude.ai/code/session_01UprtS7ZEpcgto5x88V1p7b


0xGosu:
Valid finding — fixed in 842696a. I verified each claim before acting:

- `DeleteSignature` really is unfiltered (`DELETE FROM decisions WHERE signature = ?`)
- `TotalDecisions` really was `len(history)` from the 50-row window
- nothing prunes the table — the only other `DELETE` is the escalation prune — so history grows unbounded

So past 50 decisions the prompt quoted 50 and destroyed N. It is also reachable rather than theoretical: a live install already carries a rule at 34 decisions and climbing.

**Fix:** new `Store.CountDecisionsForSignature` doing an unbounded `SELECT COUNT(*)` (served by the existing `idx_decisions_signature (signature, id DESC)` index, so it stays an index-only scan on the listing path), added to `ports.ReadStore` and used to populate `TotalDecisions` in both `Signatures()` and `SignatureDetail()`.

**Regression test** `TestTotalDecisionsCountsBeyondTheHistoryWindow` seeds 63 decisions — above the window — and covers the listing and the detail, plus the property that actually matters: the count the prompt quotes equals what `DeleteSignature` erases. Mutation-tested against the old derivation, where it fails exactly as you described:

```
detail TotalDecisions = 50, want 63
listing TotalDecisions = 50, want 63
delete erased 63 decisions but the prompt quoted 50
```

Full suite green (23 packages), gofmt/vet/golangci-lint clean.

One note on scope: `Decisions` (post-floor) deliberately keeps the 50-row window — it feeds the confidence figure, where the window is the intended recency horizon and matches the daemon's own gating read. Only `TotalDecisions`, which describes the stored rows a delete destroys, needed the exact count.

PR #141, opened 2026-07-17
#1 fix: dedup pending escalations by normalized content, not exact string

0xGosu:
## Problem

Herdr re-delivers an attention event for the same agent after a delay — the agent flips `done`→`idle` when the operator reads the pane, re-firing the same screen. The pending-escalation dedup used an **exact** `pane_excerpt` match keyed on `situation_type`, so it missed those re-fires:

- the pane's volatile chrome had ticked between captures (a spinner glyph, or an elapsed/countdown counter — e.g. the real `✻ Waiting for API response · will retry in 2m 2s` line, which becomes `2m 0s` two seconds later);
- a `done`→`idle` re-delivery **reclassifies** the same screen to a different `situation_type` (the classifier gates approval/choice on herdr reporting `blocked`).

Result: the operator gets asked the same question twice.

## Change

Match on **normalized content** instead of an exact string, keyed on agent + agent type + content (`situation_type` dropped as status-derived; `agent_status` was never in the key):

- `domain.NormalizeForDedup` elides only the volatile agent-TUI status lines — identified by their live-counter markers (`esc to interrupt`, `will retry in`) or a spinner-**exclusive** star glyph (`✽✻✶✳✢`) — and collapses whitespace. It deliberately does **not** use `MaskVolatile`, whose `<path>`/`<num>` masking would collapse two genuinely different questions (`rm -rf /tmp/foo` vs `/tmp/bar` → same string). The `●` message bullet and markdown bullets are real content and are never elided.
- Empty-content fallback keys on `situation_type` for the pane-read-failure path.
- `Store.DuplicatePendingEscalation` (bool, exact SQL) → `PendingEscalationExcerpts` (returns the agent's pending escalations for a Go-side compare). Fails **open** on a store error, per "never a silent drop".

## Safety scoping

The check stays where escalations are raised (`escalate()`, the pre-LLM-review branch, and the pane-read-failure path) and **never runs before the decision core**. The runaway guard, retry ceiling, and shadow-mode learning all work by re-processing repeated identical events, so gating them would bypass a safety control. (An earlier draft that gated in `decideAndAct` broke 8 daemon tests for exactly this reason.)

## Testing

- New: `TestNormalizeForDedup*` (incl. distinct-content and `●`/`*`-bullet regressions), `TestDuplicatesPendingEscalation`, `TestPendingEscalationExcerpts`, `TestPipelineDedupAbsorbsChromeJitter`, `TestPipelineDedupFailsOpen`.
- Full touched-package suites pass (`domain`, `store`, `config`, `daemon`, `verifyunblock`, `frontend`), including the 8 previously-regressed daemon safety/learning tests.
- `go build`, `go vet`, `gofmt`, `golangci-lint` — all clean.
- Code-reviewed; the one finding (the `●` glyph in the spinner set) is fixed and covered by a regression test.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -754,36 +754,53 @@ func (d *Daemon) hasOpenEscalation(ctx context.Context, agentID string) bool {
 	return false
 }
 
-// duplicatePendingEscalation reports whether the captured situation exactly
-// matches an escalation the user has not yet handled. It is the content-level
-// sibling of hasOpenEscalation (which is pane/agent-level and used by the
-// reconcile sweep): finer, so a genuinely new situation on a pane that already
-// has an open escalation still gets processed. The pane excerpt is truncated
-// exactly as escalate() stores it so the comparison lines up.
+// duplicatePendingEscalation reports whether the captured situation repeats an
+// escalation the operator has not yet handled — so re-raising it would ask the
+// same question twice. Herdr re-delivers an attention event for one agent after
+// a delay (the agent flips done->idle when the operator reads the pane), and
+// the exact-string check it replaces missed those re-fires because the pane's
+// volatile chrome (a spinner tick, an elapsed/countdown counter) had moved on.
+//
+// The key is agent + agent type + the NORMALIZED pane content
+// (domain.NormalizeForDedup elides that chrome). The agent status is
+// deliberately excluded — it is the field that legitimately CHANGES between the
+// duplicates — and so are trigger() (which embeds it) and situation_type (which
+// the classifier DERIVES from it). See domain.DuplicatesPendingEscalation.
+//
+// It runs only where escalations are raised (escalate() and the
+// pane-read-failure path), NOT before the decision core: the rate guard, retry
+// ceiling, and shadow-mode learning all work by re-processing repeated
+// identical events, so suppressing those would bypass a safety control.
+//
+// It is the content-level sibling of hasOpenEscalation (which is
+// pane/agent-level and used by the reconcile sweep): finer, so a genuinely new
+// situation on a pane that already has an open escalation still gets processed.
+// The pane excerpt is truncated exactly as the write paths store it so the
+// comparison lines up.
 //
 // Fails OPEN (returns false) on a store error — the opposite of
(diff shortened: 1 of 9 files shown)


0xGosu:
Thanks @charliecreates — the kernel here is valid: versus the old `SELECT EXISTS(...)` (a bool that short-circuits in SQL), the normalized check materializes the pending rows and runs a per-row regex normalization on the synchronous event loop, with no code-enforced bound. Addressed in 6d778be.

**Fix:** `PendingEscalationExcerpts` now caps the fetch at `PendingEscalationDedupLimit = 128`, newest-first (`ORDER BY a.id DESC LIMIT ?`), so the materialization and the Go-side normalization the daemon does over it are constant regardless of backlog size. Added `TestPendingEscalationExcerptsBounded` (seeds `limit + 25` pending escalations, asserts the fetch returns exactly the cap and still includes the newest row).

**On the scan itself:** it's already bounded, not unbounded. `status = 'escalated'` is served by `idx_audit_status(status, id DESC)`, and the dedup prevents identical escalations from stacking, so the *global* pending-escalated set is naturally small — the planner seeks to that partition rather than scanning an agent's full history. The LIMIT hardens the part that actually grew: the Go-side per-row normalization.

**On the persisted canonical-key suggestion:** I considered it and chose the bound instead, deliberately. Storing a hash of `NormalizeForDedup(excerpt)` would freeze the normalization algorithm into the schema — and the whole point of this change is a normalizer that can evolve (e.g. as we learn new volatile-chrome shapes). For a *transient* dataset (pending escalations resolve/prune quickly, unlike the long-lived learned signatures) that coupling and its backfill aren't warranted; the naturally-small, index-served, now-LIMITed fetch keeps the hot path constant without it.

**Correctness:** preserved in practice. An agent shows one screen at a time and the dedup stops identical escalations from stacking, so a re-delivery always matches a *recent* escalation, well inside the newest-first window. In the pathological case of 128+ simultaneous distinct unresolved escalations for one agent, the worst case is one redundant operator ask — never a silent drop (the check already fails open on error for the same reason).

Full store + daemon suites green after the change; lint/vet/gofmt clean.

PR #167, opened 2026-07-18
#157 fix: append generated tasks to the exhausted declared source

### LINKED ISSUE #157: Exhausted-source task generation registers a second task source, breaking "hap task <agent>" (ambiguous) and the default template marking commands
## Summary

Found during a live feature test (2026-07-17, v0.4.7 dev build). When an agent's **declared** task source runs dry and the LLM task generation produces new tasks, confirming the escalation writes them to a NEW per-agent file (`<state>/tasks/<agent>.md`) and registers it as a **second** `[[task_sources]]` entry for the same agent — instead of appending to the exhausted declared source.

## Consequences observed live

1. `hap task <agent> …` immediately starts failing with:
   `error: agent "<name>" matches 2 task sources (…); use --path <file> to pick one`
2. The default prompt template tells the agent to run `hap task <agent> list/start <n>/done <n>` — all of which now hit the ambiguity error. The template's `--path` fallback names only ONE of the two files, so the agent can mark generated tasks but not declared ones (or vice versa). In the live run, the agent completed tasks 2 and 3 of the generated list but the checklist stayed unmarked.
3. The TUI Tasks tab and `task-source list` show two sources feeding one agent, which also changes the retirement (`x` on header) and max_tasks accounting story.

## Expected

Generated tasks for an **exhausted declared source** should append to that source's file (the skill/docs frame generation as refilling the list; `max_tasks` is documented as capping exactly that growth). The separate per-agent file makes sense only for the bootstrap no-source case.

0xGosu:
Fixes #157.

## Problem

When an agent's **declared** task source ran dry and LLM task generation produced new tasks, confirming the escalation wrote them to a NEW per-agent file (`<state>/tasks/<agent>.md`) and registered it as a **second** `[[task_sources]]` entry for the same agent:

1. `hap task <agent> …` immediately failed with `matches 2 task sources … use --path`.
2. The default prompt template's marking commands (`hap task <agent> list/start/done`) all hit that error, so agents completed generated tasks but the checklist stayed unmarked.
3. `max_tasks` accounting and source retirement split across two files.

## Fix

The confirm path (`acceptGeneratedTask`) now **re-resolves the agent's declared sources at confirm time** and appends the generated tasks to the matched source's own file — generation refills the declared list, exactly as the docs frame it. The per-agent bootstrap file + registration remains only for the genuine no-source case.

- **`config.TaskSource.MatchesAgent`** — the agent-selector semantics (id / type / short name; `""` = any) extracted into one shared definition; the daemon's `matchTaskSource` now uses it too, so daemon and frontend can't drift.
- **`appendGeneratedTasks`** — claims the escalation first (appends are not idempotent), appends all tasks pending `[ ]` inside one locked mutation, sends the first task rendered through the **source's own template** pointing at the declared path, and flips it to `[-]` only **after** a successful send (text-guarded), so a failed send never strands an undeliverable in-progress item.
- **max_tasks respected at confirm time** — a full list refuses pre-claim (escalation stays pending so the operator can prune and retry); a race that shrinks the room truncates the appended set. The daemon's generation gate now treats an **at-cap** list as full (`n >= limit`, matching `task add`), so it can no longer raise escalations every confirm must refuse.
- **Multiple matching sources** — deterministic pick mirroring the daemon's precedence (pending work → has items → config order), never an error.
- **Defense-in-depth** — `addTaskSourceIfAbsent` refuses to register a source when an existing non-empty selector already matches the agent (by id, type, or name), converting any residual duplicate-creation path into a clear error.

## Tests

- 8 new frontend confirm tests: append/multi-append/without-send/idempotency/cap-refusal+truncation/source-template/missing-declared-file/duplicate-guard, plus daemon-parity selector cases (id, type, workspace label via locator, raw workspace-id fallback) and multi-source precedence.
- `MatchesAgent` table test; daemon cap-gate test tightened to the new at-cap boundary.
- Full suite: `go test -tags "vectors cpu" ./... -count=1` — 23 packages, 0 failures. `golangci-lint`: 0 issues.
- Local integration suite (real herdr): 5 pass, 3 expected skips.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -245,6 +245,20 @@ type TaskSource struct {
 	MaxTasks int `toml:"max_tasks,omitempty"`
 }
 
+// MatchesAgent reports whether this source's agent selector matches the given
+// agent. The selector matches the agent/pane id, the agent type, or the
+// agent's short name; an empty selector matches any agent. This is the single
+// definition of the selector semantics — the daemon's task-source matcher and
+// the frontend's confirm-time source resolution must agree, or a generated
+// task confirm can bootstrap a duplicate source next to a declared one.
+func (s TaskSource) MatchesAgent(agentID, agentType, agentName string) bool {
+	if s.Agent == "" {
+		return true
+	}
+	return s.Agent == agentID || s.Agent == agentType ||
+		(agentName != "" && s.Agent == agentName)
+}
+
 // DefaultMaxTasks is the fallback for TaskSource.MaxTasks when unset (0).
 const DefaultMaxTasks = 20
 
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -1224,3 +1224,32 @@ func TestCaptureDelayLoadedFromTOML(t *testing.T) {
 		t.Errorf("unmatched start delay = %v, want 10s", got)
 	}
 }
+
+func TestTaskSourceMatchesAgent(t *testing.T) {
+	// MatchesAgent is the single definition of the agent-selector semantics,
+	// shared by the daemon's task-source matcher and the frontend's
+	// generated-task confirm: id, type, or short name; "" matches any.
+	cases := []struct {
+		name     string
+		selector string
+		id, typ  string
+		short    string
+		want     bool
+	}{
+		{"empty selector matches any", "", "w1:p1", "claude", "alpha", true},
+		{"matches agent id", "w1:p1", "w1:p1", "claude", "alpha", true},
+		{"matches agent type", "claude", "w1:p1", "claude", "alpha", true},
+		{"matches short name", "alpha", "w1:p1", "claude", "alpha", true},
+		{"no match", "beta", "w1:p1", "claude", "alpha", false},
+		{"empty short name never matches a named selector", "alpha", "w1:p1", "claude", "", false},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			src := TaskSource{Agent: tc.selector}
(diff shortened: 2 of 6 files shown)


0xGosu:
Addressed both review findings (and rebased onto latest main, which brought in #165's reserve/release task lifecycle):

**1. Escalation resolved before the append is durable** — the append now runs **before** the claim as one locked, idempotent compare-append: tasks whose text the checklist already carries are skipped (so a retry or the loser of a double-confirm can't duplicate them), the `max_tasks` cap is enforced on the actually-missing subset inside the same lock, and a repeated task within one suggestion is appended once. Any append-side failure — cap full, unreadable file, failed write — now surfaces while the escalation is still pending and retryable; `ResolveEscalation` only gates the non-idempotent send. A first task that already sits `[x]`/`[-]` in the file also refuses **pre-claim** with an actionable message instead of dying at reserve time after the claim.

**2. First task not reserved before delivery** — delivery now mirrors `SendTaskToAgent`'s load-bearing order exactly: `reserveTask(firstIndex, text)` under the file lock **before** the send (the daemon's idle flow can never hand the item out mid-send), `releaseTask` on a failed send (back to `[ ]` for the declared flow), and the same parked-`[-]` error if the rollback itself fails.

The rebase also surfaced an interaction with #165: re-confirms/regenerations of the agent's own *bootstrapped* generated file must keep the marker-carrying rewrite flow, so the append path now applies only to sources declared elsewhere (which also makes legacy dual-source configs from the bug era prefer the declared source). New tests cover the mid-send reservation state (via a send hook), send-failure rollback, pre-claim not-pending refusal, cap-refusal retryability, and suggestion dedup. Full suite + local integration suite green.

PR #169, opened 2026-07-18
#161 fix: recognize codex hard usage-limit banner and unknown ■-banner errors

### LINKED ISSUE #161: Strengthen codex error capture: hard usage-limit banner unrecognized; generalize ■-banner error detection
## Context

Live testing (2026-07-18) plus operator-captured samples (codex_error.log) show codex error screens that hap's codex error detection does not recognize. Current coverage in `internal/domain/codexerror.go` (`CodexErrorForm`, gated on `agent_type == "codex"` in `internal/classify/classify.go`) is exactly two forms:

1. `CodexErrorInterrupted` — `codexInterruptedRE` (`codexerror.go:16`) — matches the interrupt banner. **Works** (verified live: classified as `error`, escalated in shadow mode).
2. `CodexErrorRateLimit` — the *soft* "Approaching rate limits" model-switch modal (header + "Switch to … for lower credit usage?" + 3 options + end-anchored footer, `codexerror.go:22-24, 51-68`).

## Captured samples (real screens)

Error 1 — covered today:

```
■ Conversation interrupted - tell the model what to do differently. Something went wrong? Hit `/feedback` to
report the issue.
```

Error 2 — NOT covered:

```
■ You've hit your usage limit. Upgrade to Pro (https://chatgpt.com/explore/pro), visit https://chatgpt.com/codex/settings/usage to purchase more credits or try again at Jul 23rd, 2026 4:19 AM.
```

## Problem

The hard usage-limit banner is neither the interrupt phrase nor the model-switch modal, so `CodexErrorForm` returns false and the pane classifies as ordinary idle. Consequences observed live when the account hit its weekly cap:

- The blocked agent looks *idle*, so the idle flow runs — a learned idle `@noop` rule answers "do nothing" and the operator is never told the agent is hard-blocked until Jul 23.
- The same error text also surfaced only indirectly via `[task_gen_failed]` when hap's own codex-backed task-generate CLI failed.

An agent that cannot make progress for days should classify as an **error** and escalate, not sit silently idle.

## Suggested work

1. **Add the hard usage-limit form**: e.g. `codexUsageLimitRE = (?i)you'?ve hit your usage limit\b` (optionally anchored near a line-leading `■`), new kind `CodexErrorUsageLimit = "usage-limit"` in `CodexErrorForm`. Mask the volatile reset timestamp ("try again at Jul 23rd, 2026 4:19 AM") so instances dedup to one `error:usage-limit` signature — the whole point of the stable `ErrorSummary` labels.
2. **Generalize codex banner capture**: codex prefixes blocking error/notice banners with `■` at line start. Consider an end-anchored fallback envelope (`■ <text>` as the last transcript block before the composer/footer) that classifies *unknown* codex banners as errors too — network drops ("stream error", "connection reset"), auth failures, quota variants — instead of requiring a new regex per phrasing. Known benign banners can be allowlisted out if needed.
3. **Tests**: table-driven cases in `codexerror_test` for both captured samples above (verbatim), a masked-timestamp dedup case, plus golden classifier fixtures under `internal/classify/testdata/` so `CodexErrorForm` gating and `error:<kind>` signatures are locked in.

Sample screens are preserved verbatim in `codex_error.log` (repo root, uncommitted) if more context is needed.

0xGosu:
Fixes #161.

## Problem

Codex's hard usage-limit banner (`■ You've hit your usage limit. … try again at Jul 23rd, 2026 4:19 AM.`) was neither the interrupt phrase nor the model-switch modal, so `CodexErrorForm` returned false and the pane classified as ordinary **idle** — a learned idle `@noop` rule answered "do nothing" and the operator was never told the agent was hard-blocked for days.

## Changes

- **`usage-limit` kind** (`domain.CodexErrorUsageLimit`): recognized via the stable phrase, matched only against a **live extracted banner** so a stale copy in scrollback cannot re-error a recovered agent. The stable kind label is the `ErrorSummary`, so the volatile reset timestamp never enters the signature — all instances dedup to one `error:usage-limit` rule.
- **Generic `■`-banner fallback** (`domain.ExtractCodexErrorBanner`): the last line-leading `■` block standing at the end of the pane (tolerating only blank lines, an unstripped `›` composer line, and one `model · /cwd` status-footer line after it) classifies as an error even when the phrasing is unknown — network drops, auth failures, quota variants. The kind is `banner: <masked text>` (truncated on a rune boundary), so **distinct** banners keep **distinct** signatures while volatile tokens are masked out. End-anchoring keeps scrollback copies from erroring a working agent.
- **Structural kinds now take precedence over the generic `error:`-line scrape in `enrich`** (agent-type-gated): build output printing `error: …` above a live usage-limit banner no longer fragments the signature per build error.
- A mostly-volatile banner still classifies as an error but trips the over-masking guard: no signature is minted, so it always escalates (fail-safe) — pinned by test.

## Tests

- New `internal/domain/codexerror_test.go`: both live-captured samples verbatim, masked-timestamp signature dedup, distinct-banner signature separation, end-anchoring/scrollback rejection, CRLF/empty-pane/unstripped-composer edge cases, over-mask fail-safe, rune-safe truncation.
- Golden classifier fixtures `error_codex_usage_limit.txt` + `error_codex_banner.txt` (both classified at **idle** — the live failure mode) with exact-kind assertions in `TestCodexErrorBannersDetectedAtIdle`.

## Verification

- Full unit suite (`-tags "vectors cpu"`, 23 packages): pass.
- Integration suite (`-tags "integration vectors cpu"`): 5 pass, 3 skip (optional deps absent).
- `gofmt`, `go vet`, `golangci-lint`: clean.

### DIFF
--- a/internal/classify/classify.go
+++ b/internal/classify/classify.go
@@ -60,9 +60,10 @@ func DefaultRules() []config.ClassifierRule {
 			Regex: nil,
 		},
 		{
-			// Codex's interrupt screen and footer-anchored rate-limit model
-			// switch modal are detected structurally via domain.CodexErrorForm
-			// at the error position in Classify, same as Claude above.
+			// Codex's interrupt screen, footer-anchored rate-limit model
+			// switch modal, hard usage-limit stop, and end-anchored "■" error
+			// banners are detected structurally via domain.CodexErrorForm at
+			// the error position in Classify, same as Claude above.
 			AgentType: "codex", Situation: "error",
 			Regex: nil,
 		},
@@ -181,8 +182,9 @@ func (c *Classifier) Classify(agentType, agentStatus, pane string) domain.Situat
 		if !matched && r.situation == domain.SituationError && strings.EqualFold(agentType, "claude") {
 			_, matched = domain.ClaudeErrorForm(pane)
 		}
-		// Codex's interrupt screen and rate-limit model switch modal are
-		// detected structurally the same way, scoped to codex.
+		// Codex's interrupt screen, rate-limit model switch modal, hard
+		// usage-limit stop, and live "■" error banners are detected
+		// structurally the same way, scoped to codex.
 		if !matched && r.situation == domain.SituationError && strings.EqualFold(agentType, "codex") {
 			_, matched = domain.CodexErrorForm(pane)
 		}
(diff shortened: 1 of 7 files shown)


0xGosu:
Addressed in 557cc70: the footer matcher's cwd portion is relaxed from `/\S*( \(deleted\))?` to `/[^\n]*$`, and the model portion from `[^\n·]*` to `[^\n]*` so the greedy match keys on the **last** ` · /` — now byte-consistent with `codexComposerBeforeFooterRE`'s footer shape. Accepts cwds with spaces, "(deleted)" suffixes, and footers carrying extra `·`-separated segments; rejecting a real footer reads as "no live banner" and silently idles a hard-blocked agent, so the matcher errs loose (spurious escalation fails safe, same accepted risk as the composer regex).

Regression coverage: three new `TestExtractCodexErrorBanner` rows (space cwd, deleted cwd, two-`·` footer) and the `error_codex_banner.txt` classifier fixture footer now uses `/workspaces/my project`, routing the space-cwd shape through the full classify path (golden still pins `type=error` at idle). Full unit suite (23 packages) green.

PR #176, opened 2026-07-18
#175 fix: LLM @noop no longer parks exhausted-source generation; LLM decisions get signature rows

### LINKED ISSUE #175: LLM-recorded @noop on a state-row-less idle signature silently suppresses exhausted-source task generation and cannot be removed via CLI
Found while live-verifying the recent fix batch (2026-07-18, main @ 3cbdbbb).

## Sequence observed live

1. An agent's declared task source ran dry; on the first post-exhaustion idle the daemon consulted the LLM, which correctly answered `@noop` ("nothing left to send"). That decision was recorded (`decisions` row, `source=llm`) against the idle signature — a signature that has **no state row** in `signatures` (the known LLM-auto-learned visibility gap).
2. Every later idle for that signature then resolved through the learned-`@noop` precedence in idle resolution — which runs BEFORE declared-source resolution — so the `task_source_exhausted` → generation path (`llm.task_generate_command`) was never reached again. The agent just produced `[shadow_mode]` "do nothing (no reply needed)" escalations forever.
3. Operator remediation is blocked: `hap signatures delete idle:<prefix> --yes` fails with `error: no signature matches "idle:…"` because there is no `signatures` row to address — the `@noop` decision is CLI-unreachable. (`hap signatures reset` fails the same way.)
4. The only working recovery found: resolve a pending escalation for that situation with `--action "@next_task:declared"`, which records a correction that flips the plurality; the next idle then reached `task_source_exhausted` and generation fired normally.

## Why this matters

A single LLM `@noop` — reasonable in the moment — permanently parks the "refill the exhausted list" feature for that signature, silently, with no CLI way to see the rule (`signatures list` omits it) or delete/reset it. The operator sees only repeating "do nothing" escalations with `rule=[none yet]`, which actively hides that a learned rule is driving the outcome.

## Suggested directions

- Exclude LLM-sourced `@noop` decisions from the idle learned-noop precedence when the agent has a matching declared task source with generation configured (an exhausted list is a state the daemon is designed to refill; a consult noop should not lock that out), or scope the noop-honor to operator-sourced decisions.
- Give every recorded decision an addressable signature: create the state row on first LLM decision (also fixes the `signatures list` visibility gap), so `signatures delete`/`reset` work.
- Escalation display: when a decision-table plurality (not a state row) drives the suggestion, show that provenance instead of `rule=[none yet]`.

0xGosu:
Fixes #175.

## Problem

A single LLM-answered `@noop` on an idle signature (reasonable in the moment — the declared task source had just run dry):

1. **Parked the refill feature forever** — idle resolution honored the learned-noop plurality BEFORE declared-source resolution, and the plurality was source-blind, so `task_source_exhausted` → generation (`llm.task_generate_command`) was never reached again. The agent produced `[shadow_mode]` "do nothing" escalations indefinitely.
2. **Was CLI-unreachable** — no `signatures` state row was created, so `hap signatures list` omitted the rule and `signatures delete/reset` failed with `no signature matches`.
3. **Hid its provenance** — escalations showed `rule=[none yet]` while the decision-table plurality actually drove the suggestion.

## Fix

- **`internal/domain`**: `ConfidenceResult.TopActionOperatorBacked` — true when an operator- or rule-sourced decision backs the top action (a `SourceRule` row implies past operator confirmation). The idle noop precedence now requires `TopActionOperatorBacked || DeclaredTask == nil`:
  - operator-taught noops still outrank declared tasks (unchanged behavior, now provenance-scoped);
  - a pure-LLM noop yields to a pending declared task, and to `task_source_exhausted` → `ActionGenerateTask` (the #175 repro);
  - with no declared source an LLM noop still silences a chatty idle agent (the original nudge-loop case);
  - approval/choice/error noop handling is untouched.
- **`internal/daemon`**: new atomic `Store.EnsureSignature` (single `INSERT OR IGNORE`, never touches an existing row) called after all four LLM `RecordDecision` sites (noop promotion, action send, multi-tab series, remote-env picker), so every LLM-learned rule is listable/deletable/resettable. Companion: `applyCorrection`'s clean-slate `DecisionFloorID` stamp is re-keyed from "no state row" to "no operator/rule evidence in the pre-existing history" (`domain.HasOperatorEvidence`), preserving the guarantee that the first operator correction floors out prior LLM guesses now that a row exists by then.
- **`internal/store`**: idempotent `migrate()` backfill creates shadow rows for decisions-only signatures in existing databases (type fields from each signature's newest decision by id).
- **Display**: no code change needed — with the row present, escalations render the real rule summary (`rule=[shadow — 0/N confirmations, … top action "@noop" …]`); `rule=[none yet]` remains only for truly-unseen signatures.

## Testing

- New/updated unit tests across `internal/domain` (provenance matrix, the #175 repro, exhausted-without-generation escalation, nudge-loop preservation, non-idle noop unchanged), `internal/daemon` (row creation on the noop/send/series/picker paths; first-operator-correction flooring with a pre-existing LLM-created row), `internal/store` (`EnsureSignature` contract, migration backfill end-to-end incl. delete + re-migrate), `internal/cli` (LLM-learned signature list/reset/delete + escalation rule rendering).
- Full suite `go test -tags "vectors cpu" ./... -count=1` green; `go vet` and `golangci-lint run --build-tags "vectors,cpu"` clean.

### DIFF
--- a/internal/cli/cli_test.go
+++ b/internal/cli/cli_test.go
@@ -549,6 +549,60 @@ func TestSignaturesReset(t *testing.T) {
 	}
 }
 
+func TestLLMLearnedSignatureIsCLIAddressable(t *testing.T) {
+	// #175: an LLM-recorded decision always has a signatures state row now
+	// (written by the daemon at decision time, or backfilled by the store
+	// migration), so the learned rule is listable, resettable, deletable,
+	// and named on escalations instead of hiding behind "rule=[none yet]".
+	app, st := testApp(t)
+	ctx := context.Background()
+	now := time.Now()
+	const sig = "idle:ffff0000aaaa1111"
+	if err := st.UpsertSignature(ctx, domain.SignatureState{
+		Signature: sig, SituationType: domain.SituationIdle,
+		AgentType: "claude", Mode: domain.ModeShadow, UpdatedAt: now,
+	}); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := st.RecordDecision(ctx, domain.DecisionRecord{
+		Signature: sig, SituationType: domain.SituationIdle, AgentType: "claude",
+		ChosenAction: domain.ActionNoop, Source: domain.SourceLLM, CreatedAt: now,
+	}); err != nil {
+		t.Fatal(err)
+	}
+	st.AppendAudit(ctx, domain.AuditRecord{Signature: sig,
+		Trigger: "agent idle", SituationType: domain.SituationIdle,
+		Action: "escalated", Rationale: "[shadow_mode]", Status: "escalated", CreatedAt: now})
+
+	out, err := run(t, app, "signatures")
+	if err != nil || !strings.Contains(out, "idle:ffff0000aaa") {
+		t.Errorf("LLM-learned signature must be listed (%v):\n%s", err, out)
+	}
+	out, err = run(t, app, "escalations")
+	if err != nil {
+		t.Fatal(err)
+	}
+	if !strings.Contains(out, `top action "@noop" over 1 decision(s)`) ||
+		strings.Contains(out, "rule=[none yet]") {
+		t.Errorf("escalation must name the LLM-learned rule, got:\n%s", out)
+	}
+	out, err = run(t, app, "signatures", "reset", "idle:", "--yes")
+	if err != nil || !strings.Contains(out, "reset signature "+sig) {
+		t.Errorf("LLM-learned signature must be resettable (%v):\n%s", err, out)
+	}
+	out, err = run(t, app, "signatures", "delete", "idle:", "--yes")
+	if err != nil || !strings.Contains(out, "deleted signature "+sig) ||
+		!strings.Contains(out, "1 decision(s)") {
+		t.Errorf("LLM-learned signature must be deletable with its decisions (%v):\n%s", err, out)
+	}
+	if state, _ := st.GetSignature(ctx, sig); state != nil {
+		t.Error("signature row should be gone")
+	}
+	if recs, _ := st.DecisionsForSignature(ctx, sig, 10); len(recs) != 0 {
+		t.Error("decision history should be gone")
+	}
+}
+
 func TestDismissCLI(t *testing.T) {
 	app, st := testApp(t)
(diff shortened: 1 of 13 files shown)


0xGosu:
@charliecreates Thanks — the aging-out mechanism you describe is real, but I evaluated the claimed harm against the actual data flow and it doesn't hold, while a closely related refinement does. Addressed in 09dfab3:

**Why the unbounded check would be wrong here:** every consumer of decision history caps at the newest 50 rows — the decision core (`daemon.go:1146`), the LLM promotion path (`daemon.go:2931`), `applyCorrection` itself, and the operator-facing views (`frontend.go:2814/2854`) — and the recency decay (0.85^n) zeroes rows long before the cap. So operator evidence older than the window already contributes nothing to confidence, graduation, TopAction, or display anywhere; advancing the floor over it excludes nothing that still counts. In your scenario (50+ LLM guesses burying old operator rows), the floor advance discards exactly the in-window LLM guesses and starts the rule clean at 1.00 over the operator's fresh decision — the designed outcome. An unbounded evidence check would instead *keep* those 50 contradictory LLM guesses in the live scoring window after the operator speaks, pinning the rule under the variance guard — the precise pathology the clean-slate floor exists to prevent (see the comment block above the stamp). Graduated rules are unaffected either way: their autonomous acts record `SourceRule` rows continuously, so evidence never ages out of the window. The streak/mode/graduation state live on the signatures row and are untouched by the floor in all cases.

**What was genuinely off — now fixed:** the evidence check should be scoped to the **post-floor** slice of that same window, so the floor decision keys on exactly the evidence set confidence/graduation consume. `09dfab3` changes the trigger to `!HasOperatorEvidence(DecisionsSince(history, state.DecisionFloorID))`. This fixes a real case the full-window check got wrong: a **reset** rule whose post-reset history is pure LLM guesswork — the pre-reset operator rows still inside the window used to block the stamp, letting post-reset LLM noise pollute the re-earning score. Covered by the new `TestCorrectionFloorsPostResetLLMGuesses`, and the deliberate capped-window semantics are now documented at the stamp site.

Full suite, vet, and golangci-lint are green on the branch.

PR #181, opened 2026-07-18
#180 feat: accept a busy agent's generated-task suggestion by queuing to its task list

### LINKED ISSUE #180: Escalations: add "a" key to queue LLM-generated tasks to the agent's list without sending
## Problem

Confirming an escalation that carries an LLM-generated task (`enter`/`y` on the Escalations tab) appends the task to the agent's declared source **and** sends the first task to the pane. If the agent has since started other work, the confirm is refused:

> ✗ agent is no longer idle (working); the suggested task is stale — dismiss it or wait until the agent is idle

The suggested tasks are often still valid; the operator just wants to **queue** them for the agent to pick up on its next idle, not interrupt current work. Today the only alternative is to dismiss (discard) them.

## Proposal

Add an `a` keystroke on the Escalations list and detail overlay that **adds the generated tasks to the declared task list without sending**. Because nothing is delivered to the pane, a busy agent is fine — the daemon delivers the queued task on the next idle. Adding counts as accepting the escalation: it resolves the escalation and records the acceptance in Audits, exactly like a normal confirm.

Implementation reuses `acceptGeneratedTask(ctx, audit, send=false)`, which already appends unconditionally and only sends when `send==true`; the busy/staleness refusal is gated so it only applies when actually sending.

0xGosu:
Closes #180.

## Problem

Confirming an escalation that carries an LLM-**generated task** appended the tasks to the agent's declared source **and** sent the first to the pane. If the agent had since started other work, the whole confirm was refused:

> ✗ agent is no longer idle (working); the suggested task is stale — dismiss it or wait until the agent is idle

The suggested tasks are often still valid — the operator just wants to **queue** them for the agent to pick up on its next idle, not interrupt current work. The only alternative was to dismiss (discard) them.

## Change

Gate the busy/staleness refusal in `acceptGeneratedTask` by `send`: queuing to the declared list (`send=false`) never touches the pane, so a busy agent is fine — the daemon delivers the item on the agent's next idle. Only an actual send is refused, now wrapping a new exported sentinel `ErrSuggestionStaleAgentBusy` so callers can detect it via `errors.Is`.

- **TUI** — `enter`/`y` still confirms+sends. On a busy refusal it chains a prompt (`openAddPromptMsg` → `openAddPrompt`): `agent is busy — add the tasks to its task list instead? [Y/n]`, defaulting to yes (the operator already chose to accept, and queuing sends nothing). Yes queues via `Confirm(send=false)`; no leaves it pending. Idle agents still send directly.
- **CLI** — `hap confirm <id>` **without** `--send` queues the tasks (works while busy) with a clearer message; `--send` on a busy agent still refuses and points at dropping `--send`.

Queuing resolves the escalation and records the acceptance in Audits, same as a normal confirm. No schema/store changes — it reuses the existing `ResolveEscalation` + `ActionNextDeclaredTask` correction → daemon audit-lineage row.

## Tests

Added 8 unit tests across frontend / TUI / CLI:
- frontend: add-only succeeds while busy; send-refusal wraps the sentinel.
- TUI: busy send opens the add prompt; yes queues; no leaves pending; bare-Enter queues; idle sends directly.
- CLI: `confirm` without `--send` queues while busy; `--send` refuses while busy.

Full suite (`go test -tags "vectors cpu" ./...`), `go vet`, `gofmt`, and `golangci-lint` all pass.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -715,10 +715,26 @@ func confirm(ctx context.Context, app *frontend.App, out io.Writer, args []strin
 	if err != nil {
 		return fmt.Errorf("invalid audit id %q", idArg)
 	}
+	// Detect a generated-task suggestion up front so the success line can say
+	// the tasks were queued to the agent's declared list. Without --send they
+	// are only appended (the daemon delivers the first on the agent's next
+	// idle), so confirming without --send is the way to accept a suggestion for
+	// a busy agent without interrupting it (issue #180).
+	genTask := false
+	if audit, gerr := app.Store.GetAudit(ctx, id); gerr == nil && audit != nil {
+		genTask = strings.HasPrefix(audit.Suggestion, domain.SuggestTaskPrefix)
+	}
 	if err := app.Confirm(ctx, id, *send); err != nil {
 		return err
 	}
-	fmt.Fprintf(out, "confirmed escalation #%d (recorded as a learning event)\n", id)
+	switch {
+	case genTask && *send:
+		fmt.Fprintf(out, "confirmed escalation #%d — added the suggested task(s) to the agent's task list and sent the first\n", id)
+	case genTask:
+		fmt.Fprintf(out, "confirmed escalation #%d — added the suggested task(s) to the agent's task list (not sent)\n", id)
+	default:
+		fmt.Fprintf(out, "confirmed escalation #%d (recorded as a learning event)\n", id)
+	}
 	return nil
 }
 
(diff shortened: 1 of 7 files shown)


0xGosu:
Follow-up on the review here ("the TUI decline path is not functional as rendered"): confirmed valid — the busy-agent add prompt pre-filled `y` and the input box appends keystrokes, so typing `n` produced `yn` (still `HasPrefix "y"`), making the decline unreachable. Fixed in PR #184 (commit f175688): the prompt now opens empty with a `[y/N]` hint, so a bare Enter cancels and only an explicit `y` queues. Regression tests added.

PR #184, opened 2026-07-18
#183 fix: append generated tasks to the agent's list instead of replacing it

### LINKED ISSUE #183: Generated-task confirm replaces the agent's task list instead of appending
## Problem

Confirming an LLM-generated-task escalation for an agent whose task list is the plugin's own bootstrapped generated file REPLACES the list with only the newly-suggested tasks — any existing task not re-listed by the new generation is **dropped** (data loss).

Root cause: the bootstrap path in `acceptGeneratedTask` re-renders the file from only the current suggestion (`RenderGeneratedTaskList(name, tasks)`), and `carryOverChecklistMarks` explicitly drops items only present in the existing file. This surfaces via the queue-while-busy feature (#180) and the exhausted-source regeneration (#157) when the generation lists only new work.

## Expected

Generated-task confirm must be **append-only**: preserve every existing task (its order and its `[-]`/`[x]` progress marker) and append only genuinely-new tasks (deduped by identity). Never drop or reorder existing tasks.

## Fix

- `ensureGeneratedTaskFile` merges existing + new (existing first, new-by-identity appended), returning the merged list.
- The confirm+send reservation locates the first task's real position in the merged list (it is no longer necessarily item #1) and reserves the rendered text.
- Carry-over prefers the most-advanced mark when collapsing a duplicate identity.

0xGosu:
Closes #183.

## Problem

Confirming an LLM-generated-task escalation for an agent whose task list is the plugin's own bootstrapped generated file **replaced** the list with only the newly-suggested tasks — any existing task the new generation did not re-list was **dropped** (data loss). It surfaces via the queue-while-busy path (#180, PR #181) and the exhausted-source regeneration (#157) whenever the generation lists only new work.

Root cause: the bootstrap path in `acceptGeneratedTask` re-rendered the file from only the current suggestion, and `carryOverChecklistMarks` dropped items present only in the existing file.

## Fix — append-only

- `ensureGeneratedTaskFile` now merges existing + new under the file lock: existing tasks first (preserving order and their `[-]`/`[x]` markers), then only tasks not already present (deduped by identity) appended at the end. It returns the merged list. Nothing is dropped or reordered.
- The confirm+send reservation locates the first task's **real** position in the merged list (no longer necessarily #1) and reserves the same text the file was rendered from (the stripped identity) — so a task whose normalized text itself begins with `N. ` (an LLM doubly-numbered list) no longer fails the reservation spuriously after the escalation is claimed.
- `carryOverChecklistMarks` prefers the most-advanced mark when collapsing a duplicate identity, so a task already underway is never re-armed for a duplicate send.

## Tests

- Updated the two regeneration tests to the append semantics (existing kept in place, new appended at the end).
- New regression tests: bootstrap-append (the exact reported bug), busy-append, send-appends-and-reserves-new-item, numbered-first-task reservation, and duplicate-identity mark collapse.
- Full suite (`go test -tags "vectors cpu" ./...`), `go vet`, `gofmt`, `golangci-lint` all pass. Two code-review findings (reservation-text divergence; duplicate-mark collapse) were caught and fixed before this PR.

## Live verification

Rebuilt and hot-swapped the `dev` daemon inside herdr; drove a real haiku agent to a generated-task escalation and confirmed it without send — the bootstrapped list was created correctly with the new CLI message ("added the suggested task(s) to the agent's task list (not sent)"). Test workspace cleaned up afterward.

> Note: builds on the #180 feature already merged via PR #181.

### DIFF
--- a/internal/frontend/frontend.go
+++ b/internal/frontend/frontend.go
@@ -17,6 +17,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 	"time"
@@ -1116,6 +1117,21 @@ func mergeGeneratedTasks(existing string, generated []string) []string {
 	return merged
 }
 
+// checklistMarkRank ranks a checkbox mark by how far along the task is, so a
+// collapsed duplicate identity keeps its furthest-along state: done ("[x]" and
+// its parse variants X/+/*) outranks in-progress ("[-]"), which outranks
+// pending ("[ ]"). See ChecklistItem.Mark for the mark alphabet.
+func checklistMarkRank(mark string) int {
+	switch strings.TrimSpace(mark) {
+	case "":
+		return 0 // pending "[ ]"
+	case domain.MarkInProgress:
+		return 1 // in-progress "[-]"
+	default:
+		return 2 // done "[x]"/"[X]"/"[+]"/"[*]"
+	}
+}
+
 // generatedTaskPosition returns the 1-based position of task within merged
 // (matched by domain.GeneratedTaskIdentity, so a numbered or raw form both
 // find it), or 1 if absent — the reservation's own text check then fails loudly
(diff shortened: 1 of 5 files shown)


0xGosu:
Addressed both review findings in f175688:

- **Duplicate-marker collapse (frontend.go)** — replaced the pending-vs-non-pending prepend with an explicit `checklistMarkRank` (done `[x]` > in-progress `[-]` > pending `[ ]`); `carryOverChecklistMarks` now sorts each identity's marks and assigns the most-advanced. The regression test is table-driven over all orders, including `[x]` before `[-]` and `[-]` before `[x]`.
- **TUI decline path (from the #181 review)** — the busy-agent `add to the task list instead?` prompt pre-filled `y`, and the input box appends keystrokes, so typing `n` became `yn` (still `HasPrefix "y"`) — the decline was unreachable. The prompt now opens empty (hint `[y/N]`): a bare Enter cancels (leaves the escalation pending) and only an explicit `y` queues. Added decline-by-Enter and decline-by-typing-n tests.

Full suite (`go test -tags "vectors cpu" ./...`), vet, gofmt, and golangci-lint all pass.

PR #186, opened 2026-07-19
#1 fix: build match package without the vectors tag via KNN build-tag split

0xGosu:
## Problem

`internal/match/match.go` called bleve v2.6.0's `SearchRequest.AddKNNWithFilter` unconditionally, but bleve defines that method **only** under its `//go:build vectors` constraint (`search_knn.go`; the `!vectors` build gets `search_no_knn.go`). So the `match` package — and therefore any tag-free `go build ./...` / `go vet` — failed to compile without the `vectors` tag:

```
internal/match/match.go:219:6: req.AddKNNWithFilter undefined
    (type *bleve.SearchRequest has no field or method AddKNNWithFilter)
```

This is the only symbol that broke the tag-free build; the rest of the tree already compiles without `vectors`.

## Fix

Split the KNN search call into build-tagged helpers, mirroring bleve's own `search_knn.go` / `search_no_knn.go` pattern:

- **`knn_vectors.go`** (`//go:build vectors`) — runs the real `AddKNNWithFilter` KNN query.
- **`knn_novectors.go`** (`//go:build !vectors`) — stub returning "vector matching unavailable", so callers fall back to BM25 text matching.

`MatchVector` short-circuits with a clear *"built without the vectors tag"* error when `vectorSearchSupported` is false.

The daemon's `resolveSignature` now clears `vec` on **any** `MatchVector` error so the `vec == nil` BM25 fallback actually runs — previously the *"trying text match"* log was empty and resolution silently dropped to hash-only.

## Tests

- Vector cases moved behind `//go:build vectors`; text/BM25 cases stay untagged so they run in **both** configs (validating the BM25 fallback path independently).
- Adds **`TestMatchVectorFilteredKNNExcludesNearerOutOfScope`** — a regression test that crowds the query with four out-of-scope neighbors (more than `matchK=3`) and one far in-scope row, so a dropped or unfiltered KNN filter fails loudly.

Verified:
- `go build ./...` and `go vet ./internal/match/` now pass **without** the `vectors` tag.
- `go build -tags "vectors cpu" ./...` passes.
- `go test -tags "vectors cpu" ./internal/match/ ./internal/daemon/` — 124 tests pass (8 match + 116 daemon).
- `go test ./internal/match/` (no tags) — BM25 tests pass; vector tests correctly gated out.

### DIFF
--- a/internal/daemon/semantic.go
+++ b/internal/daemon/semantic.go
@@ -116,7 +116,12 @@ func (d *Daemon) resolveSignature(ctx context.Context, cfg config.Config,
 			accept := func(h match.Hit) bool { return remapAllowed(s, sig, h) }
 			hit, ok, err := d.matcher.MatchVector(ctx, vec, scope, accept)
 			if err != nil {
+				// Fall back to BM25 below: clear vec so the vec==nil text path
+				// runs (e.g. a degraded latch, or a text-only build where KNN is
+				// unavailable). Without this the "trying text match" promise is
+				// empty and resolution drops to hash-only.
 				slog.Warn("vector match failed; trying text match", "error", err)
+				vec = nil
 			} else if ok && hit.Score >= cfg.Embedding.SimilarityThreshold {
 				slog.Info("semantic match: reusing learned signature",
 					"signature", hit.Signature, "cosine", hit.Score, "raw", sig.Raw)
--- a/internal/match/knn_novectors.go
+++ b/internal/match/knn_novectors.go
@@ -1,0 +1,23 @@
+//go:build !vectors
+
+package match
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/blevesearch/bleve/v2"
+	"github.com/blevesearch/bleve/v2/search/query"
+)
+
+// vectorSearchSupported is false in text-only builds (no `vectors` tag): bleve
+// is linked without its KNN engine, so MatchVector degrades to unavailable and
+// callers fall back to BM25 (MatchText). See knn_vectors.go for the real path.
+const vectorSearchSupported = false
+
+// knnSearch reports that vector search is unavailable in this build. bleve's
+// SearchRequest.AddKNNWithFilter exists only under the `vectors` tag, so a
+// text-only build cannot run KNN; the daemon falls back to BM25 matching.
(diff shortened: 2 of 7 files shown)


0xGosu:
Thanks — confirmed valid and fixed in 661c7dc.

**Root cause (as reported):** the `vectorSearchSupported` guard only made `MatchVector` callable without `vectors`; the index-**build** path was still unsafe. `Matcher.Rebuild` forwards the embedder's dims to `buildIndex`, which on `dims>0` calls `mapping.NewVectorFieldMapping()` and assigns its fields — and bleve v2.6.0's `mapping_no_vectors.go` returns `nil` there, so a no-`vectors` daemon with an embedder present (reachable via the `cpu` tag alone, now that #187 gates the embedder on `cpu`) nil-panicked in `initSemantic`; `semanticReady` never flipped and BM25 was never reached.

**Fix:** `Rebuild` forces `dims = 0` when `!vectorSearchSupported`, so the index is built text-only (no vector field mapping, no vector docs) and matching degrades to BM25. This composes with the existing `resolveSignature` `vec = nil` reset (already in this PR): with an embedder present, the embed still runs, `MatchVector` returns its unavailable error, and resolution falls through to `MatchText` (BM25).

**Test:** added `TestRebuildTextOnlyWithoutVectors` (`//go:build !vectors`) — `Rebuild` with `dims=3` and real vectors must not panic, `MatchText` (BM25) must resolve, and `MatchVector` must report unavailable. This exercises the tag-free runtime path the `vectors cpu`-only CI does not cover.

Verified: `go test ./internal/match/` (tag-free) — 3 pass incl. the new test, no panic; `go test -tags "vectors cpu" ./internal/match/` — 8 pass (incl. the KNN pre-filter regression), no behavior change in the vectors build.

Note on the requested daemon-level test: the defect and its fix live entirely in `Matcher.Rebuild`/`buildIndex`; the daemon path just calls `Rebuild`. The matcher-level `!vectors` test reproduces the exact panic (embedder dims → `Rebuild(dims>0)`) and the BM25 resolution without needing a real gguf model in the tag-free job. Happy to add a full `!vectors` daemon+fake-embedder test if you'd prefer that coverage too.

PR #196, opened 2026-07-20
#1 fix: dedup resolved-escalation re-fires and near-identical captures

0xGosu:
## Summary

Fixes duplicate escalations that appeared after an operator had already handled one, and adds the README docs requested alongside.

### Escalation de-duplication (the fix)
Two independent gaps, each behind a new `[limits]` config knob (TUI/CLI-editable, documented in `sample/config.toml`):

- **Recently-resolved window** — `escalation_dedup_window_seconds` (default `300`). Once an escalation was accepted it flipped `escalated → resolved` and left the dedup candidate set, so herdr's delayed re-delivery of the same screen re-escalated as a duplicate. The candidate set now also includes recently-resolved escalations **whose answer was actually delivered** (`corrections.sent = 1`).
  - **Why `sent = 1` is load-bearing:** a learn-only shadow confirm (`hap confirm`, `send=false`) also resolves the escalation but delivers nothing, so the agent stays blocked and *must* re-escalate to accumulate confirmations toward graduation. Keying the window on a delivered answer suppresses genuine stale duplicates while leaving the shadow-learning loop intact.

- **Content jitter tolerance** — `escalation_dedup_jitter_percent` (default `5`, `0` = exact-only). Near-identical captures now dedup via character-trigram (shingle) Jaccard similarity (`domain.SimilarWithin`), applied **only to large tail-window captures** behind the existing `firstLineExplained` guard, so two distinct commands sharing a large identical body can never silently collapse (the "never a silent drop" rule).

### README (docs)
- Quickstart **"Update to the latest version"** — `herdr plugin install … --yes` + `hap daemon --ensure`, plus the optional clean reinstall.
- **Roadmap** — Full-Self Prompting (FSP), OpenCode support, HAP Cloud, HAP Web (remote control).

## Testing
- Full suite green: `go test -tags "vectors cpu" ./internal/...` — 1,526 tests across 22 packages, 0 failures.
- New/updated coverage: store window (sent vs learn-only vs out-of-window), `SimilarWithin`, jitter gating (incl. the first-line safety case), config defaults/validation + field-registry parity, and an end-to-end `TestPipelineDedupAfterResolveWithinWindow`. `TestConfirmDrivenShadowToAutoPromotion` (shadow graduation) still passes.
- `gofmt` clean, `go vet` clean, `golangci-lint` 0 issues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,28 @@ herdr plugin install 0xGosu/herdr-auto-pilot --ref v0.4.0
 herdr plugin install 0xGosu/herdr-auto-pilot --yes
 ```
 
+### Update to the latest version
+
+Re-run the install command to upgrade an existing install to the newest
+release. `--yes` skips the interactive confirmation.
+
+```sh
+herdr plugin install 0xGosu/herdr-auto-pilot --yes   # download & install the latest release
+hap daemon --ensure                                  # recommended: hot-swap the running daemon now
+```
+
+`hap daemon --ensure` is optional but recommended — it swaps the running
+daemon for the new build immediately, so you don't have to wait for the next
+automatic restart.
+
+For a clean reinstall, uninstall first, then install:
+
+```sh
+herdr plugin uninstall herd-auto-prompter            # optional: only for a clean reinstall
+herdr plugin install 0xGosu/herdr-auto-pilot --yes
+hap daemon --ensure
+```
+
 The monitoring daemon starts automatically when an agent appears in the herd.
 Use the following recommended setup to make the **Auto Prompter** pane (TUI)
 and its CLI convenient to access from the host machine.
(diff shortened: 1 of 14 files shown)


0xGosu:
Thanks for the thorough review — all five findings were valid and are fixed in `6b14f4a`:

1. **Window measured from raise time, not resolution time** — the resolved branch now keys the window on the delivered correction's timestamp (`corrections.created_at >= resolvedSince`), so an escalation that sat pending past the window before an away operator answered is still recognized right after it's delivered.
2. **Successful deliveries not setting `sent=1`** — both generated-task confirm paths now `MarkCorrectionSent` after a successful send (guarded so a rolled-back/failed send never marks). Safe because idle is not a `verifyunblock` situation, so no self-check is armed.
3. **128-row limit hiding an old pending row** — pending and recently-resolved candidates are now fetched as two separate queries with independent limits, so pending rows are never crowded out.
4. **`firstLineExplained` fooled by shared content** — the order-insensitive jitter path now requires the first lines to be *identical* (`firstLinesEqual`), not merely substring-explained, so two distinct approval headers appearing in each other's scrollback can't collapse. The suffix path keeps `firstLineExplained` (it's order-sensitive/exact).
5. **Resolved branch not restricted to escalations** — the resolved query filters `action_or_escalation = 'escalated'`, so a post-hoc `hap resolve --send` of an autonomous action can't masquerade as a recently-resolved escalation. Delivery-failed escalations (auto-prefixed action) fall out too and simply re-escalate a redundant ask — the safe direction.

Every fix fails toward "re-escalate a redundant ask" rather than suppressing a real escalation. New/updated store, domain, and daemon tests cover each; full `./internal/...` suite is green.

PR #210, opened 2026-07-21
#1 refactor: print task-management instructions in `hap task <agent> list`

0xGosu:
## What

The next-task prompt carried the full task lifecycle in every send (`start <n>`, `done <n>`, how `<n>` is addressed, the `--path` fallback). This moves that text into the output of `hap task <agent> list`, which the prompt now simply points at — stated once, beside the real task numbers.

Listing output:

```
#1	[ ]	3.1 wire the logger
#2	[-]	3.2 add metrics
2 task(s): 1 pending, 1 in-progress, 0 done

Prefer using the hap CLI to manage your tasks:
- `hap task happy-pelican start <n>` to mark one in-progress when you begin working on it.
- `hap task happy-pelican done <n>` to mark it complete as you go.
Note:
- `<n>` is the task's own id when the list numbers its tasks (e.g. `done 3.1`); otherwise its position in the list, which `'#3'` always addresses (quote it — a bare #3 is a shell comment).
- when the agent name `happy-pelican` is no longer recognized, use `--path /…/happy-pelican.md` in place of `happy-pelican`
```

A `--path`-addressed list spells every command with the path and drops the last note (there is no name to fall back from).

New default prompt template:

> Your next task is {next_task_content}. Prefer the hap CLI to manage your tasks (start/done), run bash `hap task {agent_name} list` to view them (if that name isn't recognized, use `--path {task_list_path}` in place of `{agent_name}`).

## Changes

- `domain.TaskManagementHints` renders the block; `internal/cli` prints it under `list` **only** — mutating ops reprint the list without repeating it.
- `frontend.App.TaskFilePath` exports the existing resolver, so the listing and the printed path resolve exactly once, together.
- Two shell-correctness fixes surfaced in review: `'#3'` is quoted (bash strips a bare `#3` as a comment, so `done #3` reached hap with no ref), and paths go through a new `domain.ShellQuote` so a checklist under a directory with a space stays one argument.
- Fixes two pre-existing CLI assertions the new output would have silently defanged — `Contains(out, "#3")` now matches whole rows.
- Docs: `README.md`, `.claude/skills/hap/SKILL.md`. The README also gains an optional "Hotkeys for pause / resume / status" section (docs-only, bundled from the same session).

## Tests

- `TestTaskManagementHints` / `…PathOnly` golden tests, `TestShellQuote` table test.
- `TestTaskListPrintsManagementHints` (agent-addressed vs `--path`, plus mutating ops must not repeat the block), `TestTaskListHintsQuotePathWithSpace`.
- Full suite `go test -tags "vectors cpu" ./... -count=1` passes; the only failure seen was the known flaky `TestEmbedTextRealModel` 2s embed stall guard under load, which passes in isolation and is unrelated.

### DIFF
--- a/.claude/skills/hap/SKILL.md
+++ b/.claude/skills/hap/SKILL.md
@@ -592,7 +592,7 @@ agent at its own list with its name pre-filled (and a `--path` fallback for
 sources that aren't name-addressable):
 
 ```
-Your next task is {next_task_content}. Prefer the hap CLI to manage your tasks (start/done), run bash `hap task {agent_name} list` to view them (if that name isn't recognized, use `--path {task_list_path}` in place of `{agent_name}`).
+Your next task is {next_task_content}. Prefer the hap CLI to manage your tasks (start/done), run bash `hap task {agent_name} list` to view them (if that name isn't recognized, use `--path {task_list_path_quoted}` in place of `{agent_name}`).
 ```
 
 the lifecycle instructions themselves (`start <n>`, `done <n>`, how `<n>` is
@@ -602,6 +602,7 @@ being re-sent with every prompt.
 
 - `{next_task_content}` — the text of the next unchecked item (or `"none"` when the list is complete)
 - `{task_list_path}` — absolute path to the checklist file
+- `{task_list_path_quoted}` — that path as one shell word (quoted only when it needs it); use this, not `{task_list_path}`, inside a command the agent is meant to run
 - `{agent_name}` — the agent's hap-owned short name
 - `{cwd}` — the agent's working directory (the project it is in)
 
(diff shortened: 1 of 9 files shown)


0xGosu:
Fixed in b0fe4b1 — the finding was valid.

`DeclaredTask.Prompt` substituted the raw path, so the prompt's `--path {task_list_path}` fallback stayed unquoted while the listing's hints were quoted: a checklist under a directory with a space split into two arguments exactly where the agent had no name left to fall back on.

- Added a `{task_list_path_quoted}` placeholder — the path as one shell word, via the same `domain.ShellQuote` the hints use — and switched `DefaultNextTaskTemplate` to it. `{task_list_path}` stays raw for templates that put the path in prose.
- The quoted key is substituted first: `strings.NewReplacer` matches in argument order, so the shorter `{task_list_path}` would otherwise consume its prefix and leave a stray `_quoted` in the prompt.
- Prompt-level regression tests: a default-template render with `/my docs/tasks.md`, and a custom template using both placeholders side by side.
- Placeholder docs updated in `config.go`, `sample/config.toml`, `README.md`, the hap skill, and the `task-source --template` flag help.

PR #222, opened 2026-07-22
#1 fix: stop auto_send_when_idle stalling under the runaway guard; harden config

0xGosu:
## Summary

Fixes `auto_send_when_idle` silently stopping after a while, plus three requested config-hardening changes.

### 1. Bug: `auto_send_when_idle` stalls under the runaway-loop guard (FR-019)

The idle poll re-drives agents through the same pipeline as normal events, including the runaway-loop guard. Every auto-send climbs `AgentRate.ConsecutiveAuto`, which resets **only** on genuine human interaction. Because the feature drives the agent unattended (it starts "working" within 10s of each send — correctly *not* counted as human), the counter climbs by 1 per task and never resets. After `max_consecutive_auto_prompts` sends the agent hits the ceiling, escalates `rate_limited`, and gets **permanently paused** — after which every idle sweep skips it. With one agent on the source the feature goes silent.

**Fix (two parts):**
- Idle hand-outs advance only the per-minute window, not the consecutive counter (`domain.RegisterAutoPromptIdle`), scoped by the `Reserve` flag so both the poll and the same source's learned-rule idle sends are covered.
- `escalate()` no longer pauses an unattended idle send on a rate-limit — otherwise the permanent stall just moved to the (lower) per-minute cap, since `escalate` pauses on *any* `ReasonRateLimited`. The per-minute window self-heals, so the send defers to a later sweep. Reply-loop (non-idle) sends still pause exactly as before, and the noop paths still climb the counter (D3 self-flap protection).

### 2. `max_consecutive_auto_prompts` default 10 → 30

Config default plus every doc that stated it.

### 3. Escalation dedup tuning → hard constants

`escalation_dedup_window_seconds` and `escalation_dedup_jitter_percent` removed from config; now fixed daemon constants (`escalationDedupWindow = 5m`, `escalationDedupJitterPercent = 5` — identical to the old defaults, behavior unchanged).

### 4. `embedding.max_consecutive_failures` → hard constant

Removed from config; now the fixed constant `embedder.DefaultMaxConsecutiveFailures`, bumped **3 → 5**. Also removed the now-dead `ResolveMaxFailures` resolver, the `int32` failure ceiling, and the `EnvWorkerMaxFailures` worker-env forwarding (parent and worker compile in the same constant).

## Backward compatibility

All three retired keys are **ignored-with-a-warning** on `Load` and **dropped on `Save`**, matching the existing convention (`verify_unblock_ms`, `gpu_layers`). An existing `config.toml` carrying any of them loads cleanly.

## Testing

- New regression tests: runaway-counter exemption (`domain`), no-pause-on-idle-rate-limit (`daemon`), fixed-ceiling embedder latch (`embedder`), and retired-key migration (`config`).
- Full suites green across `domain`, `daemon`, `config`, `frontend`, `cli`, `embedder` (both `vectors cpu` and tag-free builds). `gofmt` / `go vet` clean.
- Two independent code reviews per change; all actionable findings addressed.

## Note for reviewers

Raising the consecutive ceiling to 30 means a genuinely runaway **non-idle** reply loop now auto-answers up to 30 times (still bounded to ~6 min by the 5/min guard) before pausing for a human — a 3× longer unattended loop. Idle hand-outs are exempt from that counter entirely. Flag if you'd rather keep the reply-loop ceiling lower.

https://claude.ai/code/session_019oaepE4739f2yCZTtE4HuL


0xGosu:
Thanks for the review — both blocking findings were valid and are fixed in af90ad6.

**Finding 1 (admission gate).** Correct: exempting only the counter *increment* left `domain.CheckRate` still *rejecting* an idle hand-out once `ConsecutiveAuto` was saturated by non-idle reply-loop sends, and since idle escalations no longer pause, it never reset → permanent stall. `CheckRate` now takes an `idleHandout bool` that skips **only** the consecutive ceiling; `Paused` and the per-minute cap still gate. `Decide` computes it from the verified classified situation + resolved action (`SituationIdle && candidate == ActionNextDeclaredTask && DeclaredTask.Reserve`); the three daemon gates pass `res.request.ReserveTask` on the delivery/promotion paths (promotion also `&& !isNoop`) and `false` on the noop gate (keeps the D3 ceiling). Scoped to reserving (`auto_send_when_idle`) sources only, so a non-reserving idle declared task still rate-limits.

**Finding 2 (stale flag).** Correct: `tr.AutoIdleSend` is set at sweep-schedule time, before the async capture. The exemption/no-pause helpers now key off verified state — `isIdleTaskHandout(sit, declared)` requires `SituationIdle` + `Reserve`, and `isUnattendedIdleSend(ctx, s)` requires `s.Type == SituationIdle` + a reserving source — instead of the stale flag. A pane that turned into a real approval between the sweep and capture is now gated and paused normally.

One thing I deliberately did **not** do: the FR-008 confidence gate stays enforced for idle hand-outs (only the FR-019 consecutive ceiling is relaxed) — an idle signature must still be graduated before it auto-sends, so I fixed the failing regression's setup (missing autonomous history) rather than bypassing the confidence gate.

New regressions: saturated consecutive counter at the `CheckRate`, `Decide`, and daemon end-to-end levels; stale-flag guard on both helpers.

Doc note (5 vs 3 embed failures) was already updated in the prior commit.

https://claude.ai/code/session_019oaepE4739f2yCZTtE4HuL

PR #228, opened 2026-07-23
#1 feat: per-command environment variables for the LLM CLIs

0xGosu:
## What

Each of the four LLM command templates can now be spawned with its own environment, so one CLI can run against a different key, provider, model, or proxy than another.

| Key | Scope |
|---|---|
| `llm.env` / `llm.env_file` | shared by all four commands |
| `llm.command_env` / `llm.command_env_file` | `llm.command` |
| `llm.command_start_env` / `llm.command_start_env_file` | `llm.command_start` |
| `llm.task_generate_command_env` / `…_env_file` | `llm.task_generate_command` |
| `llm.task_generate_command_start_env` / `…_env_file` | `llm.task_generate_command_start` |

```toml
[llm]
env_file = "~/.config/hap/llm.env"
task_generate_command_env_file = "/path/to/claude_task_generate.env"

[llm.command_env]
ANTHROPIC_MODEL = "opus"
[llm.task_generate_command_env]
ANTHROPIC_MODEL = "haiku"
```

Layering, last wins: daemon env → `env_file` → `env` → the command's `…_env_file` → its `…_env` → hap's own `HAP_*` (always authoritative). Values accept the command template's placeholders except `{pane_excerpt}` — untrusted, unbounded pane text is never put in a child's environment.

## Secrets

- Env files are read when the CLI is **spawned**, never at config load, so `config.Save` can never copy their contents into `config.toml`, and editing a file applies to the next run with no restart.
- No value reaches an operator surface: `hap config` prints variable **names** and file paths only; the TUI/`config fields` registry carries only the `.env` **paths**; parse errors name a file and line number, never content.
- `HAP_*` / `HERDR_*` names are reserved and ignored with a warning — an operator env can't redirect the MCP wiring or point a nested `hap`/`herdr` at another installation.

## Fail-safe

A configured env file that cannot be read, has a malformed line (including an unterminated quote), or defines nothing at all **fails that run** — the daemon escalates instead of launching the CLI without its credentials, which would otherwise surface much later as an opaque auth error. The CLI is not spawned in that case.

The env travels with its template through the fast-fail retry, so a rescue run never gets the other command's key; and two templates that differ *only* in environment now still qualify for that retry.

## Tests

New: `internal/llm/env_test.go` (parser table + `buildEnv` precedence, reserved keys, placeholder rules, error paths), `internal/llm/commandenv_test.go` (real subprocesses asserting each template gets its own env). Extended: config Save/Load round-trip and defaults, `hap config` masking, a frontend guard that no registry field can render an env value.

Full suite green (`go test -tags "vectors cpu" ./...`), `go vet` and `golangci-lint` clean. One daemon test (`TestAutoSendIdleBelowThreshold`) flaked once under full-suite load and passed in isolation and on re-runs — pre-existing, unrelated to this change.

### DIFF
--- a/.claude/skills/hap/SKILL.md
+++ b/.claude/skills/hap/SKILL.md
@@ -381,6 +381,13 @@ edits made through `hap config set` / `set-threshold` apply live — the command
 | `llm.task_generate_command` | (unset) | argv template for the one-shot task suggestion given to an idle agent with NO task source, or a declared source that ran out (see task sources); empty keeps escalate-only behavior |
 | `llm.task_generate_command_start` | (unset) | argv template for the FIRST task generation per agent (no-source case only; empty inherits `llm.task_generate_command`); an exhausted declared source only generates more tasks when BOTH this and `llm.task_generate_command` are set, and always uses `llm.task_generate_command` (never this) since a list already exists |
 | `llm.task_generate_timeout_seconds` | 0 (inherits `timeout_seconds`) | timeout for one task-generation run |
+| `llm.env` / `llm.env_file` | (unset) | environment shared by all four llm commands: an inline `[llm.env]` table, and/or a path to a `.env` file (`KEY=VALUE`, `#` comments, `export`, quotes, `~/` expanded) |
+| `llm.command_env` / `llm.command_env_file` | (unset) | environment for `llm.command` only, layered over the shared one |
+| `llm.command_start_env` / `llm.command_start_env_file` | (unset) | environment for `llm.command_start` only |
+| `llm.task_generate_command_env` / `llm.task_generate_command_env_file` | (unset) | environment for `llm.task_generate_command` only — e.g. a cheaper model or a different key for task ideas |
+| `llm.task_generate_command_start_env` / `llm.task_generate_command_start_env_file` | (unset) | environment for `llm.task_generate_command_start` only |
+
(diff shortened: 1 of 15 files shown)


0xGosu:
Both review findings were valid and are fixed in cc8dfd0.

**Per-command `PATH` not honored** — confirmed, and worse than the ordering alone: `exec.Command` resolves a bare name against the *current process's* PATH regardless of `cmd.Env`, so reordering was necessary but not sufficient. Now the environment is composed first, `preflight` resolves the command against the **child's** PATH (`lookPathIn`), and the resolved absolute path is handed to `exec` so no second lookup happens. Both the consult and task-generation paths. A name containing a separator, or a relative PATH entry (which would resolve against the child's working dir), still falls back to the previous behaviour rather than guessing.

Tests: a bare CLI reachable only via a configured `PATH` now runs (consult + task-gen), and one on the daemon's PATH but absent from the configured one fails the preflight instead of dying opaquely at spawn.

**Inline keys bypass `validEnvKey`** — confirmed. `[llm.command_env]` with a quoted key like `"A=B"` was emitted verbatim and re-split by `envSet.merge` at the first `=`, setting `A` to `B=<value>`. Inline names are now validated exactly like file ones and an invalid one fails the run; the error names the key (safe) and never the value. Test covers `A=B`, a name with a space, a leading digit, and the empty key.

Full suite, `go vet`, and `golangci-lint` are clean.

PR #232, opened 2026-07-24
#1 feat: semantic (embedding) search for signatures — TUI & CLI

0xGosu:
## What

Adds **search** for learned rules/signatures in both surfaces, with a new **semantic (embedding)** mode alongside keyword search.

- **CLI** — new `hap signatures search <query>`:
  - keyword by default (case-insensitive substring over the rule's fields + its salient text)
  - `--semantic` embeds the whole query and ranks rules by cosine similarity
  - `--limit` / `--min-score` bound the semantic set; the existing `--type` / `--mode` / `--agent-type` / `--min-conf` filters compose with either mode
  - semantic results show a `sem=` score column
- **TUI (Rules tab)** — `/` keeps the live keyword filter; typing a **2+-word** query surfaces a footer hint to press **Enter** for a semantic search. Results re-rank the list with a `SEM` cosine column and a match-count message; editing the query returns to live keyword filtering.

## How

`App.SearchSignatures` (new, `internal/frontend/signaturesearch.go`) reuses the enriched rows from `App.Signatures` and joins the salient text/vectors from `signature_embeddings`. Semantic mode builds a **standalone** embedder (mirroring `ReembedStandalone` — the control socket carries no reply channel, so front-ends embed directly) and ranks by cosine via a plain linear scan (vectors are L2-normalized, so cosine is a dot product) — no daemon and no `vectors` tag required, only the embedding model.

Recall-oriented by design: top-N above a lenient floor (0.3), scores shown. Degrades cleanly — drifted / dimension-mismatched vectors are skipped, and a disabled/missing model returns a clear error (never a partial ranking); keyword search is unaffected.

## Tests

New unit tests in `internal/frontend`, `internal/cli`, and `internal/tui` covering keyword substring + filter composition, cosine ranking/floor/limit, model-drift & dimension-mismatch skips, clean degradation, the TUI 2-word hint, end-to-end dispatch, single-word/edit fallbacks, and stale-result teardown. Full changed-package suite green (403 tests); `gofmt`, `go vet`, and `golangci-lint` clean with `-tags "vectors,cpu"`.

## Notes

- Semantic search loads the embedding model per call (acceptable for an explicit on-demand action; mirrors `ReembedStandalone`).
- Reviewed via the code-review subagent; findings addressed. One review suggestion (letting `--min-score 0` mean "no floor") was intentionally declined — it would make the zero value a footgun; `0`/unset uses the default floor instead.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.claude/skills/hap/SKILL.md
+++ b/.claude/skills/hap/SKILL.md
@@ -260,6 +260,21 @@ hap signatures list --agent-type claude       # by agent type
 hap signatures list --min-conf 0.85          # by minimum cached confidence
 ```
 
+search learned signatures — keyword (case-insensitive substring over the rule's fields
+and its salient text) by default, or `--semantic` to embed the whole query and rank rules
+by meaning (cosine similarity; needs the embedding model). the same `--type` / `--mode` /
+`--agent-type` / `--min-conf` filters compose with either:
+
+```bash
+hap signatures search terraform                       # keyword substring
+hap signatures search "approve the file write" --semantic   # ranked by meaning
+hap signatures search "run tests" --semantic --limit 10 --min-score 0.5
+```
+
+the TUI mirrors this on the Rules tab: `/` starts a live keyword filter, and typing a
+2+-word query surfaces a hint to press **enter** for a semantic search (results show a
+`SEM` cosine column; editing the query returns to live keyword filtering).
+
 show full detail for a signature (recent decisions, last audit context):
 
 ```bash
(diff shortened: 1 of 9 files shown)


0xGosu:
Thanks — good catch. Fixed in c2c1654: the `semanticSearchMsg` error path now clears `m.sigSemantic` so a failed rerun of the same query can't keep rendering the prior ranked rows/SEM scores alongside the error. Added a regression test `TestSemanticSearchFailedRerunClearsStaleResults` covering exactly that repro (active search → rerun errors → results invalidated). TUI suite green.

0xGosu:
@charliecreates the blocking issue you flagged (stale semantic results surviving a failed rerun, tui.go error path) is fixed in c2c1654 — `m.sigSemantic` is now invalidated on the error path, with a regression test `TestSemanticSearchFailedRerunClearsStaleResults`. CI is green. Could you re-review when you get a chance?

PR #245, opened 2026-07-25
#1 feat: reorder a task within its checklist from the TUI and CLI

0xGosu:
> **Stacked on #243** — base that branch, not `main`. The reorder must carry a task's detail block, which is `detailBlockEnd` from #243; a reorder that moved only the title line would reintroduce that bug in its worst form. Retarget to `main` once #243 merges.

## What

The Tasks tab could add, edit, complete and delete a task but not **move** one, so fixing the order of a hand-authored plan meant editing the file behind hap's back.

| Layer | Added |
|---|---|
| domain | `MoveChecklistItem(content, from, to)` |
| frontend | `App.MoveTask(agent, path, index, to, expectText...)` |
| CLI | `hap task <agent> move <n> <position\|up\|down>` (aliases `mv`, `reorder`) |
| TUI | `K`/`J` and `shift+up`/`shift+down` on the Tasks tab, plus footer hint |

The CLI verb is not optional — `hap help tui` states that everything the TUI does is available as a CLI verb, which is what scripts and agents use.

Verified live in a running TUI: repeated `J` walks a task down and the cursor follows it, each detail block travelling with its own task.

## Detail travels with the task

`MoveChecklistItem` moves the item's whole block via `detailBlockEnd`, exactly as delete and append do since #243. Moving only the title would strand the detail to be folded into whatever ends up above it while the moved task arrives bare — reordering is precisely the operation that would otherwise scramble *every* task's instructions at once, rather than one.

`TestMoveChecklistItemKeepsEveryTasksDetail` walks all nine ordered pairs of a three-task list and asserts every task still folds to exactly its own content.

## Reordering is siblings-only, and that is deliberate

A nested `- [ ]` is its own **item**, not detail, so `detailBlockEnd` stops at it — meaning it neither travels with a parent nor is stepped over. Without a guard:

```
- [ ] a          move a (1→2)      - [ ] a1     ← orphaned, no parent
  - [ ] a1            →            - [ ] a
- [ ] b                            - [ ] b

- [ ] a          move b (3→2)      - [ ] a
  - [ ] a1            →            - [ ] b
    - a1 detail                      - [ ] a1   ← adopted by b
- [ ] b                              - a1 detail
```

Positions are right and no line is lost, but the tree the operator wrote gets rewritten. Both cases are now refused by name:

- an item with nested sub-tasks cannot move (its children would not follow);
- source and destination must sit at the same indent depth.

Subtree moves were considered and rejected: they make `to` ambiguous (does the destination position count the moved children?). Flat lists — what hap writes, and what nearly every checklist is — are unaffected, and siblings at the same depth reorder normally.

## TUI specifics

Three things this action needs that no other Tasks action does:

- **`m.taskMarks` are keyed positionally** (`taskMarkKey(group, item)`), so a reorder renumbers items under them and every surviving mark would silently retarget a different task. Cleared, as the delete path does.
- **The cursor is a row index, not an item identity**, and a reorder leaves the row *count* unchanged — so `clampListViewport` does nothing and the cursor would land on whichever task swapped into place. It is nudged by delta so it follows the moved task.
- **Refused while a search filter is active.** Positions count hidden tasks, so a step would jump the task over items the operator cannot see; worse, a moved task often keeps its filtered row while its file position changes, so the nudge would land on a different task and the next keypress would move *that* one.

Both ends **refuse rather than clamp**: a clamp would rewrite the file with identical content and report success, so holding the key at the top would look like it was still working. Same for `move 2 2` on the CLI, which short-circuits before the mutation (`taskfile.Mutate` writes unconditionally).

The cursor nudge anticipates the async refresh rather than replacing it — a second press inside that window still reads stale rows and is refused by the expected-text guard. The file is never at risk; one move per refresh is the honest cadence, and the comment says so.

## Tests

- domain: both directions and distances, all nine ordered pairs with fold assertions, marks preserved, prose/headers untouched, no trailing newline, both error classes, the nesting refusals (and that siblings at the same depth still work), and blank-line behaviour pinned with a line-count invariant
- CLI: the reorder, detail travelling, relative vs absolute destinations, both edge no-ops leaving the file untouched, invalid destinations (`sideways`, `+2`, `""`, `2.5`), the alias, id-addressed sources (`move 1.3 up`), and the nesting refusals
- TUI: `K`/`J` with the cursor following, both edges refusing without a write, header rows, marks cleared, shift-arrow aliases, footer hint, the filter refusal, and the cursor nudge with **two** task-source groups

Full suite green (24 packages), `internal/domain` clean under `-race`, `gofmt`/`go vet`/`golangci-lint --build-tags "vectors,cpu"` clean.

## Known, documented behaviour

Blank separators do **not** travel with a moved item — they sit between items rather than belonging to one, so a move can leave a blank where the item was and land it tight against its new neighbour. Carrying them just trades one spacing artifact for another (the blank would follow the item to the end of the file). Pinned by `TestMoveChecklistItemBlankLines` so it is a decision, not a surprise.

### DIFF
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -13,6 +13,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode"
 
 	"golang.org/x/term"
 
@@ -1490,7 +1491,7 @@ func taskHints(agent, path string, listing bool) []Hint {
 
 // taskOp runs one checklist operation against an already-resolved target.
 func taskOp(ctx context.Context, app *frontend.App, out io.Writer, agent, path string, args []string) error {
-	usage := "usage: task [<agent> | --path <file>] list [--status all|pending|done] | get <n> | add <text> | start <n> | done <n> | undone <n> | update <n> <text> | remove <n> | send <n> [--yes]\n<n> is a task id from the list (e.g. 3.4), or #<position>\nsee: hap help task"
+	usage := "usage: task [<agent> | --path <file>] list [--status all|pending|done] | get <n> | add <text> | start <n> | done <n> | undone <n> | update <n> <text> | remove <n> | move <n> <position|up|down> | send <n> [--yes]\n<n> is a task id from the list (e.g. 3.4), or #<position>\nsee: hap help task"
 	if agent == "" && path == "" {
 		return fmt.Errorf("%s", usage)
 	}
@@ -1567,6 +1568,8 @@ func taskOp(ctx context.Context, app *frontend.App, out io.Writer, agent, path s
 		fmt.Fprintf(out, "removed task #%d\n", idx)
 		printTaskList(out, items, "all")
 		return nil
+	case "move", "mv", "reorder":
+		return taskMove(app, out, agent, path, rest)
 	case "send":
 		return taskSend(ctx, app, out, agent, path, rest)
 	}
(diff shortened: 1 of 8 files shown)


0xGosu:
Valid finding — reproduced it before fixing. The guard compared indentation **depth**, and equal depth is not siblinghood: two sub-tasks under different parents are indented identically, so moving one onto the other's position passed the check and reparented it, leaving the first parent childless.

Fixed in bde5afd by comparing actual **parents** — `checklistParent` walks back to the nearest item indented strictly less, and two items are siblings exactly when that agrees. This subsumes the depth check: the cases already refused (moving into another item's nesting) have differing parents too.

Coverage added as requested — equal-depth children under different parents, at both domain and CLI level, plus the same shape one level deeper so the parent lookup is not special-cased to top level, and a positive case pinning that true siblings under a shared parent still reorder.

Verified live:

```
$ hap task --path cousins.md move '#2' 4
error: task #2 and position #4 are under different parents — a task can only be reordered among its siblings
```

The branch has also been rebased onto current `main` (both #242 and #243 have since merged), so this PR now targets `main` directly.

PR #265, opened 2026-07-28
#1 feat: auto-accept escalations that have waited past their threshold

0xGosu:
## Why

`hap` escalates whenever it cannot act autonomously — shadow mode, a score under
threshold, a rule one confirmation short of graduating. The escalation lands in
`audit_log` and **waits forever**: the only exits are human (`Confirm`,
`Dismiss`, `PruneEscalations` — and the last only destroys, never accepts).

So the plugin's autonomy is bounded by operator attendance. A herd left running
overnight parks every gated-but-answered situation — including ones whose
suggestion hap generated itself — and each of those agents stays blocked until
somebody comes back.

This adds the missing third option: an escalation that has waited past a
threshold, and whose situation is still demonstrably on screen, is answered
automatically. The queue becomes a slow lane instead of a hard stop.

**Off by default.** The *threshold* defaults to 15m; the *feature* defaults to
off, so upgrading changes nothing until an operator opts in.

## What it does, and what it deliberately refuses to do

- **Sent, but never learned.** An auto-accept delivers the suggestion and writes
  **no `CorrectionRecord`** — no confidence contribution, no graduation
  progress. A machine's decision to stop waiting is not evidence the suggestion
  was right; otherwise the feature would slowly promote its own guesses.
- **Three guards, and the taxonomy between them is the whole design.** A kill
  switch, a paused/disabled agent, or a busy agent *suppress* and never retire.
  A guard that could not be **evaluated** — unreadable pane, unreachable herdr —
  is also transient: absence of evidence is never evidence of staleness, so an
  outage of any length dismisses nothing. Only a guard that ran and returned a
  negative retires an escalation.
- **Excluded in code, not configurable:** `never_auto_match` and
  `suspected_irreversible` (FR-015), plus `retry_exhausted` (FR-014) and
  `rate_limited` (FR-019). Each means automation has already acted as often as
  it may, and a timeout is not a human checking in.
- **Claimed before delivery**, so a concurrent operator confirm cannot
  double-send; the claim is a distinct transient status, so a crash mid-delivery
  is recoverable rather than a silently lost escalation.
- **At most one per agent per tick.**

## How it is built

Nine focused commits, each leaving the tree shippable:

| | |
|---|---|
| `internal/deliver` | the reply-delivery pipeline extracted from `App.Resolve`, so the daemon and the operator share **one** fail-closed implementation instead of two that drift |
| `sig_*` columns | each decision's full `SignatureResult` persisted as the baseline a much later staleness check needs |
| `[escalations.auto_accept]` | five per-type thresholds behind one switch |
| domain + store | two statuses, a pure eligibility predicate, six status-guarded lifecycle methods |
| `internal/daemon/autoaccept.go` | the pass, on the existing 1-minute sweep — no new ticker, goroutine or `ListAgents` call |
| TUI / CLI | `auto-sent`, `dism:stale` / `dism:gone` / `dism:failed` — never reading as `resolved` |

### Why the baseline had to be persisted

The audit row keeps only the **learning key**, which semantic resolution may have
remapped. Rebuilding a baseline from `pane_excerpt` would produce an
*unstructured* salient where a fresh compute produces a *structured* one, and
`SignatureHeldStill` refuses the fuzzy path whenever either side is structured —
so the comparison could never match and **every approval and choice would
silently degrade into a stale dismissal**. The feature would have looked enabled
while never firing. `escalate()` already receives the `SignatureResult`, so this
is plumbing: nothing is recomputed.

## Two defects found while building

- **Guard 3 re-classified using the agent's live status.** The classifier treats
  status as a *mode selector* — the same approval menu is `approval` at
  `blocked` but `idle` at `idle`/`done`. Any agent that had merely flipped
  `done → idle` (exactly the transition Guard 2 is *required* to absorb) would
  have failed the type assertion on every tick and been dismissed as stale with
  its question plainly on screen. Fixed by classifying under the mode the
  escalated situation lives in.
- Approval signatures **do not include the command**: `Bash(go test …)` and
  `Bash(npm install …)` hash identically and are one learned rule everywhere.
  Inherited coarseness rather than new exposure, but it bounds what Guard 3 can
  distinguish, and it is now recorded next to the test that would otherwise
  imply otherwise.

## Four fail-open paths closed after review

All fixed, each pinned by a test:

1. The **`@noop` sentinel** would have been typed at an agent as literal text.
2. `retry_exhausted` was auto-acceptable — which re-sends the retry the ceiling
   exists to stop, and since the pass writes no correction and is not
   rate-counted, the counter never advances: an unattended loop that repeats
   every threshold window forever.
3. A staleness comparison known to be unreliable for pane-tail salients (the
   baseline is a `--source recent` read, the re-read is `--source visible`) was
   **dismissing** on it rather than leaving it pending.
4. The oldest-first capped candidate query could starve itself to 100%
   ineligible forever on a large pre-upgrade backlog.

## Accepted trade-offs (deliberate — recorded so they are not read as oversights)

- **No never-auto re-scan at fire time** (spec AR-030). Note this differs from
  the deferred-send path at `daemon.go:3337`, which *does* re-screen. Live
  consequence: an operator who adds a never-auto pattern while an escalation is
  pending does not thereby stop it being auto-sent. Reversing this is ~6 lines
  against the situation Guard 3 already produces — say the word.
- **No global per-tick ceiling** — only one-per-agent-per-tick.
- Escalations raised **before** this lands can never auto-accept (no baseline,
  not backfilled). Logged once per run so it is not mistaken for a broken
  feature.

## Rollout

Behavioural change in safety posture: **the plugin can now act with no human
present.** Worth calling out prominently in release notes. `enabled` defaults to
`false`, so no existing install changes behaviour on upgrade.

## Testing

- Full suite green (28/28 packages), `go vet` clean, `golangci-lint` **0 issues**.
- The `internal/frontend` suite passes **unmodified** — the sole acceptance
  evidence that the delivery extraction changed nothing operator-facing.
- ~50 new tests: the transient/terminal guard taxonomy, the claim race (a
  concurrent operator confirm yields exactly one delivery), crash recovery and
  reclaim-after-a-landed-delivery, the structured-salient carve-out, a
  `pane_salient_chars` change mid-wait, retry exhaustion and counter eviction,
  and an explicit assertion that **no `corrections` row is written**.
- ⚠️ `test/integration/` (real herdr + real Claude) **not run** — worth doing
  before merge, since only that catches real CLI-shape drift.

Note: the `internal/daemon` package has **pre-existing** intermittent failures,
confirmed against an untouched `main` checkout which flaked on three *different*
unrelated tests across six runs.


0xGosu:
@CharlieHelps can you review this pr

0xGosu:
Good catch — valid, and fixed in 52c0e2b.

You're right that this was reachable and that the consequence was worse than a
cosmetic status. Confirming the analysis: `auto_accepting` is filtered out of
**both** `PendingEscalations` and `AutoAcceptableEscalations`, and
`ReclaimAbandonedAutoAccepts` only runs in the startup block — so a single
transient `MarkAutoAccepted` failure would hide a **delivered** escalation for
the entire remaining life of the daemon. That is precisely the
silently-lost-escalation failure the transient status was introduced to make
recoverable, and it contradicted that status's own doc comment ("it exists only
for the duration of one delivery attempt").

**The fix:** the audit id is remembered and the finalize is re-attempted at the
top of every sweep until it sticks.

Three properties worth calling out, since they're what make the retry safe:

- **It can never re-deliver.** The send happened once, on the tick that recorded
  the id; the retry is only the `UPDATE`.
- **It can never clobber another writer.** `MarkAutoAccepted` is guarded on
  `WHERE status = 'auto_accepting'`, so a retry can only settle the daemon's own
  claim. If the row moved on meanwhile — a restart's reclaim returned it to the
  queue and an operator then resolved it — it reports not-claimed and the entry
  is dropped.
- **It runs before every early return in the pass**, including the
  feature-disabled and kill-switch ones. Settling a row whose reply already
  landed is bookkeeping about a completed send, not a new action — gating it
  would let an operator strand the row permanently by disabling the feature or
  pausing the herd right after a delivery. (I'd initially have put it after the
  kill-switch check; writing the test for it is what surfaced that.)

Entries clear the moment the store recovers, so the map is empty in a healthy
daemon.

**Tests** (`internal/daemon/autoaccept_test.go`), each verified by mutation to
fail without the fix:

- `TestAutoAcceptFinalizeFailureIsRetriedInProcess` — delivery lands, finalize
  fails, row is stuck at `auto_accepting`; the next tick settles it to
  `auto_accepted` with **exactly one** send and no `corrections` row.
- `TestAutoAcceptFinalizeRetryRunsWhileDisabledAndPaused` — both gates.
- `TestAutoAcceptFinalizeRetryYieldsToAnotherWriter` — the operator's terminal
  state stands and the entry is dropped.

Full suite green (28/28), `go vet` clean, `golangci-lint` 0 issues.

PR #268, opened 2026-07-28
#1 feat: disable the builtin rule that forced an escalation, from the TUI

0xGosu:
## Why

When a shipped never-auto rule is too aggressive for a repo, clearing the escalation it caused meant leaving the TUI: read the regex out of the rationale, find it in `hap rules list` by eye, then `hap rules disable-seed <id>`. The rationale already names the pattern — resolve it in place.

## What

On the **Escalations** tab, and inside an escalation's `v` detail, **`b` disables the one builtin rule that forced the selected escalation**, after a confirmation naming the rule, its id, the escalation it blocked, and what stops being held for a human:

```
disable builtin safety rule 3f2a91c4? situations matching it stop being held for
a human (it forced escalation #7) — pattern: (?i)\bforce[- ]push\b [y/N]
```

The detail view gains a `Builtin rule` line showing the id — on the Audit tab too, where "which rule decided this?" is the likeliest question.

Backend plumbing already existed (`domain.SeedRuleForRationale`, `domain.SeedRuleID`, `frontend.App.DisableSeedRule`); this is the TUI surface for it.

## The gate: named ≠ caused

`domain.SeedRuleForcedEscalation` is new here, and **the CLI's existing `disable-seed` hint now uses it too — it had the same gap**.

A rule being named in a rationale does not mean it caused the escalation. The variance guard deliberately appends the suspected-irreversible diagnostic to its *own* rationale (`domain.Decide`), so a `[variance_guard]` row names a seed rule it was not blocked by. Offering to disable that one would weaken the safety net on a false premise **and leave the operator still blocked**, because the guard keeps escalating the same situation.

So the affordance now requires both:
- a `source=seed` diagnostic (already excluded an operator rule whose pattern text equals a shipped one), **and**
- an escalation reason whose rationale *is* that hit — `never_auto_match` or `suspected_irreversible`.

Anything else fails closed, including an untagged legacy rationale, and the detail line marks such a rule `(noted, not what forced this)`.

This was caught in review; without it, `b` was offered on `[variance_guard]` rows with a confirmation that claimed "it forced escalation #N".

## Bounds

- only the matched rule is disabled — the wholesale `safety.disable_never_auto_seed_patterns` switch is never touched;
- the overlay acts on the rule snapshotted at open, never the live cursor (a refresh lands every 2s), like the other per-entry actions;
- the confirmation revalidates before writing, so a rule disabled meanwhile reports that instead of a no-op "disabled";
- `b` is inert on the read-only Audit tab, and is advertised only when the selected row actually carries a builtin match.

## Tests

16 new tests in `internal/tui/seedrule_test.go`, plus `TestSeedRuleForcedEscalationRequiresTheReasonToBeTheHit` (domain) and `TestEscalationsOmitsDisableSeedWhenAnotherReasonNamedTheRule` (CLI). Fixtures use the **real persisted rationale shape** (`[reason] diagnostic`) and a **non-index-0** seed rule, so an implementation that resolved "the first seed rule" instead of the matched one cannot pass; the write test also asserts the first rule is *not* disabled.

Two existing CLI tests were built on a bare-diagnostic rationale the daemon never writes; their helper (`seedHeuristicDiagnostic` → `seedHeuristicRationale`) now produces the tagged form.

Full `go test -tags "vectors cpu" ./...` green (28 packages); `gofmt`, `go vet`, `golangci-lint` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,54 @@
+# Changelog
+
+Every change gets a line here, including patch releases — see the Changelog
+section in `CLAUDE.md`. Entries sit under **Unreleased** until the release that
+carries them is tagged, then that heading becomes the version number.
+
+## Unreleased
+
+- Added `b` on the Escalations tab (and in an escalation's `v` detail): disables the one builtin never-auto rule that forced the selected escalation, after a `[y/N]` confirmation naming the rule, its id, and the escalation it blocked — previously this meant leaving the TUI to match a regex by eye in `hap rules list` and run `hap rules disable-seed <id>`
+- Added a `Builtin rule` line to the escalation and audit detail views, showing the rule's stable id so it can be acted on from either surface
+- `b` disables only the rule that forced that escalation; it never sets the wholesale `safety.disable_never_auto_seed_patterns` switch, never resolves to a builtin when your own `never_auto_patterns` entry has the same text as a shipped rule, and does nothing on the read-only Audit tab. Undo with `hap rules enable-seed <id>`
+- Fixed a rule merely *named* in a rationale counting as its cause: a `variance_guard` escalation quotes the suspected-irreversible diagnostic without having been blocked by it, so `b` is not offered there and the detail line marks the rule `noted, not what forced this`. The same gate now guards the `hap escalations` disable-seed hint, which had the identical gap
+
+## 0.5.6
+
+- Fixed a submodule gitlink replaced by a symlink taking down every CI job with a linker error deep inside `scripts/setup-native.sh` that named neither git nor submodules; `scripts/check-submodule-gitlink.sh` now runs first in every job that builds native dependencies, and as `make check-submodules`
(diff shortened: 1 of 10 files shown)


0xGosu:
@CharlieHelps review this PR

0xGosu:
Added a second commit (`a7bcfe9`, docs only): **`CHANGELOG.md`**, plus the rule that keeps it alive.

- The `b` shortcut is the first entry, under `## Unreleased`.
- Backfilled every release since 0.5.0 — 0.5.6 (submodule gitlink guard), 0.5.5 (auto-accept of aged escalations), 0.5.4 (task-source help fixes), 0.5.3 (version header + update check), 0.5.2 (the breaking pre-delivery task review) and 0.5.1 (two Claude API error banners). 0.5.0 itself was a 45-commit release and stays on the Releases page.
- `CLAUDE.md` gains a **mandatory** Changelog section: every change, patch releases included, entry under `## Unreleased` in the same commit, verb-first one-liners saying what it *means* rather than what the diff did. `CONTRIBUTING.md`'s PR checklist points at it.

Note for merge ordering: #270 also needs a changelog line, but `CHANGELOG.md` only exists on this branch. Whichever of the two merges first brings the file; I'll merge main into the other and add its entry rather than have both create the file and conflict.

PR #270, opened 2026-07-28
#1 fix: bind seed-rule attribution to the rationale's first diagnostic

0xGosu:
Found by the audit asked for after #268: sweep every consumer that infers something from an escalation's free-text rationale, looking for more of the "named ≠ caused" defect class. One real bug, in the function #268 itself is built on.

## The bug

`domain.SeedRuleForRationale` decides whether an escalation was caused by a **shipped** never-auto rule: it finds `pattern <P> matched ` and then looks for the `(source=seed ` marker. It scanned *everything* after the pattern marker — and the hit's own excerpt sits in that region.

**The excerpt is pane content — agent output.** Two ways that resolved the wrong rule:

1. an operator rule whose pattern text equals a seed's renders `(source=operator `, but an excerpt containing `(source=seed ` satisfied the scan anyway — defeating the exact guard `TestSeedRuleForRationaleRejectsOperatorDuplicateOfSeed` exists for;
2. an excerpt carrying a whole fabricated diagnostic matched the pattern marker outright, needing no operator rule at all.

Either way the operator is pointed at a builtin rule that is not what blocked them: they disable a shipped safety rule (weaker net) and stay blocked, because the rule that actually fired is untouched. Verified live before fixing — both forms resolved, and form 2 also passed straight through #268's new causation gate.

## The fix

The doc comment already claimed the check was anchored *"immediately after that hit's excerpt"*. It now is. `%q` escapes the excerpt's own quotes and backslashes, so parsing the quoted string always lands on its real end, and the source marker read is the hit's own. Every occurrence of the pattern marker is tried, so a fabricated diagnostic inside an excerpt cannot mask a genuine hit later in the rationale.

No stored data or rendering changes — this is read-side only.

## Audit result for the rest

| Consumer | How it reads the rationale | Verdict |
|---|---|---|
| `SeedRuleForRationale` | scan-anywhere for the source marker | **broken — fixed here** |
| `IsRetryableLLMEscalation` | `HasPrefix("[reason]")` | anchored, correct |
| `AutoAcceptIneligible` | `EscalationReason` (anchored parse) | correct |
| `AutoDismissReason` | `Contains("[tag]")` anywhere | weak, see below |
| `SuggestTaskPrefix` / audit action prefixes | `HasPrefix` on daemon-written fields | correct |
| `MatchSummary` | reads a column, not text | correct |

`AutoDismissReason` matches its tag anywhere, and LLM text does reach rationales — so an LLM that wrote `[auto_dismiss_stale]` would mislabel an operator's own dismissal. **Not changed**: nothing acts on it (it only picks a status-column label, with the full rationale one keypress away), and the sound fix is a column rather than smarter string matching.

## Known residual, deliberately not fixed here

Only the excerpt is `%q`-quoted. Other untrusted text is interpolated raw — the daemon appends `"LLM: "+rationale` on the retry path and stores a consult error verbatim, and a `never_auto_match` escalation **can** carry both (`daemon.go:3524`), so #268's reason gate does not cover it either. A fabricated diagnostic placed there has real quotes and parses cleanly.

It is much harder to reach (the model must emit a shipped pattern's regex verbatim) and I did not want to widen a read-side fix into changing the daemon's stored rationale format across six interpolation sites. It is written into the doc comment, with the two ways to close it: quote untrusted text before interpolating it, or require the hit to be the first hit-shaped diagnostic in the string (sound only together with #268's reason gate). Happy to do either as a follow-up — say which.

## Tests

`internal/domain/seedhit_anchor_test.go`: both spoof forms, the `endOfQuoted` escape table (including the trailing-lone-backslash path and an empty excerpt), a no-source-marker rationale staying unattributable, and a round trip asserting **every one of the ~90 shipped rules** still resolves from its own diagnostic — that last one is what keeps the stricter parsing from silently breaking the feature for some pattern.

Full `go test -tags "vectors cpu" ./...` green (28 packages, `internal/daemon` included with no flake); `gofmt`, `go vet`, `golangci-lint` clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ carries them is tagged, then that heading becomes the version number.
 
 ## Unreleased
 
+- Fixed the builtin-rule attribution behind `b` and the `hap escalations` hint being spoofable by agent-influenced text: it searched the whole rationale for a shipped rule's `(source=seed …)` marker, so a fabricated diagnostic in a pane excerpt or in appended LLM/error text could claim a rule that never fired — and because the search ran in seed-list order, a forgery naming an earlier rule could even outrank the genuine hit. You would be offered "disable this builtin rule" for an unrelated safety control while the rule that actually blocked you kept blocking. Attribution is now bound to the first diagnostic in the rationale, which is always the genuine one
+
+## 0.5.7
+
 - Added `b` on the Escalations tab (and in an escalation's `v` detail): disables the one builtin never-auto rule that forced the selected escalation, after a `[y/N]` confirmation naming the rule, its id, and the escalation it blocked — previously this meant leaving the TUI to match a regex by eye in `hap rules list` and run `hap rules disable-seed <id>`
 - Added a `Builtin rule` line to the escalation and audit detail views, showing the rule's stable id so it can be acted on from either surface
 - `b` disables only the rule that forced that escalation; it never sets the wholesale `safety.disable_never_auto_seed_patterns` switch, never resolves to a builtin when your own `never_auto_patterns` entry has the same text as a shipped rule, and does nothing on the read-only Audit tab. Undo with `hap rules enable-seed <id>`
--- a/internal/domain/safety.go
+++ b/internal/domain/safety.go
@@ -190,28 +190,126 @@ func IsSeedPattern(pattern string) bool {
 }
 
 // SeedRuleForRationale maps an escalation rationale back to the seed rule that
-// produced it. It requires the stable Diagnostic() rendering
-// ("pattern <P> matched …") AND, immediately after that hit's excerpt, the
-// seed-source marker "(source=seed ": an operator rule whose pattern text
-// happens to equal a seed's — or a hit taken once that seed is disabled —
-// renders "(source=operator " and must NOT resolve to a builtin, or the hint
-// would name an unrelated (or already-inactive) rule. Anchoring the source
-// check to the position after this rule's own match keeps a crafted excerpt
-// elsewhere in the rationale from spoofing it. The bool is false when no seed
-// rule matches.
+// produced it, by parsing the FIRST hit-shaped diagnostic in the string and
+// requiring that hit's own provenance to say seed. The bool is false when the
+// first hit is operator-sourced, names no shipped pattern, or is absent.
+//
+// Binding to the FIRST hit is what makes attribution safe, and it is why this
+// parses a diagnostic rather than searching for one. Two regions of a rationale
+// are attacker-influenced: the excerpt is PANE CONTENT, and the daemon
+// interpolates raw LLM text after the hit (the retry path appends
+// "LLM: "+rationale, and a consult error is stored verbatim) — a
+// never_auto_match row can carry both. Searching would let either region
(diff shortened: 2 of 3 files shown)


0xGosu:
@CharlieHelps please review this PR

0xGosu:
Both blocking findings addressed at `b6027a2`.

## 1. Raw-rationale residual — closed, not documented

You were right, and the seed-list-order point is sharper than my own analysis: the failure isn't only that a forgery *resolves*, it's that a forgery naming an **earlier** seed rule **outranks the genuine hit**. And documenting a bypass in a function that gates disabling a shipped guard isn't a mitigation.

`SeedRuleForRationale` no longer searches. It parses the **first hit-shaped diagnostic** in the rationale and requires *that* hit to be seed-sourced and to name a shipped pattern:

- `firstDiagnostic` walks `pattern ` occurrences and returns the first that parses completely; occurrences that fail (a forgery inside a `%q`-escaped excerpt) are skipped rather than ending the search, so they can't hide the genuine hit behind them;
- `parseDiagnostic` splits `<pattern> matched <%q excerpt> (source=… kind=…)`, trying each ` matched ` split left-to-right — the same split `Sprintf` produced — since a pattern may itself contain that infix;
- `sourceOf` reads the hit's own provenance; a hit rendered with neither `source` nor `kind` carries no suffix and stays unattributable.

The invariant this rests on, stated explicitly in the doc comment: **the first hit is the genuine one on every path that produces these rationales.** The daemon writes `[reason] ` plus at most a literal prefix (`action review: `, `LLM action: `) ahead of `Diagnostic()`, and everything untrusted is *appended*. I checked the producers rather than assuming — this is why "immediately after the reason prefix" would have been too strict and produced false negatives on the two prefixed paths.

Remaining assumption, now stated rather than buried: the `Pattern` field is interpolated unescaped, so an operator who authors a rule whose pattern text is itself a diagnostic can name any rule. That's config authoring, not agent-reachable.

## 2. Rebase + changelog — done

Rebased onto `a5a473c`; the conflict in `safety.go` was a both-added at the same anchor (your `SeedRuleForcedEscalation` from #268 vs. these helpers) and both sides are kept. Changelog entry added under `## Unreleased`. GitHub now reports the PR `MERGEABLE` / `CLEAN`.

## Regressions

`TestSeedRuleForcedEscalationIgnoresAppendedForgery` is the one you asked for — a genuine **operator** hit with a forged **seed** diagnostic appended, across all three real shapes (`; LLM: …`, `; llm confidence …; LLM: …`, and a bare consult error), asserted through both `SeedRuleForcedEscalation` and `SeedRuleForRationale`. It also pins the seed-list-order hazard directly: a forgery for `rules[0]` appended after a genuine hit for `rules[len-1]` must still resolve to the genuine later rule.

Alongside: both excerpt spoof forms, the `endOfQuoted` escape table, a no-source-suffix rationale staying unattributable, and a round trip asserting every one of the ~90 shipped rules still resolves from its own diagnostic — that last one is what stops the stricter parsing silently breaking the feature for some pattern.

Full suite green (2,021 tests across domain/cli/tui/frontend/daemon); `gofmt`, `go vet`, `golangci-lint` clean.

PR #273, opened 2026-07-28
feat: notify the operator through herdr when the TUI needs attention

0xGosu:
## Why

The TUI alerted only with a terminal bell. A bell does nothing in many terminals, and nothing at all when you are looking at a different herdr tab or another app — so an escalation could sit unnoticed.

herdr 0.7 exposes `notification.show` over its control socket and injects `HERDR_ENV=1` + `HERDR_SOCKET_PATH` into every managed pane — and the TUI *is* a managed pane. So it can detect that it is running inside herdr and raise a real notification instead.

## What changed

When herdr launched hap as a managed pane, the TUI raises a herdr desktop notification for the same two events the bell always covered:

1. a new escalation appearing since the last 2s poll
2. the pause/kill switch becoming active because of a **different** process

**The bell becomes the fallback rather than the only channel.** Unlike the CLI's fire-and-forget `herdr notification show` (which exits 0 whether or not the toast appeared), the socket method answers `{shown, reason}` — so a toast herdr declined to paint (`disabled`, `rate_limited`, `no_foreground_client`, `busy`) rings the bell instead of being silently counted as delivered.

Outside herdr nothing changes: no socket is opened and the bell behaves exactly as before.

### Config

New `tui.herdr_notification` (default **on**), independent of the existing `tui.terminal_bell` (default on):

| `herdr_notification` | `terminal_bell` | behavior |
|---|---|---|
| true | true | toast; bell only if the toast was not shown |
| true | false | toast only |
| false | true | bell (today's behavior) |
| false | false | silent |

## Notes for the reviewer

- **The daemon already toasts on escalation** via the CLI (`daemon.escalate`), and it is deliberately left alone — so both can fire for the same escalation. herdr's in-app `busy` de-dup usually swallows the second. This overlap was an explicit decision, not an oversight.
- **Every call opens its own connection.** Verified live against herdr 0.7.3: the server *closes the connection right after answering* `notification.show`, on the success path as well as the error path, so a reused connection dies with EPIPE. This contradicts the socket-API docs' "multiple methods work on one connection", and there is an integration test pinning it so nobody optimizes it into a pooled connection.
- `ports.NotifyShower` is a new **optional** interface alongside the unchanged `NotifyPort`, type-asserted at the call site — the repo's standard optional-capability pattern.
- The notification is dispatched **off the bubbletea update loop**, on a goroutine joined to the `inflight` WaitGroup `Run` already drains, so a socket round trip can never freeze the UI.
- The one-shot request/response helper is **extracted out of `listPanes`** and shared. The rewritten `listPanes` is behavior-preserving (same scanner buffers, same conn-close ordering, same error precedence).

### Two fixes beyond the original scope

Both came out of code review and both have tests:

- `pausePending` was consumed only *inside* the alerting branch. With alerts off it latched, and the next externally-caused pause would read as self-caused and be **silently swallowed**. It is now consumed by the pause transition itself.
- A failed toast now logs at `Warn`, not `Debug` — the TUI logs at Info, so a permanently broken socket previously left the operator with an unexplained bell and nothing in the log.

## Testing

- Full suite `go test -tags "vectors cpu" ./... -count=1` — green
- `golangci-lint run --build-tags "vectors,cpu"` — 0 issues; `go vet` + `gofmt` clean
- `-race` on `internal/tui` and `internal/herdr` — clean
- New unit tests: the socket notifier against the fake herdr socket (result shape, not-shown-is-not-an-error, protocol error, unreachable socket, text clamping, unique request ids, concurrency), and the full four-corner config matrix in the TUI
- New **live** integration tests (`test/integration/notification_test.go`, build tag `integration`) driving a real herdr — they raise real toasts and assert the result shape, that two consecutive calls both land, and that an empty title is refused both locally and by herdr

> One earlier suite run reported 4 daemon failures; those were CPU contention from a concurrent race run, not regressions. Confirmed by re-running this branch and `main` serially — both green.

### DIFF
--- a/.claude/skills/hap/SKILL.md
+++ b/.claude/skills/hap/SKILL.md
@@ -415,7 +415,8 @@ per-command env notes: layering is daemon env → `env_file` → `env` → the c
 | `tui.max_content_height` | 0 (full height) | cap the rows a list body may use; 0 = full height |
 | `tui.theme` | default | TUI color theme: default, dark, light, high-contrast (in the TUI Config tab, `e` on this row opens a ↑/↓ picker of the available themes) |
 | `cli.ai_agent_friendly_output` | true | append the "Next steps" footer to command output (for AI agents driving the CLI); never affects `hap help` / `--help`, which always show theirs |
-| `tui.terminal_bell` | true | ring the terminal bell (\a) on new escalations and on pauses caused by a different process |
+| `tui.terminal_bell` | true | ring the terminal bell (\a) on new escalations and on pauses caused by a different process; also the fallback when a herdr notification is not displayed |
+| `tui.herdr_notification` | true | raise a herdr desktop notification on those same two events, when the TUI runs as a herdr-managed pane |
 | `tui.disable_check_for_update` | false | turn off the GitHub release check (at most every 6h, TUI only) that puts `↑ vX.Y.Z available` in the header. it is the plugin's ONLY outbound network call; a locally built (`plugin link`) binary never checks |
 
 TUI palette colors (`tui.palette.*`) are config.toml-only — roles: `title`, `section`, `error`, `ok`, `paused`, `running`, `warn`, `help`. values are 256-color codes (`"205"`) or hex (`"#ff5faf"`).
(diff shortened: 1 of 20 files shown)


0xGosu:
@CharlieHelps review this pr

PR #277, opened 2026-07-29
#1 fix: stop a pending escalation from silently withholding auto-send tasks

0xGosu:
## The bug

`enable_auto_send_task_when_idle` went **permanently quiet** on any agent that had an escalation waiting. `eligibleIdleAgents` skipped it with no log and no audit row.

That deadlocked the feature against itself:

1. A declared source with a pending `[ ]` item is exactly what raises `noop_vs_pending_tasks` — a learned `@noop` rule matching an idle screen while real work is queued.
2. That escalation then blocked the very poll that would have delivered the item.
3. Item stays pending → next idle event → escalates again.

Found live: one agent with 8 open escalations sitting idle beside its own task list, its next task never handed out. Six of the eight were that exact loop.

## The fix

**A pending escalation no longer withholds queued work.** It is a question for the operator about what to answer on the agent's *screen*, not a verdict on whether the agent can take its next declared task.

**The bound on an undeliverable task moves to the per-ITEM attempt counter.** This is what made a plain deletion of the gate unsafe: a failed send rolls its item back to `[ ]` and records **no** reservation (the ledger row is written only after herdr accepts the keystrokes), so `reclaimStrandedTasks` can never age it toward `maxTaskHandouts`. The gate had been standing in for that bound.

`deliverAutonomousClaimed` now counts its own failed attempt (`RecordTaskHandoutAttempt`, returning the post-bump count) and at the ceiling skips the rollback — leaving the item `[-]` and escalating, exactly as `escalateNeverStartedTask` does for a hand-out that landed but was never taken up.

| | before | after |
|---|---|---|
| escalation blocks hand-out | any pending escalation, any agent | never |
| bound on undeliverable task | none (the gate stood in for it) | per-item attempt counter |
| failed send below ceiling | item → `[ ]`, agent benched forever | item → `[ ]`, agent keeps working |
| failed send at 3rd attempt | — | item left `[-]`, escalated, agent moves to next item |

## Two traps recorded in the code

**Do not reintroduce a trigger-based "is this the poll's own escalation" test.** An earlier version of this PR did exactly that and it was wrong: `daemon.trigger` derives the audit trigger from `tr.AutoIdleSend`, so *every* escalation raised on a poll-driven episode is stamped `auto-idle-send:` — and for an agent that parks without a further herdr event (the case this feature exists for) the poll is the only thing raising episodes, making that the common shape. It would have moved the latch, not lifted it. The `poll-raised` cases in `TestAutoSendIdleOrdinaryEscalationDoesNotBlockHandout` pin the trap.

**A standing approval does not classify as `approval` at idle.** A comment justifying the removal asserted it did. The classifier treats herdr's status as a *mode selector*: approval/choice content rules only match while the agent reads `blocked`, and an unmatched idle/done pane falls through to `SituationIdle`. The comment now names the three real protections (herdr normally reporting `blocked`; the two structural detectors that survive at idle/done; a menu-bearing pane rarely matching a graduated idle rule) and the residual risk, instead of a safety property that does not hold.

## Also in here

- **Re-drive backoff.** The poll deliberately ignores `episodeHandled`, so it re-drives a parked pane every sweep. With the gate gone, an agent whose episodes keep not sending was re-read once a minute forever, one `ignored` audit row at a time, in a table nothing prunes. `Daemon.pollRedrive` widens that interval (1, 2, 4 … capped at 15m); any delivered send clears it. A delay, never a bench.
- **Silent-loss hole closed.** `escalateUndeliverableTask` writes its audit row *before* clearing the counter and reports whether the operator was actually told; a failed write falls back to the ordinary rollback, so a task can never end up `[-]` with no escalation, no reservation and a cleared counter.
- Stale comments in `config.go`, `daemon.go` and `tasklistreview.go` that still described the old gate as intended; `TriggerAutoSendReclaim` applied at all three sites instead of two literals.

## Tests

New: ordinary escalations (including poll-raised) still hand out; the ceiling fires at exactly `maxTaskHandouts` and **not earlier**; a capped task lets its agent move to the next item; a non-reserving source counts nothing; the backoff widens, caps, and is cleared by a delivery; the LLM-review delivery path counts its failed send — previously untested, `setFailSend` had never appeared in `tasklistreview_test.go`; plus a store-level test for the shared counter.

Four behaviours were **mutation-checked** rather than assumed, because the pre-fix suite was green *on the bug*:

| mutation | result |
|---|---|
| restore the old escalation gate | all 9 subtests of `…OrdinaryEscalationDoesNotBlockHandout` fail |
| disable the ceiling (`case false && attempts >= max`) | `…UndeliverableTaskIsCappedNotRetriedForever` fails |
| scope the count away from the review path (`&& callerRollback == nil`) | `TestTaskReviewFailedSendCountsTheHandoutAndRollsBack` fails |

Full suite, `golangci-lint`, `gofmt` and `go vet` clean. The nine race-sensitive tests verified at 5/5 in isolation.

## Notes for review

- `internal/daemon` has documented rotating flakiness (~1 arbitrary test per package run, a different one each time). Two runs during verification: one clean, one failing only `TestShutdownDrainsVerifyUnblockBeforeStoreClose`, unrelated to this change.
- One test gap left deliberately: mixed evidence across both ceilings (landed-but-unstarted hand-outs plus refused deliveries summing to one ceiling). The store test proves the counter is shared; a daemon-level test would need reservation backdating interleaved with send failures.
- Changelog written as `## 0.5.11` on top of `main` at 0.5.10.


0xGosu:
@CharlieHelps please review

PR #279, opened 2026-07-29
#1 fix: deliver enable_auto_send_task_when_idle tasks without waiting to graduate

0xGosu:
Follow-up to #277. That PR fixed the escalation gate that was blocking the poll; this one fixes why the poll still never delivered.

## The bug

Turning `enable_auto_send_task_when_idle` on still produced no unattended send.

Every idle episode goes through the two **learning** gates in `domain.Decide` — shadow mode (`:293`, FR-004/FR-006) and the idle confidence threshold (`:305`, FR-008). So a declared task only went out once its *situation signature* had graduated, which takes two operator confirmations. And because every idle screen mints its own signature, each one started at 0/2, escalated `shadow_mode` with the task as a suggestion, and waited.

The feature could not deliver anything without exactly the human attention it exists to remove.

A second blocker sat behind it: a learned `@noop` plurality took precedence over the declared task, either escalating `noop_vs_pending_tasks` or standing down outright, parking the queue.

Observed live on `deft-heron`: source opted in, one pending task, agent parked — and a `shadow_mode` escalation carrying that exact task as its suggestion, every time.

## Why the gates are the wrong question here

Both ask *"has this **signature** earned the right to act on its own?"*

That is the right question for an inferred reply to a screen. It is the wrong one for a task the operator typed into a checklist and flagged for unattended delivery: the content comes from their file, not from history, so there is nothing to have learned. `enable_auto_send_task_when_idle` is an instruction about a **queue**; a learned action is an inference about a **screen**. When they disagree, the instruction wins.

## The change

Two places, one behavioural edit each:

- **`resolveSituation`** — a source with `DeclaredTask.Reserve` (which *is* the flag) and a real pending item resolves to that item ahead of every noop precedence.
- **`Decide`** — that resolved hand-out bypasses the mode gate and the confidence gate.

The bypass keys off `idleHandout`, the predicate the file **already computed** for the FR-019 rate-guard exemption: the *verified* classified situation plus the *resolved* action, never a sweep-time flag. A pane that became an approval since the sweep, or an episode resolving to `@noop`, cannot reach it.

## It skips graduation, not safety

Placement is deliberate — every safety control runs **before** the bypass or **after** it at delivery:

| control | still applies | where |
|---|---|---|
| variance guard | ✅ | before, in `Decide` |
| rate ceilings (FR-019) | ✅ | before, in `Decide` |
| suspected-irreversible heuristic | ✅ | before, in `Decide` |
| kill switch | ✅ | `Decide` + daemon |
| never-auto patterns | ✅ | daemon, at delivery |
| per-agent disable | ✅ | daemon, at delivery |
| pre-delivery LLM review | ✅ | daemon, at delivery |
| shadow mode / confidence threshold | ❌ **bypassed** | these are learning gates |

`TestDecideUnattendedSourceStillObeysEverySafetyControl` asserts four of those still stop an unattended hand-out.

One detail worth noting: the decision records `Source: SourceRule`, so an unattended send counts as a rule-sourced vote and can never inflate its own signature toward graduation as though a human had approved it.

## Scope

Sources **without** the flag are attended by definition and keep the historical behaviour in full — a shadow signature still suggests rather than acts, and a learned noop still escalates over pending work. Pinned by `TestDecideAttendedSourceStillWaitsToGraduate` and `TestAutoSendIdleAttendedSourceStillEscalatesWithNoLearnedRule`.

`enable_llm_review_before_auto_send` needed no change — verified that it already never gates. Its only stand-down path is the kill switch; every other outcome sends the original task unchanged. It shapes *which* task, not *whether*.

## Tests

The bypass fires for a brand-new signature, in shadow mode, one confirmation short of graduating, and below the confidence threshold; it outranks a learned noop of every provenance in both modes; an exhausted source still sends nothing. Two daemon tests drive the end-to-end shape with **no seeded rule at all** — one asserting delivery with `noEscalations`, its mirror asserting an attended source still escalates.

Both behaviours mutation-checked:

| mutation | result |
|---|---|
| disable the bypass (`if false && idleHandout`) | 5 domain subtests + `TestAutoSendIdleUnattendedSourceSendsWithNoLearnedRule` fail |
| disable the noop override in `resolveSituation` | `TestDecideUnattendedSourceOutranksALearnedNoop` fails |

Full suite, `golangci-lint` and `gofmt` clean. All new tests verified 5/5 in isolation.

## Note for review

`internal/daemon` has documented rotating flakiness — roughly one arbitrary test per full-package run, a different one each time. Four package runs during verification: two clean, two failing a different test each (`TestShutdownDrainsVerifyUnblockBeforeStoreClose`, `TestAutoSendIdleCapsAtExactlyMaxHandouts`), both passing 5/5 in isolation.


0xGosu:
@CharlieHelps please review

PR #283, opened 2026-07-30
#1 feat: raise daemon herdr notifications over the socket API

0xGosu:
## Why

The daemon raised operator toasts by shelling out to `herdr notification show`, which **exits 0 whether or not a toast was actually painted**. So when herdr dropped an escalation — notifications disabled, rate limited, no foreground client, an in-app toast already standing — the daemon assumed the operator had been told. For a tool whose entire job is escalating to a human, that is a real gap.

herdr's socket method `notification.show` answers `{shown, reason}`. The TUI already used it for exactly this reason (IR-003); the daemon now does too.

This is the first step of a deliberately **selective** socket migration. Measured against a live herdr, the socket is *not* meaningfully faster — herdr's own ~120 ms server-side handling dominates both paths and process spawn is ~3 ms — so this moves for **capability**, not speed.

## What changed

- **`ports.NotifyResult` gains `Known`** — separates "herdr answered and said no" from "we never found out". The CLI cannot report delivery, so the zero value is the unknown one: a result nobody filled in never claims delivery.
- **New `herdr.FallbackNotifier`** — prefers the socket, drops to the CLI **only on transport failure**. A request herdr *answered and refused* is never re-fired (same herdr, same verdict, and it would bury the real error code); neither is one refused locally (empty title) or cancelled by shutdown. `method_not_found` still falls back — an older herdr lacking the method is precisely the capability gap a backstop is for.
- **One budget bounds the whole chain**, so adding the socket hop cannot stall the daemon's select loop longer than the CLI alone did (CLAUDE.md: don't stall the main loop).
- **`escalate` logs `notified=` / `notify_reason=`**, emitted only when herdr reported the outcome — the log never turns "we don't know" into a claim. Extracted as the pure `escalationLogAttrs` so the invariant is directly testable.
- IR-003 and the protocol-surfaces table updated to match.

## Testing

- `internal/herdr`: all notification tests pass, including under `-race` (the chain is lock-free by design: fresh connection per call, atomic request ids).
- New coverage: socket-preferred delivery, declined-toast-not-retried, transport-failure-falls-back-unreported, herdr-answered errors (table: `invalid_params` refuses vs `method_not_found` falls back), cancelled ctx, empty title, joined errors, both-nil, and a concurrent case.
- `escalationLogAttrs` table-tested across displayed / declined / unreported.
- Full suite: the `internal/daemon` package has **pre-existing** full-suite flakiness. Verified against a clean `origin/main` worktree — a `./...` run there failed 4 arbitrary daemon tests, including one this branch never hit. Every targeted test here passes.

## Not in this PR

- **`pane.focus`** (would replace the `tab focus` + `pane zoom --on` hack in `CLI.FocusPane`, which mutates the operator's layout) — it is **undocumented on herdr.dev** and confirming it switches the active tab needs a live test that moves the operator's view.
- The polling migrations (the 1-minute `agent list` sweep; `retrySubmitEnter`'s up-to-15 spawns per send) — separate PRs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ section in `CLAUDE.md`. Entries go under the version that will carry them: mergi
 to main auto-releases the next patch, so a PR adds its lines under that number
 (or under the minor/major version it bumps the manifest to).
 
+## 0.5.14
+
+- Changed the daemon to raise its herdr notifications over the socket API instead of the `herdr notification show` CLI, so it now learns whether a toast was actually displayed. The CLI exits 0 even when herdr paints nothing, which meant an escalation could be dropped — notifications turned off, rate limited, no foreground client — with the daemon assuming the operator had been told
+- Added the delivery outcome to the daemon's `escalated` log line: `notified=true`, or `notified=false` with the reason herdr gave. Absent when nothing reported it, so the log never turns "we don't know" into a claim
+- Kept the CLI as the fallback for when the socket itself is unreachable. A request herdr answered and refused is never re-fired through it — same herdr, same verdict — and neither is one already refused locally or cancelled by shutdown
+- Changed the notification timeout to bound the whole socket-then-CLI attempt rather than each hop, so adding the socket cannot make a wedged herdr stall the daemon longer than the CLI alone did
+
 ## 0.5.13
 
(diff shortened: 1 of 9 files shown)


0xGosu:
@CharlieHelps please review

PR #293, opened 2026-07-31
#1 feat: fall back to BM25 when similarity search finds no match

0xGosu:
## The problem

`daemon.resolveSignature` resolves a situation onto a learned signature through a chain: exact hash → cosine → BM25 → mint a new key. The BM25 step was guarded by `if vec == nil`, so it only ran when embedding was **skipped** (below `min_salient_chars`), **unavailable**, or **errored**.

A vector search that ran cleanly and simply found nothing above `similarity_threshold` fell straight through to the mint block — silently, without even a log line.

So a screen that was a near-verbatim render of a learned one minted a brand-new signature whenever its embedding happened to land below 0.90. The operator got a fresh escalation for a situation hap had already been taught, and the new rule had to graduate from scratch.

## The change

Drop the guard. All five paths that reach that point want BM25:

| path | before | after |
|---|---|---|
| salient below `min_salient_chars` | BM25 | BM25 |
| embedder nil / not warmed | BM25 | BM25 |
| `EmbedText` errored (incl. degraded latch) | BM25 | BM25 |
| `MatchVector` errored | BM25 | BM25 |
| **KNN ran, nothing above threshold** | **mint new** | **BM25** |

The retry uses the same `remapAllowed` filter and the same `bm25_min_score` bar as the existing fallback. The vector pass's extra `min_salient_chars` veto on candidates is deliberately **not** carried over: the floor closes *cosine* to a short rule while text matching stays that rule's reachable path. No new config key.

### Also in here

- **Stall guard (2s) on the BM25 search.** It now runs inline on the daemon select loop for every non-exact situation, where before it only ran when the embedder was down. Measured ~9ms over a 200-rule corpus, growing with corpus size. On timeout the existing error branch degrades to the hash key — the fail-safe direction.
- **Stop discarding the computed vector when the vector *search* errors.** That `vec = nil` existed only to trigger the old guard, and it persisted a vectorless row despite a successful embed — leaving the signature invisible to cosine until a later daemon start re-embedded it.
- **Record the model id from the embedder that produced the vector**, not whichever is installed at mint time. `reloadEmbedder` swaps the port before clearing `semanticReady`, so a reload racing this call could persist model A's vector labelled model B. That row is self-consistent, so `reembed.Reconcile` keeps it forever and `CountStaleSignatureEmbeddings` never reports it.

## Threshold characterization

`internal/match/bm25_test.go` pins the normalized-BM25 curve rather than just exercising the path — monotonic decay as a query drifts, unrelated text never reaching the stored rule, and the valley that justifies `0.35`:

| relatedness | short salient (47 ch) | long salient (147 ch) |
|---|---|---|
| identical | 1.0000 | 1.0000 |
| one word differs | 0.5073 | 0.6567 |
| partial overlap | 0.1436 | 0.1231 |
| unrelated | does not reach the rule | does not reach the rule |

`0.35` sits in the middle of a wide valley, which is what makes it safe.

**Length shifts the score, not the verdict.** The score is normalized against what the stored salient earns against itself, so one differing word costs a large share of a seven-term salient and a small share of a twenty-term one. Symmetrically, longer salients score *lower* at partial overlap — they separate better at both ends. Both lengths land on the same side of the default.

### Corpus size is load-bearing in these tests

The tests seed a 25-rule corpus deliberately. A **single-row** bleve index has uniform IDF and an average document length equal to that row's own, so both of BM25's discriminators are switched off and any curve measured there is an artifact — the same one-word variant scores **0.33 at one row and 0.53 at a hundred**. `TestMatchTextCorpusSizeDoesNotFlipTheVerdict` guards that trap explicitly.

## Accepted trade-off

Structured salients now get a second chance at `bm25_min_score` after cosine refused them, and two approvals differing by **one word** merge onto one rule (e.g. `deploy … to the staging cluster` vs `… production cluster`, 0.66). This is deliberate and pinned by `TestResolveSignatureCosineMissMergesNearIdenticalApprovals`.

Why it's acceptable:
- A real embedding model puts that pair far above `similarity_threshold` anyway — the test only forces a miss with an orthogonal canned vector.
- A shared learning key is not a delivered answer. The rule still has to graduate, and the kill switch, never-auto patterns, irreversible heuristic and rate guards all still gate delivery.
- `bm25_min_score` remains the operator's lever, verified by re-running the merged pair under a stricter config.

Bounded on the other side by `TestResolveSignatureCosineMissKeepsUnrelatedApprovalsApart`: unrelated approvals sharing byte-identical option-set boilerplate stay apart at 0, 4 and 24 learned rules — the boilerplate never carries them past the bar, because normalization divides by a self-score the verb terms dominate.

`TestResolveSignatureCosineMissKeepsUnrelatedChoicesApart` covers the least-guarded structured type: `remapAllowed` only gates approvals, so a choice salient (which is entirely its option set) reaches the retry with no type-specific gate at all.

## Testing

- `go build`, `gofmt`, `go vet`, `golangci-lint run --build-tags "vectors,cpu"` — all clean, 0 lint issues.
- `go test -tags "vectors cpu" ./... -count=1` — 25/26 packages pass. The single failure was `TestShutdownDrainsVerifyUnblockBeforeStoreClose`, which passed 5/5 in isolation; it is a shutdown-lifecycle test with no path to signature resolution. Across three full-suite runs, three *different* arbitrary daemon tests flaked and none repeated — matching the documented `internal/daemon` baseline.
- 14 new tests, plus every pre-existing signature-resolution, salient-floor, reembed and `MatchSummary` test green.

## Docs

`MatchBM25`'s enum doc, the `bm25_min_score` config doc, the `internal/match` package doc, the architecture doc, `sample/config.toml`, `README.md` and the `hap` skill all described BM25 as an embedder-unavailable-only fallback. All updated. `MatchSummary` now renders `text fallback` instead of `embedding fallback`.


0xGosu:
@CharlieHelps please review — worth a close look at the accepted trade-off: the BM25 retry now gives structured salients a second chance after cosine refused them, so two approvals differing by one word merge onto one rule (pinned by `TestResolveSignatureCosineMissMergesNearIdenticalApprovals`). Unrelated approvals sharing identical option-set boilerplate stay apart, verified at 0, 4 and 24 learned rules.

0xGosu:
@CharlieHelps addressed in e9b93d2 — you were right, and investigating it turned up a second reason the fix had to be stronger than a threshold.

**The finding stands.** `internal/domain/signature.go:306-313` does rule the other way on this exact `test`→`live` example, and the cosine-miss retry contradicted it. A structured salient (approval verb + options, choice option set, error summary) that the embedding search has already refused is now **not reconsidered by BM25 at all** (`daemon.bm25RetryAllowed`).

**It is a rule, not a stricter bar** — I tried the stricter-bar route you offered as the alternative and it cannot work, for two independent reasons:

1. **BM25 cannot express "preserve target identity."** Over a 25-rule corpus, the target swap (`… to the test service` → `… live service`) and a *harmless rewording of the verb against the same target* (`apply the pending migration …` → `run the pending migration …`) both score **0.657887** — exactly equal. It is a bag of words: it knows how many terms differ, never which. No threshold admits the second and refuses the first. Pinned by `internal/match.TestMatchTextCannotSeparateATargetSwapFromARewording`.

2. **The score is not stable enough to threshold.** `MatchText` normalizes by a *second* search (`textCandidates` → `textSelfScore`, `match.go:416`) that is not snapshot-consistent with the first. On an incrementally built index — which is every production index, since each minted signature calls `matcher.Add` — a background scorch segment merge landing between the two searches inflates the ratio. Over 40 identical trials the same pair scored 0.658 in 80% of them and **0.766 / 0.813 / 0.861 / 0.944 / 1.000** in the rest. A bar cannot refuse a pair that intermittently scores a perfect 1.0.

That second point is **pre-existing and wider than this PR** — it affects `bm25_min_score` too, and is tolerable there only because those margins are wide (near-duplicates 0.45–0.87 against partial overlap 0.10–0.15). It is documented in `bm25RetryAllowed` and the architecture doc, and deserves its own issue.

**Scope.** Only the cosine-miss path is closed. A structured salient reaching BM25 because the embedder was *unavailable* keeps `bm25_min_score` exactly as before — `TestResolveSignatureBM25FallbackOnEmbedFailure` pins that the new rule did not leak onto it.

**Also added** `embedding.bm25_highbar_score` (default 0.70) for **pane-tail** salients at or above `min_salient_chars` once cosine has refused them: there the drift is a repainted screen rather than a changed target, so the retry stays open and this bounds how loose it gets. Salients *below* the floor are never embedded — no cosine verdict to overturn, and text matching is their only matcher — so they keep `bm25_min_score`; raising their bar would strand them on exact hash, which the floor exists to avoid.

**On the test you flagged:** yes — `TestResolveSignatureCosineMissMergesNearIdenticalApprovals` is inverted, now `…KeepsOneWordApprovalVariantsApart`. It first checks *against the matcher directly* that text matching **would** have merged the pair (bm25 0.6579), so a pass proves the rule did the work rather than the pair simply not matching. A companion assertion drops `bm25_highbar_score` to the floor and confirms the refusal does not move — it is threshold-free by construction. A choice-situation case covers the least-guarded structured type, since `remapAllowed` only gates approvals.

Full suite green: 30/30 packages, `golangci-lint` 0 issues, and the affected packages run 5/5 stable (the earlier scoring flakiness is gone now that the characterization tests build their index in one batch).

PR #295, opened 2026-07-31
#1 fix: make every config.toml key reachable from `hap config set`

0xGosu:
## The problem

`embedding.bm25_highbar_score` shipped in v0.5.19 settable in `config.toml` but never registered in `frontend.ConfigFields` — the single registry backing **both** the TUI config screen and `hap config fields` / `hap config set`. Verified live:

```
$ hap config set embedding.bm25_highbar_score 0.8
error: unknown config field "embedding.bm25_highbar_score"
```

So the knob was reachable only by hand-editing the file.

**Why no test caught it.** The existing parity tests compare the registry against a hand-written sample map in *both* directions, so a stale sample fails loudly — but nothing walked the config struct. A key added there and forgotten in the registry was invisible to every test.

## The fix

Per the stated invariant — *the CLI and `config.toml` must always be in sync; the TUI uses `TUIHidden` to hide non-important keys*:

- **`embedding.bm25_highbar_score`** registered, TUI-visible beside its sibling `bm25_min_score`.
- **The eight `tui.palette.*` roles** registered as `TUIHidden`. These had *never* been CLI-settable. Hidden keeps the TUI screen readable — eight color strings would bury the settings an operator reaches for — while `hap config set` reaches them, which is exactly what `TUIHidden` is for.
- **`TestEveryConfigKeyIsRegistered`** makes the invariant structural. It walks `config.Config`'s TOML tags and fails if any scalar is unregistered.

Two deliberate choices in that guard, both load-bearing:

1. It walks **types, not values**, so a nil pointer sub-table is still descended into. A value walk cannot see through a nil pointer, and `config.Default()` leaves some nil.
2. It classifies by **exclusion** (skip only slices/maps/interfaces), not by an allowlist of scalar kinds. An allowlist would silently skip the first `uint`, `float32` or `time.Duration` field anyone adds — the same blind spot that let this bug through.

Mutation-tested both ways: adding a `uint` field, and a scalar behind a nil pointer sub-table, each make it fail with an actionable message.

It also surfaced that `safety.disable_seed` and `llm.auto_act` were being skipped *because they are pointers*, not because they are deprecated — accidental correctness. Both are now explicit exemptions with reasons, so the exemption map is an honest record rather than a partial one.

## Palette values are validated

Storing them as-is was a footgun: lipgloss resolves an unrecognized color to **no color at all** and an out-of-range number to an out-of-spec SGR code, both silently. Because these keys are `TUIHidden`, the config screen cannot show what happened and `hap config fields` would echo the bad value back as though it were in effect — `config set` refusing it is the only feedback loop there is. It now matches how `tui.theme` already rejects an unknown name two cases away in the same switch. `""` clears a role back to the theme.

## Tests

`TestNewConfigFieldsRoundTrip` covers what registry parity cannot: parity only asserts `SetField` **accepts** a sample, so a case returning `nil` without assigning anything passes it. Each new key is set, read back, cleared, and checked against the values that must be rejected.

Full suite: 29/30 packages pass, `golangci-lint` 0 issues. The one failure is `TestResolveSignatureSalientFloorSelectsWhichBM25BarApplies`, which this branch does not touch — it is a flaky test on `main` from #293 and is being handled separately.

## Docs

Three places asserted the opposite of the new behavior and are corrected: `SKILL.md` called the palette config.toml-only *and* listed it among keys not settable via `hap config set`; `README.md` said the same and hard-coded "Ten advanced fields" (this makes it eighteen — replaced with "the advanced fields" so the count stops rotting); the `PaletteOverrides` doc comment said the TUI shows them read-only.


0xGosu:
@CharlieHelps please review — this is a follow-up to #293, not a revision of it (#293 is merged and released as v0.5.19).

Two things worth your attention:

1. **`TestEveryConfigKeyIsRegistered`** is the substance here — it makes "every config.toml key is reachable from `hap config set`" structural rather than conventional. It walks TYPES rather than values (so a nil pointer sub-table is still descended into) and classifies by exclusion rather than an allowlist of scalar kinds (so a future `uint`/`float32`/`time.Duration` field cannot be silently skipped). I mutation-tested both properties; if you can find a config shape that slips past it, that's the finding I most want.

2. **Palette validation** rejects anything lipgloss would resolve to no color. I'd value a check that the accepted set is right — `0-255`, `#rgb`, `#rrggbb`, and `""` to clear — and that I haven't excluded a form lipgloss genuinely supports.

Note the one full-suite failure, `TestResolveSignatureSalientFloorSelectsWhichBM25BarApplies`, is not from this branch: it is a flaky test I introduced in #293 and it is being fixed separately.

0xGosu:
@CharlieHelps addressed in 905a9f5 — the finding was valid, and I verified each premise against BurntSushi/toml v1.6.0 rather than taking it on trust. All three shapes really are accepted by the decoder:

| shape | decoder result |
|---|---|
| untagged exported field | decodes under the Go field name, and case-folds (`untagged = "x"` reaches `Untagged`) |
| anonymous embedded struct | fields promoted to the parent level |
| `TextUnmarshaler` type | decodes as a **scalar** despite `Kind() == Struct` |

So each could be settable in `config.toml` while `hap config set` rejected it — the same drift, through a different door.

`tomlScalarKeys` now models the decoder: effective field set, anonymous embeds recursed with the **parent** prefix, custom-unmarshal types treated as scalars. An untagged exported field is reported as its own distinct failure rather than keyed, because the right fix there is to add a tag, not to register a MixedCaps key unlike every other one — that is the "explicitly reject those shapes" option you offered.

`TestTOMLScalarKeysModelsTheDecoder` adds the regression cases you asked for, and its subtest round-trips the same struct through `toml.Unmarshal` so that if the decoder's rules ever change, the test fails rather than quietly pinning a model of a decoder that no longer exists.

Worth reporting: **that test immediately caught a bug in my own fix.** `reflect` sets `PkgPath` on an anonymous field whose embedded *type* is unexported, so my unexported-skip was dropping exactly the promoted fields the decoder does reach — the `promoted` case failed on first run. Fixed, and all three shapes are now mutation-tested against the real `config.Config`: each fails with an actionable message.

Thanks for the catch — it was the same class as the kind-allowlist gap, and I had not carried the lesson far enough.

PR #296, opened 2026-07-31
#1 fix: stop a BM25 bar test sampling a scoring race

0xGosu:
## The problem

`TestResolveSignatureSalientFloorSelectsWhichBM25BarApplies` has been failing roughly one full-suite run in five since #293 merged. It is my test, from that PR, and it has been failing on `main` — not pre-existing background noise.

It asserts that a long pane-tail one-word variant does **not** merge, on the strength of `0.6567` sitting below the `0.70` `bm25_highbar_score`. That margin is **0.043**, well inside the excursions this scorer produces.

## Why it was hard to see

It passes cleanly in isolation. I ran it **40 times alone: zero failures** — which is exactly why it kept getting waved through as "known daemon flakiness" on re-run.

It needs CPU contention, which a full suite supplies. Reproduced deliberately with 8 busy-loops on 4 cores:

```
before fix:  5 failures in 50 runs under load
```

Every failure identical:

```
merged = true, want false (method "bm25", score 0.8082; min 0.35, high 0.70)
```

## Cause

`MatchText` normalizes a hit by dividing it by a **second** search (`textCandidates` → `textSelfScore`, `match.go:416`) that is not snapshot-consistent with the first. Seeding through `resolveSignature` grows the index by repeated `matcher.Add`, leaving many small scorch segments, and a background merge landing between those two searches inflates the ratio.

Measured over 40 identical trials on the same corpus:

| index built by | score distribution |
|---|---|
| `Rebuild` (one batch) | `0.657887` — every single trial |
| `Add` (incremental) | `0.658` in 80%; then `0.766` / `0.813` / `0.861` / `0.944` / **`1.000`** |

## The fix

`compactMatchIndex` rebuilds from the store in one batch before the decisive resolve, so the test measures the threshold semantics it is about rather than segment timing — the same reason the `internal/match` characterization tests build in one batch.

```
after fix:  0 failures in 50 runs under the same load
            0 failures in 75 runs under load (independent verification)
```

## What this does NOT fix — please do not let a green test hide it

This makes the **test** deterministic, not the **scoring**. A production index is `Add`-built (every minted signature calls `matcher.Add`), so the same inflation is reachable in production. Consequences, both recorded in the helper's doc comment so a passing test is not misread as evidence the normalization is sound:

- `bm25_highbar_score` **bounds** pane-tail drift rather than guaranteeing it.
- This is precisely why a structured salient refused by cosine is closed by a **rule** (`daemon.bm25RetryAllowed`) rather than held to a threshold — a threshold cannot be trusted against a score that intermittently reaches a perfect 1.0.

The underlying snapshot-consistency defect in `internal/match` is unfixed and still needs its own issue. This is the second concrete cost it has incurred.

## Scope

Test file plus a changelog fragment. No production code changes.

### DIFF
--- a/changelog.d/fix-flaky-bm25-bar-test.md
+++ b/changelog.d/fix-flaky-bm25-bar-test.md
@@ -1,0 +1,1 @@
+- Fixed a test that could fail spuriously on a busy machine, making CI runs fail for reasons unrelated to the change under review.
--- a/internal/daemon/cosinemiss_test.go
+++ b/internal/daemon/cosinemiss_test.go
@@ -571,6 +571,7 @@ func TestResolveSignatureSalientFloorSelectsWhichBM25BarApplies(t *testing.T) {
 
 			seedIdleRules(t, d, 24) // measure in the production regime
 			learned := d.resolveSignature(ctx, cfg, storedSig, storedSit)
+			compactMatchIndex(t, d)
 			got := d.resolveSignature(ctx, cfg, swappedSig, swappedSit)
 
 			merged := got.Signature == learned.Signature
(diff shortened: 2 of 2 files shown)


0xGosu:
@CharlieHelps please review — a flaky-test fix, so the evidence matters more than the diff.

The claim is that this test was **not** background flakiness: it fails ~20% of full-suite runs because it asserts a merge verdict on a score with a 0.043 margin, and that score inflates on incrementally-built indexes. Reproduced deliberately (5 failures in 50 under CPU load, 0 in 40 without load, which is why re-running kept hiding it), then 0 in 125 after the fix.

Two things I would value your eyes on:

1. **Is compacting the index the right fix, or does it hide the property under test?** The argument is that the test is about which *threshold* applies, not about segment-merge timing, so removing the timing variable sharpens it. The counter-argument is that production indexes are `Add`-built, so the test now asserts something production does not strictly guarantee. I chose to fix the test and document the gap loudly rather than weaken the assertion — push back if you think that trade is wrong.

2. **The pre-existing defect it exposes.** `MatchText` normalizes by a second, non-snapshot-consistent search (`match.go:416`), so the same pair scored 0.658 in 80% of trials and up to a perfect 1.000 in the rest. That is unfixed here and needs its own issue. If you think it should block this PR rather than follow it, say so.

PR #299, opened 2026-07-31
#1 fix: restore agent sends broken by herdr 0.7.5 removing `agent send`

0xGosu:
## What broke

herdr **0.7.5 deleted the `agent send` verb outright**. Every reply hap delivered exited 2 with a usage banner and nothing reached the agent — reported live on 0.5.19 as `herdr agent send: exit status 2`.

```
$ herdr agent send x y ; echo $?
herdr agent commands:
  ...
2
```

## The fix is not a rename

`agent send` quietly did two things, and the two survivors each do one. `submitText` now routes on the content:

| content | verb | why |
|---|---|---|
| single-line | `pane send-text` + `pane send-keys enter` | literal terminal input, so a **menu digit arrives as the key it is** |
| multi-line | `agent prompt` | bracketed-paste aware, so a task hand-out arrives as **one** message |

The single-line route is safety-critical. hap answers an approval by mapping the chosen option to its digit (`domain.MenuKeystroke`). Verified live (2026-07-31) against a real Claude question form:

- `pane send-keys 2` / `pane send-text 2` → **Banana** (correct)
- `agent prompt '2'` → pastes `2` as text; its Enter commits whichever option the caret was on → **Apple**

That is a silently wrong answer with a success exit code, so routing a digit through paste is never acceptable. Conversely `pane send-text` is not paste-aware — each embedded newline is a literal Enter, which would split a multi-line task into several truncated messages.

Routing trims a trailing newline first: nothing upstream trims it (`DeliverOutbound` returns the action unchanged when no menu is parsed, and actions can come from an LLM or `hap resolve --action`), so a one-line answer carrying one would otherwise take the paste route.

Both routes fall back to legacy `agent send` **only on exit status 2** — herdr rejecting the verb — which keeps `min_herdr_version = 0.7.0` honest. A pane-level failure exits 1 with a JSON error body and is returned as-is, so a real delivery error is never retried as a duplicate send. The fallback earns the cached shape only when it actually delivers, so a herdr where neither verb exists re-probes rather than wedging every later send.

## Also fixed, found alongside it

- **Confirming an escalation with no suggestion** said only `carries no suggestion to confirm; use resolve with an explicit action`. It now names the control that withheld it and the exact `hap resolve <id> --action TEXT --send` / `hap dismiss <id>` commands. It only blames a control when one of the four deliberate vetoes really withheld an answer — a variance guard over an unfamiliar option set resolved *nothing*, and saying otherwise sends the operator hunting for a safety rule to relax. (The vetoes stay non-confirmable; that invariant is unchanged and still tested.)
- **The plugin log grew without bound** — a live state directory reached **1.9 GB**. The daemon holds the file open for its whole life, so an open-time rotation (what the captured stderr log does) never fires for the process writing nearly all the volume. Rotation is now enforced on write, to a single `.old` sibling at 64 MiB.
- **Config deprecation warnings fired on every `Load`** — every daemon reload nudge, TUI refresh and CLI invocation. One tail window of that log held 1842 copies of three lines. Once per process now.
- **The integration harness broke the same way**: herdr reshaped `agent start` to `--kind KIND --pane ID`, so all five real-herdr cases failed. Updated to create a pane first, then start the agent (or `pane report-agent` a script stand-in).

## Verification

- `go test -tags "vectors cpu" ./... -count=1` — **3204 passed, 0 failed, 5 skipped** (30 packages).
- `golangci-lint run --build-tags "vectors,cpu"` — 0 issues.
- `go test -tags "integration vectors cpu" ./test/integration/` — **passes against real herdr 0.7.5**, including `TestRealConfirmDeliversMenuDigit`, the end-to-end regression for this exact bug. It failed before this change.
- Live-verified against a real Claude agent: single-line digit delivery selects the chosen option, and multi-line arrives as one message.

New invariant tests: `TestSingleLineSendTypesTheTextSoAMenuDigitSelects`, `…NeverPastes`, `…TrimsATrailingNewlineInsteadOfPasting`, `TestMultiLineSendPastesAsOneMessage`, `…DoesNotLatchLegacyWhenTheFallbackAlsoFails`, plus legacy-fallback coverage on both routes.

## Note

`agent prompt` also refuses a pane whose agent is not its foreground process (`agent_not_ready`, exit 1), and does not commit while an agent is still starting up. The existing status-gated retry-Enter loop covers the latter; the former surfaces as a delivery failure and escalates, which is the existing fail-safe. Both are documented in CLAUDE.md.


0xGosu:
@CharlieHelps please review

PR #301, opened 2026-07-31
#1 fix: answer the option the operator chose, not option 1

0xGosu:
## The bug

An approval answered with **"Yes, and don't ask again"** committed plain **"Yes"** instead.

Claude Code renders that option with a **typographic apostrophe** — `Yes, and don’t ask again for: npm *` (U+2019) — while every source of a chosen reply (a learned rule, an LLM answer, this repo's own classifier fixtures) writes the ASCII `don't`. Comparing raw, the label matched no option, `MenuKeystroke` fell through to sending the reply as **literal text**, and on a standing menu that is not a harmless no-op: the agent ignores the letters and the trailing Enter commits whichever option the caret rests on — **the first one** — silently, with a success exit code.

## Verified live

Against **Claude Code 2.1.220 in a real herdr pane**, both before and after the fix:

| | before | after |
|---|---|---|
| ASCII reply naming option 2 | ran the command under plain "Yes"; the next command **prompted again** | selected option 2; the next command **did not prompt** |
| reply matching no option | ran the command under plain "Yes", reported success | **refused**, menu left standing, nothing committed |

## The fix

- **`domain.FoldMenuText`** folds typographic punctuation, case and whitespace for every label *comparison*. Nothing delivered is ever folded. The fold set is narrow on purpose — a backtick, an acute accent and an ellipsis are left alone, since each widening brings two genuinely different labels one character closer to comparing equal.
- **`domain.UnmatchedMenuReply`** fails closed: a reply matching none of the offered options escalates or refuses instead of being typed. **All four send paths** honour it — `daemon.act`, the LLM promotion in `handleLLMOutcome`, the LLM-rewritten reply in `handleActionReviewOutcome`, and `deliver.Deliver` for operator-confirm/auto-accept.
- The rewritten reply in `handleActionReviewOutcome` is now mapped to the option's **digit** like every other path, and re-checked against the live pane it was rewritten for.
- An **unreadable pane** refuses a menu answer when the decision's own capture proves a menu was standing; with no such evidence the literal send stands, so legacy rows carrying no excerpt behave as before.
- Label matching is **unique-or-refuse on the exact pass** as well as the prefix pass: one capture can hold two renders of a menu that number the same label differently, and returning the first hit delivered a stale digit while reporting a clean mapping.

### Ordering that is deliberate

The gate runs **after** the multi-tab answer-series and remote-environment branches (each answers its own protocol), and **after** `llm.enable_rewrite_action` dispatches in `act` — adapting a drifted label to the live options is exactly what the rewrite is for, and its result is re-checked downstream, so nothing skips the gate by going that way.

### Accepted trade-offs

- An approval whose real prompt is a bare `y/n` while unrelated numbered lines sit in the scrollback now escalates instead of typing `y` — fail-closed, and visible to the operator either way.
- An unreadable pane only refuses when there is menu evidence in the capture.

## Tests

New: `TestMenuKeystrokeFoldsTypographicPunctuation`, `…FoldKeepsDistinctOptionsDistinct`, `…DuplicateRendersRefuse`, `TestFoldMenuText`, `TestUnmatchedMenuReply`, `TestDeliverMatchesLabelAcrossTypography`, `TestDeliverUnmatchedMenuReplyRefuses`, `TestDeliverUnreadablePaneWithMenuEvidenceRefuses`, `TestAutoActMatchesLabelAcrossTypography`, `TestAutoActUnmatchedMenuReplyEscalatesInsteadOfSending`, `TestLLMPromotionUnmatchedMenuReplyRejects`.

`go vet` and `golangci-lint` clean. The only full-suite failures are the two documented flaky categories (arbitrary `internal/daemon` timing flakes and the load-sensitive embedder stall guard) — each passes in isolation, and a different test fails per run.


0xGosu:
@CharlieHelps please review

PR #303, opened 2026-08-01
#1 fix: reflect Tasks-tab toggle/delete/add/edit immediately instead of after the slow refresh

0xGosu:
## Problem

Toggling a task's done status in the TUI (`d` on the Tasks tab) wrote the file, but the checkbox kept its old state until the next full refresh round trip (daemon `GetStatus`, per-agent pane reads, store queries) — often a second or more. Operators read that as "nothing happened", pressed `d` again, and flipped the task straight back. Same lag applied to `x` (delete), `a` (add), and `e` (edit).

## Fix

Two layers, both cheap:

1. **Optimistic flip at the keypress** — `flipTaskCheckboxes` flips the checkbox in the loaded snapshot immediately, matched by canonical path + item number + item text (the same identity the write's expected-text guard enforces, so a shifted file is left alone). The mark is SET from the target's prior state, never inverted in place — `applyTaskLists` aliases one items array across duplicate-path groups, and a per-group in-place invert would net out to a no-op (documented in the code and pinned by a test).
2. **The write's own result is the refresh** — `SetTaskDone`/`DeleteTask`/`AddTask`/`EditTask` already return the renumbered checklist; the action commands now carry it back in `actionResultMsg.taskLists` and the handler applies it directly. The follow-up full `refresh()` still runs for global consistency.

`applyTaskLists` deliberately runs before the error branch: a partial bulk-toggle failure still carries the authoritative list for the writes that landed, so the succeeded items stay flipped and the failed item shows the file's real state alongside the "toggled N, skipped #x" error.

## Tests

- `TestTasksTabToggleReflectsImmediately` — checkbox flips before the async write runs; second press toggles back with no refresh in between
- `TestTasksTabToggleDuplicatePathDoubleToggleNoRefresh` — pins the set-semantics/aliasing contract for duplicate-path sources
- `TestTasksTabTogglePartialFailureShowsTruth` — pins apply-before-error-branch ordering
- `TestTasksTabDeleteReflectsImmediately` — deleted row disappears and survivors renumber without a refresh
- immediacy assertions added to the existing add/edit prompt tests

Full suite run: all packages pass; the only failures were the two documented load-induced `internal/embedder` real-model stall-guard flakes (host load ~28 on 4 cores at the time; both pass in isolation). `golangci-lint` clean on the touched package.


0xGosu:
@CharlieHelps please review

PR #305, opened 2026-08-02
#1 perf: cut idle `hap tui` CPU ~6x (config parsing + N+1 decision reads)

0xGosu:
Started from "why are there so many `hap` processes eating CPU". There were only three, and two were idle — the whole cost was the TUI, doing the same wasted work every 2 seconds.

## Result

Idle TUI CPU, isolated environment, identical build flags, against a copy of a real 196-rule state directory:

| build | CPU per 2s tick | % of a core |
|---|---|---|
| `main` | 528ms | **26.4%** |
| this branch | 85ms | **4.2%** |

`hap signatures list` on the same data: **1166–1497ms → 152–314ms**.

For scale: the `hap daemon` and `hap embed-worker` beside it measured **0.0%** at idle, and `herdr server` burns ~35% of a core independently of hap (verified with zero hap calls in flight — hap's polling adds ~17ms per poll).

## Three separate bugs

**1. `App.Signatures` re-parsed `config.toml` once per rule.** It resolved the operator-confirmation weight *inside* the per-signature loop; that call chains to `config.Load`, which is deliberately uncached. 196 rules × 9ms, every 2 seconds. The weight cannot change mid-listing.

**2. `config.Load` parsed the same file eight times.** One decode into `Config` plus seven "probe" decodes, each a full re-parse of the whole file to detect one deprecated or removed key. BurntSushi's lexer is a goroutine/channel pipeline, so each pass is real work. One shared probe now answers all of them: **3.83ms → 1.3–1.8ms**, allocations 3502 → 1035. Every `hap` command and the daemon pay this, not just the TUI.

**3. `App.Signatures` issued two SQL queries per rule** — ~400 round-trips every 2 seconds at idle. Now two bulk queries.

## The parts that needed care

**The probe merge is only safe because every probe field is type-free.** BurntSushi aborts the *whole* decode on the first type mismatch, and it walks the file's key map — so which fields were populated before an abort varies run to run. With nine probes sharing one decode, an ill-typed key that hap otherwise ignores (`gpu_layers = 0.5`) could intermittently swallow the `[thresholds]` → `[confidence_thresholds]` migration and silently reset an operator's confidence gates. `*any` cannot mismatch. The two table probes are `*any` as well, not `*map[string]any`: a map field still has to match structurally, and `thresholds = 5` survives today only on a dead branch in toml v1.6.0 — a fixed upstream would quietly reinstate the coupling. The table assertion lives in `legacyValue`, where a failure is local.

The one probe whose *values* are migrated is read by a separate decode that runs only when that table is present, so neither it nor the removed-key warnings can be collateral damage for the other. An unreadable `[thresholds]` now warns instead of vanishing on the next save.

**Both bulk reads require the signature list.** Ranking the whole table would trade a cost bounded by the listing's size for one that grows with total decision history forever — and nothing prunes `decisions`. Chunked `IN (...)`, with the input deduped first: `IN` dedupes *within* a chunk, but a signature repeated *across* chunks would append its history twice and double-count every decision in the confidence computed from it.

`BatchDecisionReader` is an **optional** interface, type-asserted with the per-signature loop as fallback, per the repo's "optional capabilities are optional interfaces" rule — no existing fake breaks.

## Verification

Beyond the tests, the config change was checked behaviorally against `main`: a config holding **every** legacy key, and a second where legacy and canonical keys are both set, produce **byte-identical warnings (11) and identical `hap config show` output**.

Tests added:
- batch vs per-signature equivalence, spanning the 50-row window boundary (over-cap, under-cap, and no decisions)
- the multi-chunk merge with a duplicate straddling the boundary — **verified to fail without the dedupe**
- batch vs fallback agreement, via a store wrapper that embeds the interface to defeat method promotion
- canonical `[confidence_thresholds]` beating the legacy table (nothing covered this before, and it is the assumption the probe merge rests on)
- an ill-typed removed key not breaking the migration, looped 50× because the failure was nondeterministic
- a scalar written under a table name being ignored rather than fatal

Full suite: the six touched packages (`config`, `store`, `ports`, `frontend`, `cli`, `tui`) pass clean. Three failures elsewhere were the two documented flaky areas under load ~31 (`internal/daemon` timing, `internal/embedder`'s fixed 2s stall guard) and all passed on isolated re-run. `go vet -tags "vectors cpu" ./...` and `gofmt` clean.

## Deliberately not done

The remaining 85ms/tick is spread thin: `Signatures` ~24ms, herdr CLI spawns 9ms of CPU (116ms *wall* — I/O-bound, not CPU), the rest in rendering and GC. The big lever left is **skipping the refresh when nothing changed**, which would take idle to near zero — but it has to detect changes across signatures, decisions, audit, kill events, agent names, task files, config, and the herdr agent list, and missing one signal means the operator console silently goes stale. That trade-off deserves an explicit decision rather than being slipped into a perf PR.

Also left: `Load` still parses twice (main decode + probe). `MetaData.IsDefined` could make it one, but it is case-sensitive where struct matching falls back to `EqualFold`, so `GPU_LAYERS = 1` would stop being detected — a behavior change, not a free win.


0xGosu:
@CharlieHelps please review

PR #309, opened 2026-08-03
#1 fix: bound hap's own log and audit footprint

0xGosu:
## 📌 Background

An operator reported hap's logs consuming too much disk. Investigation of a live state directory (14 days of light use) found the space is **not** where it looks.

`herd-auto-prompter.log` — the plugin's own slog file — is **not** the problem. The 0.5.19 fix already capped it, and it held 735 lines / 256 KB over the whole period.

## 🔧 Reason for Changes

The space was going to things nothing bounded:

| Consumer | Measured | Bounded before? |
|---|---|---|
| `audit_log` table | **4.4 MB** of the 8.2 MB DB, ~58 rows/day | ❌ nothing ever pruned it |
| DB WAL | **4.0 MB** stuck at its high-water mark | ❌ `journal_size_limit = -1` |
| `daemon.stderr.log` | ~100% llama.cpp banner, ~11 KB per embed-worker start | ⚠️ rotated only at daemon **spawn** |

`audit_log` is 809 rows at 5.0 KB/row, and `pane_excerpt` alone is **3.8 KB/row — 76% of the table**. `store.go` already noted in a comment that nothing prunes it.

Two log lines that fire while nothing happens turned out to be **functional bugs**, not just noise:

- `events.go:127` logged every ordinary pane split as "connection lost" **and advanced the reconnect backoff**. The healthy-stretch reset needs a full uninterrupted minute, so routine churn ratcheted the delay to its 30 s ceiling — hap went *blind to agent status* for that long after a normal split.
- `autosendidle.go:245` logged ~1/min per task source forever; its `break` precedes `noteRedrive`, so nothing ever aged it out.

## ✅ Changes Implemented

- **`hap gc`** (`--days N`, `--dry-run`) plus a daily daemon sweep: blanks `pane_excerpt` on aged audit rows while **keeping the rows**, so `hap audit` history stays complete.
- **The exclusion rule is a safety control, not tidiness.** `pane_excerpt` is still read on live paths: `AutoAcceptableEscalations` → `autoAcceptDeliver` passes it into `deliver.Request`, where an unreadable pane refuses delivery *only* when that excerpt proves a menu was standing. Blanking the wrong row converts a safe refusal into a literal send at a live menu — which commits option 1. So rows in `escalated` **and `auto_accepting`** (the same row mid-delivery) are spared at **any** age, along with unprocessed LLM retries and recently corrected rows. The cutoff is clamped so a misconfigured `0 days` cannot reach live rows.
- `PRAGMA wal_checkpoint(TRUNCATE)` after VACUUM, and `journal_size_limit`. Without the checkpoint the reclaim measures *bigger* than it started — VACUUM writes the rebuild through the WAL.
- `daemon.stderr.log` now rotates from the **running** daemon (truncate, not rename: fd 2 is already bound to that inode, and the daemon is detached so no parent can wrap it). The llama.cpp model-load banner is silenced — ~250 lines per embed-worker start, which the file was almost entirely made of.
- Expected resubscribes no longer read as dropped connections; the backoff is reset rather than ratcheted.
- New `[logging]` section: `level`, `max_size_mb`, `audit_excerpt_retention_days`. There was previously no way to turn the log *down* — only `HAP_DEBUG=1` to turn it up, and that never applied to `hap tui`, which writes to the same file. Default cap drops 64 MiB → 16 MiB.
- `hap status` gained a `disk:` line.

## 🔍 Testing Results

- [x] Unit tests passed — full suite green: **30 packages, 0 failures**.
- [ ] Integration tests passed — not run (no live herdr in this environment).
- [x] Manually tested in local/dev environment.

Verified against a consistent snapshot of the real database (`--days 1`):

- **12.2 MiB → 5.5 MiB**
- **814 rows before, 814 after** — nothing deleted
- all escalated rows kept their excerpts
- `rationale`, `suggestion`, `action_or_escalation`, `status`, `sig_raw`, `llm_output` all untouched

The `auto_accepting` guard was **mutation-tested**: reverting the SQL makes the new test fail, restoring it makes it pass.

`gofmt` / `go vet` clean; **golangci-lint 0 issues**; builds under all four tag combinations.

- No schema migration. Retention only blanks a column.
- New `[logging]` keys are optional; absent config keeps working. **Set `audit_excerpt_retention_days = 0` to disable the sweep entirely** (the field is a pointer precisely so an explicit `0` survives `fillZeroes`).
- `level` and `max_size_mb` are read **once per process** — a running daemon keeps its level through a config reload; restart with `hap daemon --ensure` to apply.
- The default log cap drops from 64 MiB to 16 MiB, so an untouched install reserves 32 MiB instead of 128 MiB counting the `.old` sibling.

---

### 📝 Additional Notes

Two accepted trade-offs, documented in code rather than hidden:

1. `frontend.Resolve` (`hap resolve --send`, TUI confirm, MCP `submit_decision`) is a **third** reader of `pane_excerpt` and works on terminal rows — exactly what this blanks. Not excluded, because those rows are 87% of the win. The exposure is narrow: the excerpt is consulted only when the *live* pane read fails, and a blanked row then behaves like a pre-migration row, which `deliver` already handles by design.
2. A stderr rotation loses anything written between the copy and the truncate (microseconds, and only past the cap), and carries only the trailing `StderrLogCap` bytes into `.old` — a crash trail is a tail.

Out of scope by the operator's decision: transcripts written by the LLM CLIs hap spawns (23 MB and growing) belong to that tool, not to hap.


0xGosu:
Updated the meaning of `audit_excerpt_retention_days = 0` per review (2544fe2).

**Before:** `0` disabled the sweep (keep every excerpt forever), and `hap gc --days 0` did something different again — it fell through to the config value rather than pruning anything.

**Now**, and the two agree:

| Value | Meaning |
|---|---|
| omitted | 14 days (the default) |
| `0` | keep **no** excerpts — blank every eligible row |
| negative (e.g. `-1`) | never prune |

`0` reads the way it looks — retain for zero days — so it is the most aggressive setting rather than the off switch. Negative is the off switch, matching the `journal_size_limit = -1` convention already in this PR. `--days` gained a sentinel so `0` cannot double as "not given".

**The safety floor is what makes `0` safe to offer**, and there is a new test for exactly that (`TestPruneAuditExcerptsZeroRetentionStillHonoursEveryExclusion`): the exclusions do not scale with the window. At retention `0` a pending escalation, a row mid auto-accept, and a row with an unprocessed LLM retry all still keep their excerpts, because that excerpt is the evidence a menu was standing. The cutoff is also still clamped to `AuditExcerptDedupMargin`, so even `0` cannot reach a row the daemon is comparing against this second.

Verified on a snapshot of a real database — `0` blanks 739 rows vs 724 at 1 day, reclaiming 3.2 MiB, with wording that is actually a sentence:

```
Dry run: would blank EVERY eligible pane excerpt (retention is 0).
Cleared pane excerpts on 739 audit row(s) — retention is 0, so none are kept.
```

`hap status`, `hap config fields`, sample config, README and the changelog fragment all updated. golangci-lint 0 issues; config/store/cli/frontend all pass.

PR #313, opened 2026-08-03
#1 fix: record the LLM session id on delivered rows, not just escalations

0xGosu:
## 📌 Background

Follow-up to #310, found by **live-testing that feature against a real herdr + claude agent** — the unit suite did not catch it.

#310 gave every LLM invocation a session id and recorded it on the audit row, so a decision could be traced to the transcript the CLI wrote. Driving a real agent showed it only half worked.

## 🔧 Reason for Changes

The id was stamped only on the `escalate()` paths. A **delivered** LLM decision — audit status `auto`, the most common LLM outcome — carried nothing:

```
#827  escalated  idle    act=escalated      bd207bdb-be35-47e8-bcfd-a3b6bc3c2b16
#828  auto       choice  act=auto:Apple     *** EMPTY ***          <-- the gap
#829  escalated  idle    act=escalated      902fc941-b89c-4bef-b809-e14c8fae6a45
```

Its `llm_requests` row held the id the whole time; the audit row it produced was blank. So exactly the rows an operator most wants to trace — the ones where hap acted on its own — could not be tied to their transcript.

A systematic scan (`AuditRecord` literals carrying `LLMOutput`/`LLMConfidence` but no `LLMSessionID`) found **7 such sites**, not one.

## ✅ Changes Implemented

Every audit row that carries LLM detail now carries the id:

- both promotion paths in `handleLLMOutcome` — delivered action, and no-op
- the action-review no-op, and the autonomous delivery behind it
- the multi-tab series and remote-environment deliveries in `sweep.go`
- the task-list review rows
- `auditAgentDisabled`, which records the action that *would* have happened

**Design note.** `domain.LLMDecision` gains a **transient** `SessionID` that the adapter stamps beside `CapturedOutput`, so the delivery paths holding a decision (but not a request) can name the conversation without threading a parameter through four more signatures. It is deliberately not a column — reading a decision back from the store yields an empty one.

Only `withAgentAutomation` / `auditAgentDisabled` take a real new parameter, because the paths reaching them hold neither a request nor a decision. Their non-LLM callers pass `""` explicitly rather than relying on a zero value, so the intent reads at the call site.

`LLMRequest.SessionID` remains the single source of truth: `consultWithSession` writes the *effective* id back onto the request (hap's own for a CLI that accepts one, the CLI's own for `codex`, which mints its own), so it is correct for both CLIs and on the failure path.

## 🔍 Testing Results

- [x] Unit tests passed — full suite green apart from two **known pre-existing** daemon flakes (below); both new tests pass.
- [x] Manually tested in local/dev environment — **this is how the bug was found**.

**Live verification** against a real herdr claude agent, after hot-swapping the fixed daemon and driving a second approval:

```
#828  auto  act=auto:Apple  llmconf=92  *** EMPTY ***          <- old build
#830  auto  act=auto:Blue   llmconf=92  4c57e625-7dff-416a…    <- fixed build
```

That transcript exists (36 KB) and contains the exact consult:
`{'select_options': [3], 'confident_score': 92, 'rationale': 'Harmless preference question…'}` — matching the audit row's `act=auto:Blue llmconf=92`.

New tests in `internal/daemon/sessionid_audit_test.go`:

- `TestLLMAutoRowCarriesSessionID` — **mutation-tested**: removing the stamp makes it fail with `got "" want "3fb15733-…"`, restoring it passes.
- `TestNonLLMRowCarriesNoSessionID` — a learned/operator row must stay empty rather than inherit an id.

Also re-ran the scan after the fix: **0 unstamped LLM audit rows remain**.

**The two daemon failures are pre-existing flakes, verified:**

| Test | In full run | In isolation |
|---|---|---|
| `TestAutoSendIdleBelowThreshold` | FAIL | **PASS** |
| `TestManualCaptureRecognizesIdleCodexPlanApproval` | FAIL | **PASS** |

`TestAutoSendIdleBelowThreshold` was separately baselined on **untouched `main`** earlier today: 3 runs of `./internal/daemon/`, 1 failure — the same test, the same rate, on code containing none of these changes.

No migration — the columns already exist from #310. No config change. Rows written before this fix keep an empty id; nothing is backfilled, because there is nothing to backfill it from.

---

### 📝 Additional Notes

Worth recording as a process point: this class of bug is invisible to the unit suite, which fakes herdr and never exercises the real delivery path end to end. The `CLAUDE.md` recommendation to run against a live herdr after finishing a feature is what surfaced it.


0xGosu:
Follow-up: **codex had not been tested against the real CLI** — only a shell stub and a banner pasted from 0.143.0. Now verified against real codex-cli **0.146.0** (8f3c994).

**It works.** Driving the actual binary through the adapter:

```
```

and that id names a real 56 KB transcript containing that consult. hap passed no `--session-id` (correct — codex has no such flag), and the id it read back beat the one hap had minted.

**The real run surfaced a trap worth its own test.** codex prints an unrelated `shell_snapshot` ERROR line *before* its banner, and that line embeds **the same UUID inside a path**:

```
... ERROR codex_core::shell_snapshot: ... /root/.codex/shell_snapshots/019fc84d-…-8ea2c12822f4.tmp-1785771960838110783: line 1327: syntax error …

OpenAI Codex v0.146.0
--------
session id: 019fc84d-a8b8-77f2-8e20-8ea2c12822f4
```

A pattern that searched for any uuid, or matched `session` anywhere on a line, would read from the wrong place on a completely ordinary run. The existing line-anchored match already handles it; it is now pinned by `TestExtractSessionIDFromRealCodex0146` and `TestExtractSessionIDIgnoresTheSnapshotPath`, using the verbatim capture as the fixture.

**Also documented: the two CLIs do not agree on where transcripts go.**

| CLI | Transcript path |
|---|---|
| `claude` | `<CLAUDE_CONFIG_DIR>/projects/<slugified-cwd>/<session-id>.jsonl` |
| `codex` | `<CODEX_HOME>/sessions/<YYYY>/<MM>/<DD>/rollout-<ISO-ts>-<session-id>.jsonl` |

So the future cleanup command needs a **suffix glob** for codex, not an exact filename — and codex's store was the *larger* of the two on the machine this was verified on (**120 MB** vs 24 MB). Both are recorded in `sessionid.go` and the README so that work does not have to rediscover it.

One unrelated observation for a possible separate issue: codex refuses to run with `Not inside a trusted directory and --skip-git-repo-check was not specified`. hap runs the LLM CLI from `Adapter.WorkDir()`, which is the state dir — not a git repo — so a codex-configured operator likely needs `--skip-git-repo-check` in their `llm.command`. Not touched here.

PR #315, opened 2026-08-03
#1 fix: read a reported session id from the full CLI output, not the audit copy

0xGosu:
## 📌 Background

Follow-up to #310 / #313, from live-testing the session-id capture against the **real `codex` CLI** (codex-cli 0.146.0).

Scope is narrow and deliberate: capturing the session id correctly for each LLM invocation and persisting it. Nothing here is about where transcripts live.

## 🔧 Reason for Changes

`runConsult` merges stdout and stderr into a single buffer and keeps the first 16 KiB for the audit row:

```go
raw := out.String()
captured := truncate(raw, 16*1024)   // head, not tail
```

Extraction read `captured`. For a CLI that **reports its own id** rather than accepting one — `codex` — enough output before the announcement pushes the banner past the cap, and hap falls back to the id it minted itself.

That fallback is silently **wrong** for codex: codex ignores anything hap passes and mints its own, so the recorded id names a conversation that never existed. No error, no warning — just a column that cannot be reconciled with anything.

It is not reachable with codex today (it prints its banner at startup, and the truncation keeps the head), but stdout and stderr share one buffer, so the ordering that protects it is incidental rather than guaranteed.

## ✅ Changes Implemented

One line: extract from the raw output rather than the truncated audit copy. The audit row still stores the 16 KiB `captured` copy exactly as before — only the scan changed.

## 🔍 Testing Results

- [x] Unit tests passed — `internal/llm` green.
- [x] Manually tested in local/dev environment — real `codex` 0.146.0 driven through the adapter.

**Live verification** (before this fix, on the existing code) that codex capture works at all:

```
```

hap passed no `--session-id` (correct — codex has no such flag), and the id it read back beat the one hap had minted.

**New test:** `TestConsultReadsCodexSessionBehindBulkyOutput` puts 20 KiB on stdout before the banner. **Mutation-tested** — restoring the truncated read makes it fail exactly as the bug would:

```
session id = "11111111-2222-4333-8444-555555555555",
        want "019fc84d-a8b8-77f2-8e20-8ea2c12822f4"
```

which is the failure mode itself: hap's own id, silently recorded, for a session codex never had.

None. No schema change, no config change, no behaviour change for `claude` (which is passed an id and never reports one).

---

### 📝 Additional Notes

Unrelated, and not touched here — worth its own issue if it bites anyone: codex refuses to start with `Not inside a trusted directory and --skip-git-repo-check was not specified`. hap runs the LLM CLI from `Adapter.WorkDir()`, which resolves to the state directory — never a git repo — so a codex-configured operator likely needs `--skip-git-repo-check` in their `llm.command`.

### DIFF
--- a/changelog.d/fix-codex-session-id-untruncated.md
+++ b/changelog.d/fix-codex-session-id-untruncated.md
@@ -1,0 +1,1 @@
+- Fixed the session id being read from the truncated copy of an LLM CLI's output rather than the whole of it. For a CLI that reports its own id (`codex`), enough output before the announcement would push it past the 16 KiB the audit row keeps, and hap would silently record the id it had minted instead — naming a conversation that never existed, with no error to explain it.
--- a/internal/llm/llm.go
+++ b/internal/llm/llm.go
@@ -307,7 +307,8 @@ func (a *Adapter) runConsult(ctx context.Context, spec commandSpec, self string,
 	runErr := cmd.Run()
 	elapsed := time.Since(started)
 
-	captured := truncate(out.String(), 16*1024)
+	raw := out.String()
+	captured := truncate(raw, 16*1024)
 
 	// Regardless of exit status, the authoritative signal is the staged
 	// submission in the DB.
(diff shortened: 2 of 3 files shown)


0xGosu:
Folded in recovered work (f0b6067).

The codex verification tests were pushed to `fix/llm-session-id-auto-rows` **five minutes after #313 had already been squash-merged**, so they belonged to no PR and never reached main. Same topic as this branch, so they ride along here rather than opening a third PR.

What was missing from main:

- `TestExtractSessionIDFromRealCodex0146` — extraction pinned against a verbatim capture from a real codex-cli 0.146.0 run
- `TestExtractSessionIDIgnoresTheSnapshotPath` — the trap: codex prints an unrelated `shell_snapshot` ERROR line **before** its banner, and that line embeds **the same UUID inside a path**. A pattern searching for any uuid, or matching `session` anywhere on a line, would read the wrong one on a completely ordinary run. The line-anchored match already handled it; now it is a test.
- `realCodexStderr` — the fixture, kept verbatim so the banner format has to be re-proved as codex moves (the older fixture came from 0.143.0).

All six `ExtractSessionID` tests pass; golangci-lint 0 issues.

Not carried over, per the narrowed scope of capture-and-persist: the transcript-path documentation (which directory each CLI writes to). It is still on `fix/llm-session-id-auto-rows` at commit `8f3c994` if it is wanted later.

PR #317, opened 2026-08-04
#1 test+docs: pin codex session-id extraction against a real 0.146.0 run

0xGosu:
## 📌 Background

Tests only — no production code changes.

Follow-up to #310 / #313 / #315. The `codex` extraction path had only ever been exercised against a **shell stub** and a banner pasted from codex-cli **0.143.0**. Driving the real binary through the adapter confirmed it works, and surfaced something worth pinning.

## 🔧 Reason for Changes

Codex prints an unrelated `shell_snapshot` error **before** its banner, and that line embeds **the same UUID inside a file path**:

```
2026-08-03T15:46:01Z ERROR codex_core::shell_snapshot: Shell snapshot validation failed:
  … /root/.codex/shell_snapshots/019fc84d-a8b8-77f2-8e20-8ea2c12822f4.tmp-1785771960838110783: …

OpenAI Codex v0.146.0
--------
session id: 019fc84d-a8b8-77f2-8e20-8ea2c12822f4
```

A pattern searching for any uuid, or matching `session` anywhere on a line, would read from the wrong place on a **completely ordinary run** — no error, just a wrong id. The existing line-anchored match already handles it; nothing was broken. This makes that guarantee a test instead of an accident.

The banner format also has to be re-proved as codex moves: the previous fixture was 0.143.0, this one is 0.146.0.

## ✅ Changes Implemented

- `realCodexStderr` — a verbatim capture from a real 0.146.0 run, noise and all.
- `TestExtractSessionIDFromRealCodex0146` — extraction against that capture.
- `TestExtractSessionIDIgnoresTheSnapshotPath` — feeds only the pre-banner noise and asserts nothing is extracted, isolating the trap above.

## 🔍 Testing Results

- [x] Unit tests passed — all six `ExtractSessionID` tests green.
- [x] Manually tested in local/dev environment — real codex 0.146.0 driven through the adapter:

```
```

hap passed no `--session-id` (correct — codex has no such flag), and the id it read back beat the one hap had minted.

None. Test-only.

---

### 📝 Additional Notes

This work was stranded twice before landing here, both times by pushing to a branch whose PR had already been squash-merged — first to `fix/llm-session-id-auto-rows` five minutes after #313 merged, then to `fix/codex-session-id-untruncated` six hours after #315 merged. Neither push was ever part of a PR. It is on its own branch now.

Not included, per the narrowed scope of capture-and-persist: documentation of which directory each CLI writes transcripts to. That is on `fix/llm-session-id-auto-rows` at `8f3c994` if it is ever wanted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1331,6 +1331,16 @@ The id is recorded for failed consults too — a timeout or a no-submit still
 wrote a transcript, and still raises an escalation. It is bookkeeping only:
 nothing decides anything from it, and an empty value simply means unknown.
 
+Where the transcript lands differs per CLI, so anything looking one up cannot
+assume claude's layout:
+
+| CLI | Transcript path |
+|---|---|
+| `claude` | `<CLAUDE_CONFIG_DIR>/projects/<slugified-cwd>/<session-id>.jsonl` |
+| `codex` | `<CODEX_HOME>/sessions/<YYYY>/<MM>/<DD>/rollout-<ISO-ts>-<session-id>.jsonl` |
+
+codex therefore needs a suffix match rather than an exact filename.
+
 An optional **`command_start`** runs *instead of* `command` on an agent's
 **first consult** — the first time the plugin needs the LLM for that agent
 this daemon lifetime. Every later consult uses `command`. A genuinely new
--- a/changelog.d/test-codex-session-id-real-capture.md
+++ b/changelog.d/test-codex-session-id-real-capture.md
@@ -1,0 +1,1 @@
+- Added tests pinning `codex` session-id extraction against a verbatim capture from a real codex-cli 0.146.0 run, including the case where codex prints an unrelated error carrying the same UUID in a file path *before* its banner — where a looser pattern would read the wrong id on a completely ordinary run.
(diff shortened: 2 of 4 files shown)


0xGosu:
Added the transcript-path documentation (b036e16), recovered from `fix/llm-session-id-auto-rows` — it was the last content that existed only on that branch. **That branch has now been deleted**, and nothing was lost: the only other thing unique to it was the pre-#315 truncated read, i.e. the bug itself.

The two CLIs do not agree on where they write, and the difference is not guessable:

| CLI | Transcript path |
|---|---|
| `claude` | `<CLAUDE_CONFIG_DIR>/projects/<slugified-cwd>/<session-id>.jsonl` |
| `codex` | `<CODEX_HOME>/sessions/<YYYY>/<MM>/<DD>/rollout-<ISO-ts>-<session-id>.jsonl` |

So anything resolving a recorded id back to a file needs a **suffix match** for codex, not an exact filename. Verified live 2026-08-03 against claude and codex-cli 0.146.0 — and on that machine codex's store was the *larger* of the two (120 MB vs 24 MB), which inverts the assumption the original investigation started from.

Documentation only; nothing here resolves a path, consistent with the capture-and-persist scope.

The remote is now just `main` and this branch.

PR #325, opened 2026-08-10
feat: pluggable task-list storage — local_fs and github_gist

0xGosu:
Makes task-list storage pluggable and adds a second backend: a task source's markdown checklist can live in a GitHub gist instead of on the machine hap runs on. Designed as an adapter boundary so a third provider (Linear, Notion) is additive.

**Off by default.** `local_fs` remains the default and its behaviour is byte-identical — the daemon suite, `hap task-source list` output and the TUI picker are all unchanged for an install that never touches the setting.

## Shape

One interface — `ports.TaskStore` (`Read` + `Mutate`) — with two optional capabilities callers type-assert (`EnsureCreator`, `RemoteTaskStore`), matching the repo's existing optional-port convention. Adapters live in `internal/taskstore/{local,gist}`; `internal/taskstore.Registry` maps config to an adapter.

`Mutate` takes the transform rather than exposing `Get`/`Put` because every mutator hap has checks and claims in the same pass (`Reserve` = `ExpectText` + mark; `ApplyReview` re-resolves and re-screens; the review re-reads the kill switch inside the lock). A split would move those checks outside the critical section. Keeping `fn` inside also means a new adapter inherits the entire existing mutator layer for free.

`TestOnlyTheRegistryImportsAConcreteBackend` fails any production file outside `internal/taskstore` that names a concrete backend — verified by mutation. The failure it prevents is real: a `local.Store` used directly would bypass a configured remote provider and fork the operator's list in two. `internal/taskstore/adapter.go` documents the five-step recipe for the next provider.

## Config

`[task_source_provider]` is a **default**, and each `[[task_sources]]` entry may override `provider` / `gist_id` — so some agents can be local while others are in a gist at the same time.

An empty per-source key **is** the inheritance and is never materialized. `normalizeTaskSources` runs on Load *and* Save, so filling it in "like we do for `max_tasks`" would stamp the current default onto every existing source the first time any surface writes config — and the operator's later change of the default would then move nothing, silently, for every install that already had a config. It fails no obvious test; `TestInheritedProviderIsNeverMaterialized` is the guard.

## Three API facts that shaped the gist adapter

Each verified against the SDK source rather than assumed:

- **go-github v90 is the functional-options rewrite.** `NewClient` returns an error, `baseURL` is unexported, `Gists.Edit` is now `Gists.Update`. Anything written from a pre-v78 example does not compile. This is also convenient: `WithURLs` takes `*string` and `WithTimeout` exists, so the adapter imports neither `net/url` nor builds a dialer — keeping both off the egress allowlist by construction.
- **`PATCH /gists/{id}` has no compare-and-swap.** No `If-Match`, no precondition of any kind; gist ETags are GET-side caching only. Writes are last-write-wins, so hap's existing advisory flock — keyed on the canonical locator — is what serializes them. A writer that does not take that lock (the web UI, `gh gist edit`, hap on another host) can still lose an edit; that is documented, not defended against. A check-then-PATCH would narrow the window while reading like a guarantee it cannot make.
- **`GistFile` has no `Truncated` field.** GitHub truncates content above ~1 MB while still reporting the real size, so `Size != len(Content)` is the only signal — and missing it would PATCH a clipped body back over a complete one, silently deleting the tail of the operator's checklist. Guarded on read *and* before every write, with tests asserting **zero** PATCH requests when it trips.

The PATCH also names only the target file: `UpdateGistRequest.Files` maps a filename to `nil` to **delete** it, so a request naively rebuilt from a `Get` could clobber another agent's list.

## Privacy (NFR-007)

Deliberately amended, not quietly widened. Two allowlisted egress files now, both operator opt-in and both off by default: the existing release check, and the gist adapter.

The allowlist bans the **GitHub SDK by its own import path** as well as `net/http`, because the walker checks *direct* imports — an adapter naming only the SDK would have egressed while passing. Verified by mutation: dropping just the SDK entry fails the test naming `gist.go`.

The gist provider carries the task text of the sources using it and nothing else — no pane content, no learned rules, no audit history.

## Daemon

Remote reads are served from an off-loop snapshot (20 events cost one backend read; a store whose `Read` takes three seconds never delays the loop). Local reads still go straight through, which is why a `hap task add` from the CLI is still seen on the very next event.

The safety argument, which any change here must keep: the cache feeds candidate **selection** only, and every write re-reads inside the lock — so a stale snapshot can make the daemon *propose* a task that is gone, but never *commit* one. `TestStaleSnapshotCannotReserveTheWrongItem` pins it over four ways the list can move.

Stranded hand-outs are released off the loop too, funnelled back over one channel handled beside `llmResults`; all state mutation stays on the loop.

**Deliberately still inline: the delivery reserve.** It is ordered audit → reserve → send → ledger, and the send is `herdr agent send`, a subprocess that already runs inline on this loop — moving only the reserve buys nothing and would spread that ordering across loop turns, where a config reload or a kill event can interleave. The reasoning is recorded in `internal/daemon/taskwriteback.go` rather than left implicit.

## Bugs found and fixed during review

A code-review pass and a later verification pass each turned up real defects behind work that looked done:

- The **idle poll never reached a remote list** — it still read `src.Path`, so every gist-backed `enable_auto_send_task_when_idle` source logged "unreadable" once a sweep and handed out nothing. The flagship integration, silently dead.
- A remote read failure resolved to **"no task source matched"**, the branch that licenses *inventing* tasks. With generation enabled, a config reload reliably made hap generate work for an agent whose real list was full.
- All three front-end send paths passed the **locator** as `{task_list_path}` and never set `Remote`, so the agent was told its list was at `gist://…` and got the local template with its `--path` clause.
- The generated-task bootstrap **wrote locally while registering the store**, leaving the agent's real list empty. Routing it through the store also fixed a pre-existing data-loss bug: it swallowed a read error as "empty" and then overwrote, discarding every task in an unreadable-but-present list.
- The mutate deadline was smaller than the lock wait it existed to accommodate, so contention guaranteed failure; a cache refresh in flight during a write resurrected pre-write bytes.

## Release

Minor bump, hand-written per the reserved manual path: `herdr-plugin.toml` carries `0.6.0` and the changelog is assembled into a single `## 0.6.0` section in this branch, since no bump commit is created for a minor. `changelog.d/` holds only its README.

## Verification

- `go build` / `go vet -tags "vectors cpu"` and `gofmt` clean; submodule gitlink intact, no submodule churn.
- Local integration suite against **real herdr**: 10 pass, 6 skip, 0 fail — the skips are the token-spending `HAP_ITEST_CLAUDE` / `HAP_ITEST_CODEX` cases. `TestRealConsultContextTaskSourceSummary` passes, which exercises the task-source path this branch rewrote.
- Unit suite green except one `internal/daemon` test per full run — a *different* one each time across seven runs, every one passing 5–10/10 in isolation, all `waitFor` timeouts at load 4–5.4 on 4 cores. Baselined: the `TestAutoSendIdle` group is 0/11 on untouched `main` and 1/10 here. Worth one confirmation run on an idle machine.

## Not covered

No live gist round trip — that needs a real token, so the gist adapter is exercised only against `httptest`.

https://claude.ai/code/session_016t5eQ9jbYfHUNVGVNWceEn


0xGosu:
@CharlieHelps please review

0xGosu:
Thanks — the locator finding was right, and worse than it looked, because every symptom was silent. `filepath.Abs` does not fail on `gist://<id>/<file>`; it returns `<cwd>/gist:/<id>/<file>`.

Fixed in 272fab4, all three sites:

- **`TaskSourcePathFor`** ran `Abs` over the locator, so CLI `task send` reserved and delivered against a **local** file of that name instead of the store.
- **`TaskSourceTemplateFor`** compared a bare gist file name (turned into a cwd path) against a locator, never matched, and silently sent the **default** prompt rather than the operator's configured template.
- **`taskSourceLimit`** did the same, so a configured gist source read as unregistered and therefore **uncapped** — worse than having no cap, since the operator has set one and believes it applies.

All three now resolve the source the same way the locator was produced and compare canonically (`sourceLocatorMatches`), which also keeps two spellings of one local file matching. That single helper replaces both hand-rolled comparisons — and each had already broken once in this branch: the macOS failure you saw was the same mistake in its other direction (symlink-resolving one side only).

Coverage added as requested: remote send addressing, remote template lookup, and remote `max_tasks` enforcement. Mutation-verified — reinstating the `Abs` makes the first fail with exactly the mangled `<cwd>/…/gist:/3f2a1b9c/brave-otter.md` you described.

On the CI note: you reviewed `f7674a9`. The lint issues and the macOS path/cap failures were fixed in `ad19199` and `99eaf9a`, and all six checks passed on `99eaf9a`. I've since run the full gate locally — build, vet, gofmt, `golangci-lint` (0 issues), the complete `vectors cpu` suite, and `-race` over the changed packages.

One thing still uncovered, stated plainly: there is no live gist round trip anywhere in this branch. The adapter is exercised only against `httptest`, since a real run needs an operator token.

PR #329, opened 2026-08-11
feat: make every task-source field editable in place

0xGosu:
Follow-up to #326: a task source's fields could all be set when it was **created**, but four of them could never be **changed**. This closes that.

## The gap

`hap config task-source add` has always accepted every field of a `[[task_sources]]` entry. `set` accepted five of nine — probing the shipped binary:

```
set 0 auto-send-when-idle false                  -> ok
set 0 enable-llm-review-before-auto-send true    -> ok
set 0 max-tasks 40                               -> ok
set 0 provider local_fs                          -> ok
set 0 gist-id aa11bb22                           -> ok
set 0 path /tmp/other.md                         -> error: unknown task-source key "path"
set 0 agent other-agent                          -> error: unknown task-source key "agent"
set 0 workspace 'codex-*'                        -> error: unknown task-source key "workspace"
set 0 template X                                 -> error: unknown task-source key "template"
```

So re-pointing a source — the most ordinary edit there is — meant removing it and re-adding it: retyping every other field, and renumbering every later source, to change one.

The previous rationale was that those fields "silently re-point an agent's work". That doesn't survive contact with the rest of the surface: `hap config task-source remove` is unguarded by design, so withholding the smaller edit while allowing the larger one bought nothing. The three that re-point a source now **say what they changed from** and warn, which is what that concern actually called for.

## What's editable now

```sh
hap config task-source set brave-otter path /new/tasks.md
hap config task-source set brave-otter agent swift-heron
hap config task-source set brave-otter workspace 'codex-*'
hap config task-source set brave-otter template 'Do: {next_task_content}'   # "" restores the default
```

`TestEveryTaskSourceFieldIsEditable` walks `config.TaskSource` and fails the build on a field with no `set` key — and drives each key through the real dispatcher, so a map entry naming a key that doesn't exist can't pass green either.

## Addressing a source: name, not just index

`list` prints the index (`#0`, `#1`) — but the index was the weak part, not its discoverability. It is **positional**: removing a source renumbers every one after it, so a number an operator noted a minute ago can silently mean a different entry, and it says nothing about what it points at.

So `set` and `remove` now take **the agent name** as well:

```sh
hap config task-source set brave-otter path /new/tasks.md
hap config task-source remove brave-otter
```

`#0` is still accepted verbatim, so a listing row can be copied unedited. The two forms cannot collide: a herdr agent name must start with a lowercase letter, so a purely numeric token is never one. A name matching no source — or more than one — is refused, naming the indexes that disambiguate it; a workspace-scoped source has no agent to be addressed by, so it takes an index.

Two discoverability fixes came with it: `list` used to end by suggesting `list` (the command you had just run), and the empty listing was a dead end. The footer now names a real source, and the empty case points at `hap agents` and `add`. That meant making the command `SelfHints`, so `add`/`set`/`remove`/`provider` each print their own footer, all leading with `list` — because a removal renumbers the rest.

## Behaviour worth a reviewer's eye

- **A relative `path` is expanded and absolutized in the operator's process**, exactly as `add` does, because the daemon runs from the state dir and would otherwise resolve a stored relative path somewhere else. A remote (gist) provider is left alone: its value is a locator, and `filepath.Abs` doesn't reject one, it silently mangles it.
- **Emptiness is left to `config.ValidateTaskSource`**, not pre-empted. An empty path is illegal under a local provider (it already says so, naming the provider) but *legal* under a remote one, where it means "one list per agent".
- **A whitespace-only template normalizes to empty.** Only the empty string falls back to the built-in default, so `"  "` stored verbatim would be rendered as the entire outbound prompt — a hand-out with no task text and no instructions.
- **An empty `agent`/`workspace` matches any**, which is the widest re-point available, so it is called out rather than applied silently.
- **The confirmation echoes the STORED value**, not the typed one — otherwise following the help's own example prints a relative path apparently replacing an absolute one.
- **`set` keeps strict 3-argument arity** (it does not join trailing words like `config set` does). A template with spaces needs quoting, which the examples show; joining would let `set brave-otter agent a b` silently store a nonsense selector.
- **`template` carries no re-point warning** — it changes the prompt's format, not its target.
- **A purely numeric agent selector is refused** at both write surfaces (`add --agent`, `set … agent`). It is what makes the index/name discrimination true by construction rather than by assertion: stored, `--agent 3` would leave that source unaddressable by name while `set 3 …` silently meant index 3. A real pane or agent id (`1-1`, `w1:p1`) is unaffected — it does not parse as an integer.
- **The `list` footer only names an agent when it resolves uniquely.** Two sources feeding one agent is legal, so naming that agent would print a command the resolver refuses — on the very screen meant to resolve the ambiguity.

## Verification

- Full tagged suite green on the final commit — all 34 packages including `internal/daemon`. `go vet`, `golangci-lint` and the integration suite clean. (Earlier runs saw three *different* `internal/daemon` tests flake under load; each passed in isolation, this branch touches no daemon file, and that package is documented as broadly flaky under concurrent load.)
- Four behaviours mutation-verified: the parity test, the strengthened key check, the whitespace-template normalization, and the field-editability coverage — each fix reverted and the test confirmed to fail.
- Every key smoke-tested against a real binary in an isolated config/state dir.
- Reviewed by code-review subagents at each stage. The first found the whitespace-template and empty-path bugs above plus four smaller issues; the second found that the double-footer sweep could not see a MISSING footer (the only failure mode `SelfHints` creates), that the `list` footer could suggest a refused command, and that the numeric-selector invariant was asserted but unenforced. All addressed, with tests.
- Every new behaviour mutation-verified: the fix was reverted and the test confirmed to fail.

`README.md`, the `hap` skill, `CLAUDE.md` and a changelog fragment are updated.

https://claude.ai/code/session_016t5eQ9jbYfHUNVGVNWceEn


0xGosu:
@CharlieHelps please review

PR #331, opened 2026-08-11
#330 fix: TUI Tasks tab works with gist task sources; task-source index becomes hap task's provider-independent selector

### LINKED ISSUE #330: TUI Tasks tab cannot display or act on gist-backed task sources
The Tasks tab renders a derived (one-list-per-matched-agent) gist source as `✗ one list per matched agent — open an agent's list from the Agents tab` even when the source names exactly one concrete agent (e.g. `agent=honest-ibex`), whose list is unambiguous.

Additionally, Tasks-tab item actions key groups by `Source.Path`, which under a remote provider is only a file name (or empty for a derived source) — so done/edit/delete/move/send address the wrong identity for every non-local_fs source (a bare file name canonicalizes against the TUI's cwd, not the gist locator).

Found while live-testing the github_gist task-source provider (PR #325) after migrating an agent's source to the gist default.

0xGosu:
Fixes #330.

Two related changes: the TUI Tasks tab display/actions fix for non-`local_fs` providers, and a follow-up refactor making the task CLI provider-independent via a task-source index selector.

## Part 1 — TUI Tasks tab vs gist sources

**What was broken.** A derived (one-list-per-matched-agent) gist source always rendered `✗ one list per matched agent…` — even with exactly one matching agent (`agent=honest-ibex`), whose list is unambiguous. Underneath: Tasks-tab item actions keyed groups by `Source.Path`, which under a remote provider is a bare file name (or empty) that canonicalizes against the TUI's cwd — done/edit/delete/move/send addressed a list the store never serves.

**The fix.** `TaskGroups` receives the refresh's status snapshot and resolves a derived source when exactly ONE live agent matches (zero/several keep the note — guessing would show another agent's list). Rows and every group-identity site address lists by the resolved locator (`TaskGroup.ListAddress`). One shared matcher (`frontend.SourceMatchesAgent`) keeps the header's `→ name` annotation and the displayed list in agreement. Group count/order never depend on live agents (the TUI addresses groups by config index). Displays never echo a full gist id (`gist:<8-chars>…/<file>`, the `shortGistIDForDisplay` rule).

## Part 2 — the task-source INDEX as `hap task`'s fallback selector

**Why.** The CLI's fallback for a name-unresolvable source was `--path <file>` — in resolver errors, in the default next-task prompts, and in the `list` hints. `--path` always reads a LOCAL file, so under a remote provider all of that advice was dead. The task-source index addresses the source through hap's own config, so it works identically under every storage provider.

**What changed.**
- `hap task <N> …` (also `'#N'`) addresses source N from `hap config task-source list`, wherever an agent name is accepted. Bare digits can never be an agent name (herdr names start with a lowercase letter). `--path` is unchanged — still the escape hatch for ad-hoc local files.
- Both default next-task templates and the `hap task … list` hints now offer the index as the fallback; the remote default GAINS a fallback (a shared remote list scoped by workspace/type is exactly the source a name cannot address, and was CLI-unreachable before).
- New template placeholder `{task_source_index}`, threaded from the config walk that matched the source (daemon match, CLI resolver, TUI group index, generated-task flows) — never recovered by comparing entries, since duplicate sources are legal. A sender that cannot know the position leaves it empty and the placeholder renders the agent name — the always-working selector, so a prompt never carries a broken command.
- Fail-safe refusals: an index names a SOURCE, so a derived per-agent source addressed by index refuses toward the agent name, and `task <N> send` refuses toward a live agent name.

**Accepted trade-off (stated deliberately):** an index is positional, not identity — `task-source remove` renumbers later sources, so a prompt rendered before a config edit can name an index that now points elsewhere. That is why the agent NAME stays the primary selector everywhere and the index is only the fallback for sources a name cannot address.

## Deliberately unchanged

- `hap config task-source list` (CLI) still renders the per-agent template note for derived sources (documented as intentional in `tasklocator`).
- A resolved derived list that does not exist yet shows the truthful `file does not exist` read error (created on first hand-out).

## Known follow-up (not this PR)

A resolved gist source is read once per TUI refresh tick with no cache — pre-existing behavior for explicit-path gist sources since #325; a short-TTL cache in `frontend.readList` would cover both.

## Tests

TUI/frontend: sole-match-resolves; 4-case template-stays table (ambiguous / no match / AgentsKnown=false / unnamed match); rows+targets carry the locator, never echo the full gist id; per-provider add-prompt refusals; send staleness against a re-resolved derived source. Index selector: local + gist sources by index (`N` and `'#N'`), mutation through the index, out-of-range and derived-by-index refusals with discovery hints, `send`-by-index refusal, `{task_source_index}` rendering + empty-fallback, both default templates pinned to the index fallback with `--path` banned from both, index-addressed hints drop the fallback note. Full suite green (`-tags "vectors cpu"`).


0xGosu:
@CharlieHelps please review

PR #336, opened 2026-08-12
#335 fix: gist task sources — create a list on demand, settle reads after a write, honest config/TUI reporting

### LINKED ISSUE #335: gist task sources: list cannot be created from hap, config writes fail without a daemon, TUI hides lists of offline agents
Found by live-testing the github_gist task-source provider end to end (real gist, CLI + TUI).

1. **A gist-backed list cannot be seeded from hap at all.** `hap task <agent> add` on a configured source whose list does not exist yet fails with `file does not exist`; the TUI's add refuses with "use `hap task <agent> add` instead"; and the remote task-management hints say "there is no file here to open or edit, so `hap task` is the only way to read or change it". The list is only ever created by the daemon's first hand-out — which cannot happen until the list holds a task. Net effect: a fresh derived (one-list-per-agent) gist source is a dead end unless the operator edits the gist on github.com.

2. **Every `hap config …` write reports failure when the daemon is not running.** It exits 1 printing only `daemon nudge failed (daemon not running?)` and suppresses the success output, although the config was saved. Configure-then-start is the ordinary first-run order, so the whole gist setup flow looks broken, and re-running a command double-adds sources.

3. **The TUI Tasks tab hides a task list whose agent is not currently live.** A derived source scoped to an agent NAME shows `✗ one list per matched agent` when that agent's pane is closed, while `hap task <name> list` prints the list fine — same config, same list, two answers. (#330 only covered the case where exactly one live agent matches.)

4. **The TUI Config tab shows no storage location for gist sources.** The row renders the raw `Source.Path`: a bare file name for an explicit gist file, and EMPTY for the derived shape — a blank column where the operator looks to confirm their provider took effect.

0xGosu:
Fixes #335. Found by live-testing the `github_gist` task-source provider end to end against a real gist (CLI + TUI), on latest main.

## The dead end that started it

A gist-backed task source could not be seeded from hap at all:
- `hap task <agent> add` → `file does not exist`
- the TUI's add → `use \`hap task <agent> add\` instead` (the command that fails)
- the remote hints → "there is no file here to open or edit, so \`hap task\` is the only way"

The list was only ever created by the daemon's first hand-out, which cannot happen until the list holds a task. For a derived (one-list-per-agent) source there is no file to create by hand, so the only way out was editing the gist on github.com.

`add` is the op that creates content, so it now creates a **configured** source's list on demand. `--path` still fails loudly on a typo (`ports.EnsureCreator`'s rule), and a DERIVED source additionally requires evidence its selector names an **agent** — a selector can equally be an agent TYPE, and `<type>.md` is a list no hand-out ever reads, so creating one would queue work that never arrives. A refusal says which case it is.

## A real GitHub behavior this exposed

**Gist reads are not read-after-write consistent.** Measured live: PATCH-create a file, then GET the gist immediately — the new file was **missing in 1 of 3 rounds**. "Absent" is load-bearing in this adapter (`Read` turns it into `fs.ErrNotExist`, `Ensure` turns it into a create), so every create-then-use flow failed intermittently — the new CLI add, and equally **the daemon's first hand-out**.

The store now re-reads briefly (250ms/500ms/1s, 15s window) but **only for a file this process wrote**, so a typo'd name still fails on the first answer. And `Ensure` refuses rather than PATCHing when a file it wrote still reads as absent: the generated-task bootstrap ensures with EMPTY content, so a clobber there would erase a populated list.

## The rest

- **Every `hap config …` write reported failure when the daemon was not running** — exit 1, success output suppressed, though the config had been saved. Configure-then-start is the ordinary first-run order. The nudge only asks a *running* daemon to re-read what is already persisted, so it can never be the write's error; it now reports delivery for WORDING only, and `config set` says which state you are in instead of always claiming "(daemon reloaded)".
- **The TUI Tasks tab hid a list whose agent was not currently live** — a source scoped to a known agent's name now shows its list either way, matching what `hap task <name> list` always printed. (#330 only covered a live sole match.)
- **The TUI Config tab showed no storage location for gist sources** — a derived source rendered a blank column that read as unconfigured. Each row now names where its list lives.
- **`max_tasks` was ignored for a task added from the TUI**: the cap lookup and the create path now resolve the same source, so a limit applies however the list is addressed.

## Verified live against a real gist

Full CRUD (list/add/get/start/done/undone/update/move/remove) by agent name and by source index; remote content byte-compared after each mutation; provider `set … local_fs`/`inherit` round-trip with the never-materialize invariant holding; `max_tasks` enforcement; error paths (missing env file, bad gist id); the TUI's Tasks tab, item actions writing through to the gist, and the Config tab. Create-then-add passed 4/4 live rounds after the settle fix (it failed on the first attempt before it).

## The daemon hand-out, verified live

Run against a real daemon built from this branch, with a scratch `claude` agent in its own workspace bound to a gist source (`--auto-send-when-idle`):

1. `hap task <agent> add` **created** `rosy-orca.md` in the gist — the same command failed outright on the previous build.
2. The daemon read the gist-backed list through its remote snapshot, reserved the item `[-]` **in the gist**, and delivered the prompt unattended (audit row `auto … enable_auto_send_task_when_idle`).
3. The agent received it and replied — the rendered prompt carried the remote template including the `{task_source_index}` fallback (`use the task-source index \`1\``).
4. The reservation survived past the reclaim grace, i.e. the `working` transition confirmed it and `reclaimStrandedTasks` correctly left it alone.

Worth noting for anyone reproducing: a newly configured remote source is not delivered on the very next tick — the daemon serves remote lists from a snapshot refreshed on `refresh_seconds` (default 30s), so the first hand-out lands within about a minute.

## Not covered

The LLM-generated-task bootstrap (`task_generate_command` → operator confirm → list created → send) was not exercised; it spends model tokens and is a separate flow from the declared-task hand-out tested above.

## Tests

Create-on-demand (local, TUI-shaped locator call, derived+known agent, derived+unknown selector refusal, `--path` typo still refuses); cap parity for a locator-shaped add; config write with no daemon; reload wording; derived resolution by named selector, type selector with nothing live, several live matches, unread registry; gist settle after our own write, no wait for a file we never wrote, `Ensure` refusing to clobber, and context cancellation mid-settle.


0xGosu:
@CharlieHelps review

PR #339, opened 2026-08-12
#338 fix: create a gist task list with its header, never blank

### LINKED ISSUE #338: Accepting an LLM-generated task escalation always fails with 422 under a github_gist task source
## What happens

Accepting an LLM-generated task escalation in the TUI fails whenever the task-source provider is `github_gist`, and no task list is created:

```
✗ write tasks file: write "proud-robin.md" to gist 4300688e…: PATCH https://api.github.com/gists/4300688e…: 422 Validation Failed [{Resource:Gist
```

This affects every operator on a `github_gist` provider, on every generated-task confirm — the feature cannot work at all.

## Root cause

`frontend.ensureGeneratedTaskFile` creates the agent's task list with **blank** initial content and writes the real tasks in the `Mutate` that follows:

```go
if _, err := a.ensureList(ctx, cfg, locator, ""); err != nil {
```

**GitHub cannot store a blank gist file.** Verified live against api.github.com (2026-08-12) — `""`, `"\n"` and `" "` are all rejected identically:

```
422 {"message":"Validation Failed",
     "errors":[{"resource":"Gist","code":"missing_field","field":"files"}]}
```

GitHub reads a blank body as "this entry carries no file", drops the entry, and then rejects the request for having an empty `files` map. That is why the error names `files` — never the filename, never `content` — and reads as a malformed request rather than the one thing actually wrong. Nothing is created.

So a gist cannot represent an empty list at all, and **create-blank-then-fill is not a valid pattern**, even though it is invisible on a local file.

## Why the unit suite did not catch it

The fake gist server in `gist_test.go` accepts blank content. Every unit test passed while the live call could never succeed.

## Fix

- Seed the create with `newListHeader`, the header the other two create-on-demand callers already used.
- Enforce never-blank at three layers so the class cannot recur: `frontend.ensureList` (every backend — a local file accepts a blank seed, so guarding only where it breaks leaves the next caller green in CI and broken for the first gist operator), the `ports.EnsureCreator` contract, and `gist.Store.put` with a new `ErrBlankContent` on every write path. Only CREATE is closed — emptying a local list by removing its last item stays legal.
- Make `fakeGist` answer a blank write with GitHub's real 422.

0xGosu:
Fixes #338.

## The bug

Accepting an LLM-generated task escalation always failed under a `github_gist` task source, and no task list was created:

```
✗ write tasks file: write "proud-robin.md" to gist 4300688e…: PATCH https://api.github.com/gists/4300688e…: 422 Validation Failed [{Resource:Gist
```

`ensureGeneratedTaskFile` created the agent's list with **blank** initial content and wrote the real tasks in the `Mutate` that followed. That is invisible on a local file and fatal on a gist: **GitHub cannot store a blank gist file at all.**

Verified live against api.github.com (2026-08-12) — `""`, `"\n"` and `" "` are refused identically:

```
422 {"message":"Validation Failed",
     "errors":[{"resource":"Gist","code":"missing_field","field":"files"}]}
```

GitHub reads a blank body as "this entry carries no file", drops the entry, then rejects the request for an empty `files` map. That is why the error names `files` rather than the list or the content, and reads as a malformed request. Nothing is created.

## The fix

The list is seeded with `newListHeader` — the header the other two create-on-demand callers already used; all three now share the helper.

The never-blank rule is enforced at **three layers**, deliberately:

| Layer | What it closes |
|---|---|
| `frontend.ensureList` | Every backend. A local file accepts a blank seed, so guarding only where it breaks leaves the next caller green through the entire unit suite and broken for the first `github_gist` operator — which is exactly how this shipped. |
| `ports.EnsureCreator` | States it as the interface's contract, which previously promised nothing about `initial`. |
| `gist.Store.put` | Refuses with a new `ErrBlankContent` on **every** write path, so a mutation that empties a list reports what is wrong instead of the 422. Spends no request. |

Only **create** is closed. A mutation may still empty a *local* list — that is what removing the last item from a headerless one does.

## Why CI could not have caught this

The fake gist server accepted blank content, so every unit test passed while the live call could never succeed. `fakeGist` now answers a blank write with GitHub's real 422, modeling the drop-then-reject behavior faithfully.

## Tests

- `TestGistRefusesToWriteBlankContent` — create and empty-a-list paths, asserts the refusal names the file (GitHub's does not) and spends no PATCH.
- `TestGistFakeRejectsBlankContentLikeGitHub` — pins the fake against the real API so the two cannot drift.
- `TestBootstrapSeedsTheListWithItsHeaderNotBlank` — asserts the **seed**, not the finished list (the finished list is non-blank either way, which is why this shipped), and pins the precondition that makes that assertion meaningful.
- `TestEnsureListRefusesABlankSeedOnEveryBackend` — asserted under the *local* provider, which would otherwise accept one.
- `TestNewListHeaderIsNeverBlank`.

## Verification

- `go test -tags "vectors cpu" ./... -count=1` — all packages pass.
- `golangci-lint run --build-tags "vectors,cpu"` — 0 issues. `gofmt`/`go vet` clean.
- `go test -tags integration ./test/integration/ -v` — 9 pass, 7 skip (all real-agent cases gated on `HAP_ITEST_CLAUDE`/`HAP_ITEST_CODEX`).
- Live probes against a real gist confirmed the 422 and that no file is created by a rejected write.

Not yet done: end-to-end confirmation of a real pending escalation against the live gist. That needs the operator's go, and is tracked separately.

https://claude.ai/code/session_01U1oHuzzJ6r2keXRWYVHtEy


0xGosu:
@CharlieHelps please review

PR #345, opened 2026-08-12
#342 fix: handle a task-list locator as a locator on every confirm path

### LINKED ISSUE #342: Generated-task confirm with a send fails on a gist source: the reservation reads the task list as a local file
Follow-up to #338. With that fix in, the list is now created — but confirming a generated-task escalation **with a send** still fails:

```
✗ task source created, but reserving task #1 (nothing was sent):
  stat gist://4300688e…/proud-robin.md: no such file or directory
```

## Root cause — not a consistency lag

A task-list **locator is not a filesystem path**. Under a remote provider it is a `gist://<id>/<file>` URI, and `os.Stat` can only ever fail on it.

`frontend.go` reserved the task at send time with `mutateTaskFile` — a thin alias for `taskfile.Mutate`, which does `os.Stat` → read → atomic write on a local path — while the surrounding read and write had already been moved to the store:

```go
if _, err := mutateTaskFile(path, reserveTask(pos, itemText)); err != nil {
```

The rollback on a failed send had the same bug.

Two things rule out eventual consistency:

- A genuine lag surfaces as the gist store's own message (`task list "proud-robin.md" in gist 4300688e…: file does not exist`), never `stat gist://…`.
- The post-write settle retry from #335 already covers that lag and would have applied, since hap had just written the file.

## Impact

The failure lands **after** the list is created, the source registered, and the escalation CLAIMED — so the escalation is consumed, the tasks sit in the list, and nothing is ever delivered. Confirmed against a real gist: the file exists with all its tasks, and only the send is missing.

## Second instance of the same class

`pickAppendTarget` chooses which of an agent's matched sources receives the tasks — first with a pending `[ ]` item, else first with any items, else first in config order. It read each candidate with `os.ReadFile(filepath.Abs(src.Path))`, which cannot resolve a gist source's bare file name. Every read failed, the loop treats a failed read as "no content", so precedence silently collapsed to **always the first source**: wrong list, no error, no log.

## Fix

- Both the reservation and its rollback go through the store (`mutateList`); the `mutateTaskFile` alias is deleted so nothing can reach for it again.
- `pickAppendTarget` reads each candidate through that source's own store.
- An AST architecture test bans `taskfile.Mutate`/`MutateWithin`/`WriteFileAtomic` outside `internal/taskstore/local`, where a locator genuinely is a path — closing the class rather than the instance, since any future local twin of a store method breaks identically and passes every (local) test.

0xGosu:
Fixes #342. Follow-up to #339 (which got the list created; this gets the task delivered).

## The bug

Confirming a generated-task escalation **with a send** failed under a `github_gist` task source:

```
✗ task source created, but reserving task #1 (nothing was sent):
  stat gist://4300688e…/proud-robin.md: no such file or directory
```

It fails **after** the list is created, the source registered and the escalation CLAIMED — so the escalation is consumed, the tasks sit in the list, and nothing is ever delivered. Confirmed against a real gist: the file exists with all its tasks, only the send is missing.

Not the read-after-write lag. That surfaces as the gist store's own `task list "…" in gist …: file does not exist`, and #335's settle retry already covers it. `stat gist://…` is `os.Stat` on a URI.

## Root cause: a locator is not a path

Under a remote provider a task-list locator is a `gist://<id>/<file>` URI. **Three** call sites in the same confirm still treated it as a filesystem path:

| Site | What broke |
|---|---|
| send-time **reservation** + rollback | `mutateTaskFile` (a `taskfile.Mutate` alias) while the read/write around it had already moved to the store → the reported failure |
| `pickAppendTarget` | read each candidate with `os.ReadFile(src.Path)`, which cannot resolve a gist source's bare file name. Every read failed; a failed read reads as *"no content"*, collapsing precedence from *"the source with pending work"* to *"always the first source"* — wrong list, no error, nothing logged |
| bootstrap **exclusion** | compared that same absolutized path against the local bootstrap path, so an agent's own list read as external and a second confirm took `appendGeneratedTasks` — whose dedupe keys on raw text while the bootstrap flow stores NUMBERED items, appending every listed task again, un-numbered |

## The fix

All three go through the store. The `mutateTaskFile` alias is **deleted** so nothing can reach for it again.

Two subtleties worth reviewing closely, both found by review rather than by me:

- **`isBootstrapList` needs locator equality AND a declared path.** A *derived* source (`path` unset — the documented one-list-per-agent form) resolves to `DerivedFileName(agent)`, byte-identical to what the bootstrap registers. Equality alone excludes the most idiomatic gist setup, runs the bootstrap flow, and then fails **every retry** on `addTaskSourceIfAbsent`'s refusal to register a second source — after tasks were already written. The declared spelling is never compared alone either: the same file name in another gist is another list.
- **The bootstrap locator's resolution error is deferred** past the append early-return. The append path never needed that resolution, so failing early broke a confirm a healthy `local_fs` source could still serve.

The reservation rollback takes `context.WithoutCancel`: a remote store uses the caller's ctx, and the likeliest cause of the failed send it compensates for — the operator quitting the TUI, Ctrl-C in the CLI — is the same cancellation that would abort the release and strand the item at `[-]`.

## Why every one of these shipped green

**Every test in `internal/frontend` ran on a local file, where a locator IS a path.** That is the whole reason. `App.TaskStoreFor` is a test-only seam at the one lookup both resolvers share, so a REMOTE backend stands up in-process; the fake refuses a non-gist locator the way the real one does, and honors `ctx`, which the local backend does not.

`TestOnlyTheLocalBackendTouchesATaskListAsAFile` closes the mutation half by construction — an AST scan banning `taskfile`'s path-taking mutators outside `internal/taskstore/local`. Matched on the *selector*, so `f := taskfile.Mutate` is caught too, and a dot-import is refused outright since it would make the ban unenforceable. Exemptions carry a reason and are checked live, matching `boundary_test.go`.

## Every fix is mutation-tested

Each was reverted individually to confirm its test fails:

| Reverted | Result |
|---|---|
| reservation → `taskfile.Mutate` | reproduces the reported error exactly, + AST guard names the line |
| rollback → live ctx | *"the rollback itself failed, stranding the item"*, list left `[-]` |
| `pickAppendTarget` → `os.ReadFile` | task landed in `empty.md` instead of `has-pending.md` |
| exclusion → path compare | `- [ ] 1. Alpha task` **and** `- [ ] Alpha task` |
| `isBootstrapList` without declared path | *"refusing to register a second"* |
| bootstrap error fatal early | append broken by an unrelated bad default |

## Verification

- `go test -tags "vectors cpu" ./... -count=1` — all packages pass.
- `golangci-lint run --build-tags "vectors,cpu"` — 0 issues. `gofmt`/`go vet` clean.
- Changed packages at `-count=3` — no flakiness.
- `go test -tags integration ./test/integration/` — 9 pass, 7 skip (real-agent cases behind `HAP_ITEST_CLAUDE`/`HAP_ITEST_CODEX`).

Not yet done: end-to-end confirmation of a real pending escalation against the live gist. That needs the operator's go and is tracked separately.

https://claude.ai/code/session_01U1oHuzzJ6r2keXRWYVHtEy


0xGosu:
@CharlieHelps please review

PR #350, opened 2026-08-14
#1 feat: bundle the hap skill doc — print via --skill, install via TUI/CLI

0xGosu:
## Summary
- **`hap --skill` / `hap skill` / `hap skill show`** prints the hap agent skill document (`.claude/skills/hap/SKILL.md`) to stdout. The file ships **inside the binary** via the repo's first `go:embed` (new root package `skilldoc` — the only package that can reach `.claude/` without duplication, since embed paths cannot contain `..`).
- **`hap skill install <claude|codex|agents>...`** writes the bundled SKILL.md to `~/.claude/skills/hap/`, `~/.codex/skills/hap/`, or `~/.agents/skills/hap/` (atomic temp+rename; validates targets before touching the filesystem; a mid-list failure still reports what did install).
- **TUI Config-tab quick shortcut** "Install hap agent skill" opens a new **multi-select picker mode** (space toggles `[x]`, enter submits the checked set, highlighted-row fallback when none checked — same convention as the list-tab multi-selects). Reports through the usual async `actionResultMsg` → status note; the existing install-hap symlink confirm flow is untouched.
- **Idempotent pause/resume**: `hap pause`/`hap resume` and the TUI `p`/`r` keys now detect a no-op (pausing while paused, resuming while running) — they report "already paused/resumed" and record **no kill-history event**, so repeated presses no longer flood the history. The TUI's self-caused-pause bell suppression survives a rapid double-press (`pausePending` is only cleared on a no-op the TUI had already observed as paused), and `pauseCmd` gained the inflight-WaitGroup accounting so a `p` just before quit still lands.
- `hap skill --help`, `hap --skill --help`, and `hap help skill` all render the registry guide (`help()` now resolves a dashed registered alias instead of falling back to the overview).
- Deliberately not using `taskfile.WriteFileAtomic` (repo guard test bans new callers); the installer carries its own local temp+rename.

## Testing
- New tests: `skilldoc_test.go` (embed presence, all-target install, overwrite, fail-before-write, partial-write reporting, validate-before-home), `cmd/hap` `TestRunSkill` (print/`show`/typo/no-target/real install under fake `$HOME`), `internal/tui/skillinstall_test.go` (row placement, toggle render, async submit, fallback, esc, error surfacing), `TestWantsCommandHelp` alias cases, `TestPauseResumeNoOpsRecordNoKillEvent`, `TestBellPausePendingClearedOnNoOpPauseAction`, `TestBellNoOpPauseBeforeObservedTransitionKeepsSuppression`.
- Full suite `go test -tags "vectors cpu" ./... -count=1`: all touched packages green across five runs; only failures are the two documented pre-existing flakes (`internal/daemon` shutdown-drain timing, `internal/embedder` stall-guard under load) in untouched packages.
- `gofmt`, `go vet`, `golangci-lint run --build-tags "vectors,cpu"`: clean.
- Manual smoke: `hap --skill | wc -c` = 77320 (byte-identical to the repo file); `hap skill show` identical; install under a fake `$HOME` creates all three layouts; typo/no-target refused with exit 1.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -14,6 +14,11 @@ Prefer these for how-to detail — this file keeps only what must stay in view.
 - **`hap-development-local`** — the local dev loop: link the working tree, rebuild,
   hot-swap the daemon (`hap daemon --ensure`), and live-test against a real agent.
 
+The hap skill also ships inside the binary: `hap --skill` prints it, and
+`hap skill install <claude|codex|agents>...` (or the TUI Config-tab shortcut)
+installs it into `~/.claude/skills/hap/`, `~/.codex/skills/hap/`, or
+`~/.agents/skills/hap/`.
+
 ## Build, test, lint
 
 The semantic matcher links native code (llama.cpp via CGO, FAISS behind bleve's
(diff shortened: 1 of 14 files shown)


0xGosu:
@CharlieHelps please review

PR #361, opened 2026-08-17
feat: bound full self-prompting by [limits], let it accept generated tasks, and attribute its audit rows

0xGosu:
Three opt-in changes to full self-prompting (FSP). **All default off**, so an existing install behaves exactly as before.

## 1. `full_self_prompting.honour_limits`

FSP never checked the `[limits]` ceilings before acting — it advanced the counters and let the *next* decision notice, which always overshoots and only pauses that one agent. With this key set, the ceilings are checked **before** each delivery, and reaching one switches the whole mode **off**: an in-memory latch stands it down immediately, then a tracked goroutine writes `enabled = false`, records the `fsp_off` row, and notifies the operator naming which ceiling tripped.

Three bounds are load-bearing:

- **A pause is not a ceiling.** `CheckRate`'s first branch answers `rate_limited` for a paused agent, and Guard 1b already handles that correctly one row later. Reading it as a ceiling would stand the mode down for the whole herd over a per-agent state.
- **The ceiling is only read on a row that could actually be delivered.** `ConsecutiveAuto` is reset only by human interaction, so a killed agent carries it forever; its leftover escalation would stand the mode down, re-trip after every re-enable (the latch clears on reload), and never advance the absence bookkeeping that would retire it — a livelock with no end.
- **The stand-down `continue`s rather than `break`s**, so remaining candidates still register in `stillEligible` and keep their delivery budgets and absence counts.

Worth knowing before enabling: the ceilings are **per agent** while the mode is **global**, so one runaway agent stands the mode down for everyone. The per-minute window is also shared with every other autonomous send on that agent, so on a busy herd it is the one likely to fire first. Both are documented in the README.

## 2. `full_self_prompting.accept_generated_task`

FSP may now act on an idle escalation whose suggestion is an LLM-**generated** task — writing the task list, registering the source, handing the first task over — instead of leaving it for the operator.

- Reached through an optional seam on `daemon.Options`, implemented by `frontend.App` and wired in `cmd/hap`, so `internal/daemon` still does not import `internal/frontend`.
- The daemon keeps sole ownership of the audit row. An `automated` flag makes `acceptGeneratedTask` skip `ResolveEscalation` **and** `InsertCorrection`, preserving the invariant that an automatic acceptance is never a learning event.
- **New safety screen.** The task text is authored by the generator LLM *after* the decision that raised the escalation, so nothing had ever vetted it — the operator's confirm was the gate, and this feature removes it. Never-auto and suspected-irreversible now run in the fork, on the **rendered** prompt rather than the stored text (an encoded `\n` defeats a line-anchored rule). A hit leaves the escalation for a human.
- The seam call runs inside `WithAgentAutomation`, like every other delivery, so a mid-flight disable still stops it.

Known limitation, documented: idle situations compare by unstructured pane text, so Guard 3 usually answers "unevaluable" and the row waits for the operator. This fires less often than the key suggests.

## 3. `while_fsp_mode_on` audit column

`auto_accepted` is what timed auto-accept writes too, so an FSP answer and a threshold quietly expiring were indistinguishable in the log. The new column is written at finalize in the same guarded `UPDATE` as the status, and carried through the finalize retry (`autoAcceptNeedsFinalize` is now id → was-FSP; a bare id set would drop it on exactly the rows whose bookkeeping already failed). Such rows read `fsp-sent` in `hap audit` and the TUI, render amber on the Audit tab, and name the cause in the detail view. Existing rows are not backfilled.

## Testing

- 33 new tests across `config`, `domain`, `store`, `daemon`, `frontend`, `tui`.
- Full suite green: `go test -tags "vectors cpu" ./... -count=1`. `gofmt`, `go vet` and `golangci-lint` clean.
- Local integration suite: 9 ran, 7 skipped (need `HAP_ITEST_CLAUDE=1`).
- Several guards proved by **mutation** — the safety screen, the claimed-row status guard, and the lifecycle barrier each fail their test when the fix is removed.

A code review caught one bug worth calling out: `acceptGeneratedTask`'s early-out compared the row's status against `"escalated"`, but the automated caller arrives with it already claimed as `auto_accepting`. Every acceptance errored, `autoAcceptDeliveryFailed` burned an attempt per sweep, and the escalation was **dismissed** once the budget was spent — the feature deleted the suggestions it existed to act on. It passed because the daemon tests wire a fake seam; the fix is pinned by `frontend` tests driving the real one.

## Not yet done

The live verification steps (tripping a ceiling on a real agent, the config rewrite and kill-history row, an end-to-end generated-task accept, the amber rows in a running TUI) have **not** been run — they need a live herdr, real agents and token spend.


0xGosu:
@CharlieHelps please review

0xGosu:
Thanks — I worked through all four. **Three fixed, one fixed with a note on provenance.** Each fix is pinned by a test that fails without it (verified by mutation).

---

### 1. Screening the wrong prompt — **valid, fixed**

You're right, and it's the one that mattered most. The daemon's pre-check can only render with the *default* template, because the target source (and so its `next_task_template`, resolved path and index) is chosen inside the seam. A custom template can frame a benign task into something the rules refuse.

The seam now takes a `screen func(string) error` — the same shape `tasklistreview`'s `safe` closure already uses — and calls it with the exact bytes immediately before the send, and **before the reservation**, so a refusal strands nothing. The daemon-side pre-check stays: it refuses obvious cases before any list is written or source registered.

Coverage needed both halves, which is worth spelling out: the daemon tests drive a *fake* seam, so they only prove the callback refuses. `TestAutomatedGeneratedTaskScreensTheSourceTemplatePrompt` (frontend) proves the real path calls it, with a custom `next_task_template`, and asserts nothing was reserved or sent. Plus `TestFSPGeneratedTaskScreensTheExactOutboundPrompt` and a control (`…AcceptsASafeRenderedPrompt`) so the screen can't pass by refusing everything.

### 2. No crash recovery for the reservation — **valid, fixed**

Confirmed the mechanism: `[-]` sets `Done` (`tasklist.go:689`), so after a crash the retry hits the already-claimed branch, burns its attempt budget, and the escalation is dismissed. The task is stranded.

The automated path now writes the same `task_reservations` row the daemon's own hand-outs use, so `reclaimStrandedTasks` returns the item to `[ ]` once its agent is parked past the grace window; a rolled-back send retires the row again. `RecordTaskReservation`/`DeleteTaskReservation` are added to `ports.FrontendStore` for this.

One asymmetry, documented in the code: these rows carry no `TerminalID` (the front end holds no transition). Both readers treat empty as "matches any" — the confirm SQL and reclaim's `sameTenant` — so the row confirms and ages normally. What it gives up is telling a *recycled* pane apart: a busy successor would pin the item until `staleHandoutTTL` retires the row with a warning. Degraded, never a wrong send.

The **operator** path deliberately records nothing (`TestOperatorConfirmRecordsNoReservation`): a human is present to read the error, and ledger rows there would start barring manually-confirmed agents from the idle poll.

Tests: `TestAutomatedGeneratedTaskRecordsAReservation`, `…ReleasesTheReservationOnAFailedSend`, `TestOperatorConfirmRecordsNoReservation`.

### 3. FSP check not serialized with the claim — **valid, fixed**

Added a `stillPermitted` closure re-asked immediately before `ClaimForAutoAccept`. `TestFSPRechecksTheModeBeforeClaiming`.

One deliberate narrowing: it's the *cheap* check (config snapshot + ceiling latch), not `fspActive`. `fspActive` queries the store for the graduated-rule count, and this runs once per candidate row on the select loop. The runtime preconditions it checks are evaluated once when the sweep begins and don't need re-polling per row; what changes under us mid-sweep is an operator action, and both of its forms are already in memory here.

### 4. `context.Background()` on the forward mutation — **valid, fixed, but it predates this PR**

Worth flagging so it's clear I evaluated rather than rubber-stamped: that line is from commit `ce6de0a` (#338/#339), not this PR — both your comments anchored to the `appendGeneratedTasks` signature line, which is the one line I touched in that function.

I fixed it anyway, because this PR is what first gives it a *cancellable* ctx: under an automated acceptance the daemon's run ctx reaches it, so a shutdown could still create a remote list. The create is idempotent and the escalation stays pending, so aborting loses nothing.

---

Full suite green, `gofmt`/`go vet`/`golangci-lint` clean. Live verification (a real agent tripping a ceiling, an end-to-end generated-task accept) is still outstanding and noted in the PR description.

Issue #356, opened 2026-08-16
reconcile: an unrelated pending escalation hides a parked agent, and episodeHandled latches before the check


0xGosu:
## Symptom

After the daemon restarted at 04:33 (2026-08-16), agent `witty-ocelot` (`wF:p30`) sat
visibly `blocked` with a form on screen and hap never looked at it again. The log shows the
reconcile re-driving only the *other* agent:

```
04:33:00 reconcile: re-driving parked agent agent=wF:p4Z pane=wF:p4Z status=idle
```

## Cause

`reconcileAttentionWith` (`internal/daemon/daemon.go`) is the only thing that surfaces an
agent already parked before the daemon subscribed — herdr's subscription delivers only
FUTURE transitions, and a continuously-blocked pane emits none. Two guards, in this order:

```go
handled := d.episodeHandled[a.PaneID]
if !handled { d.episodeHandled[a.PaneID] = true }   // set BEFORE the next check
if handled { continue }
if d.hasOpenEscalation(ctx, a.AgentID) { continue } // agent-wide, type-blind
```

Two problems:

1. **`hasOpenEscalation` is agent-wide and type-blind.** A pending `idle` escalation (#1053,
   a task suggestion that will never be auto-answered — `[escalations.auto_accept]` has no
   `idle` key) hid a genuinely `blocked` pane. This is inconsistent with the transition path,
   whose `duplicatePendingEscalation` keys on agent + type + normalized content and therefore
   happily raises a `choice` escalation while an `idle` one is pending.

2. **`episodeHandled` latches before a guard that can change its mind.** It is set on the
   first sweep and cleared only by a `working` transition or a pane recycle, so once an agent
   is skipped for having an open escalation, *resolving that escalation does not unstick it*
   — every later sweep short-circuits before re-reading escalations. The operator's obvious
   move (`hap dismiss <id>`) looks like it should work and silently does nothing.

Both skips are unlogged, so nothing anywhere says the agent is benched.

## Workaround

`hap capture <agent>` bypasses reconcile. Otherwise `hap dismiss <id>` **and** a daemon
restart are both required.

## Suggested fix

- Set `episodeHandled` only after the agent is actually re-driven, so a guard that later
  changes its answer can take effect.
- Scope `hasOpenEscalation` to the situation type (or drop it in favour of the transition
  path's existing dedup), so an idle escalation cannot hide a blocked screen.
- Log both skips.

[...]

PR #363, opened 2026-08-17
#1 fix: dedup an idle escalation before the task generator runs, and widen the window to 10m

0xGosu:
## What

Two fixes to escalation duplicate-ask handling, both found by verifying the dedup logic against a live herdr + real Claude agents.

**1. Dedup now runs before the task-generation LLM, not after it.**

The duplicate-ask check lived only inside `escalate()`. On the task-generation path that is too late: `generateTask` shells out to the operator's LLM CLI first, and its outcome funnels back through `escalate()`, which then discards it. Every suppressed duplicate still cost a full subprocess — tens of seconds to minutes, with tool calls, in the agent's cwd.

Measured on one agent in one morning: **14 task-generation subprocesses produced 8 escalations**, so six complete runs executed purely to be thrown away.

```
788 03:24:59 → 1082 03:25:25 ignored
790 03:31:20 → 1084 03:31:44 ignored
792 03:42:04 → 1086 03:42:26 ignored
798 05:00:05 → 1095 05:00:28 ignored
800 07:18:11 → 1097 07:18:42 ignored
803 07:21:59 → 1100 07:22:57 ignored
```

`HasPendingLLMConsult` does not cover this — it blocks only *concurrent* generations, never sequential ones.

**2. `escalationDedupWindow` 5m → 10m.**

The window is set by when the operator next looks at the pane, which the daemon does not control, so 5 minutes regularly expired before the re-fire arrived. Still well inside `store.AuditExcerptDedupMargin` (1h), so the retention pin holds.

## Why it is safe

The justification here matters more than usual, and the first version of it was **wrong** (caught in review).

The tempting argument is "safe to gate, because this path can only produce an escalation and never touches the pane." That is false: `escalate()` does not end at the audit row, it ends at `maybeFSPAcceptNow`, which under `[full_self_prompting] enabled` + `accept_generated_task` claims the row and **delivers the generated task to the pane**. With that mode on, a generated-task escalation is precisely a potential send.

What actually makes the early gate sound is **verdict equivalence**: `escalate()` applies this same verdict *before* its audit write and *before* the FSP hook, so a situation the early gate suppresses could never have reached the delivery either. The suppressed set is identical; only the LLM subprocess in between is skipped. The code comments now say this, and explicitly warn against restating it as the weaker rule — with FSP on, *every* escalation is a potential pane delivery, so that rule gives the wrong answer at any other call site.

`consultLLM` is deliberately **not** gated. `handleLLMOutcome` can promote a consult to a SEND on a path that does not run the dedup first, so gating it would convert a real answer into silence and leave the agent blocked. It also costs nothing to leave out — every wasted call measured above was a generation, not a consult.

## Tests

Five new tests, two of them mutation-proven:

- `TestDuplicateIdleEventSkipsTheTaskGenerator` — asserts the generator **call count**, which is the whole point: the escalation count stayed correct throughout the bug, which is why it shipped green. Reverting the gate makes it fail *while the "ignored duplicate" log still appears*.
- `TestResolvedEscalationSevenMinutesOldStillDedups` — the discriminating pin on the window bump: fails at 5m, passes at 10m. No fake clock needed — the window applies to the correction's `created_at`, and `MarkCorrectionSent` does not touch that column, so back-dating on insert ages the row deterministically.
- `TestNewIdleSituationStillReachesTheTaskGenerator` — the gate suppresses only genuine duplicates, so the fix cannot trade wasted LLM calls for dropped work.
- `TestConsultIsNotGatedByTheDuplicateCheck` — pins the scope limit above as a safety bound.
- `TestEscalationDedupWindowSpansTenMinutes` — guards that the query still spans the constant at all.

Full suite green (`go test -tags "vectors cpu" ./... -count=1`), `golangci-lint` clean, `gofmt` clean.

## Verified live

Swapped the daemon into a running herdr and forced three captures of one standing screen on a real Claude agent:

```
#1107  escalated          ← 1 task-generation LLM call
#1108  duplicated event   ← no LLM call, 3s later
#1109  duplicated event   ← no LLM call
gentask_calls = 1
```

Pre-fix that would have been three calls. The ~3s latency is itself the evidence the subprocess was skipped — the old path took 25–60s waiting on it.

## Deliberately not in this PR

- **The root cause of the duplicate escalations themselves.** For `idle`/`no_task_source`, dedup is keyed on normalized pane content, but that class's question ("this agent is parked with no task source") does not depend on the pane at all. An actively-driven agent's screen changes between every idle event, so the class can never dedup. This PR fixes the *cost* of that, not the cause. The fix needs a reason-keyed identity with a store-level field, and it must key on the escalate reason rather than situation type — `noop_vs_pending_tasks` is also `idle`-typed and must keep escalating.
- **An under-escalation bug found while testing.** Twice, an agent settling to idle produced no attention event at all — herdr never observed a "working" state for a fast reply, so no transition fired. Opposite direction, adjacent to #356.

### DIFF
--- a/changelog.d/fix-dedup-before-llm-and-window.md
+++ b/changelog.d/fix-dedup-before-llm-and-window.md
@@ -1,2 +1,3 @@
 - Fixed an idle agent's duplicate escalation costing a full task-generation LLM run before being discarded — the duplicate-ask check now runs before the generator is invoked, not after its result comes back. Measured on one agent in one morning, six of fourteen generation subprocesses were executed purely to be thrown away.
 - Changed the escalation duplicate-ask window from 5 to 10 minutes, so a re-fire of a situation the operator just resolved no longer slips through as a second escalation when they come back to the pane a few minutes later.
+- Fixed the startup warning for the retired `limits.escalation_dedup_window_seconds` key still telling operators the window is fixed at 5 minutes.
--- a/changelog.d/fix-dedup-before-llm-and-window.md
+++ b/changelog.d/fix-dedup-before-llm-and-window.md
@@ -1,0 +1,2 @@
+- Fixed an idle agent's duplicate escalation costing a full task-generation LLM run before being discarded — the duplicate-ask check now runs before the generator is invoked, not after its result comes back. Measured on one agent in one morning, six of fourteen generation subprocesses were executed purely to be thrown away.
+- Changed the escalation duplicate-ask window from 5 to 10 minutes, so a re-fire of a situation the operator just resolved no longer slips through as a second escalation when they come back to the pane a few minutes later.
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1732,8 +1732,19 @@ func Load(path string) (Config, error) {
 	// `escalation_dedup_jitter_percent`: the escalation duplicate-ask tuning is now
 	// a fixed internal constant. Probe the raw file to warn rather than silently
 	// drop them; Save omits them (no struct fields).
+	//
+	// The values below are spelled out by HAND and cannot be derived: they live in
+	// daemon.escalationDedupWindow / escalationDedupJitterPercent, and daemon
+	// imports config, so config cannot import them back. This package's own tests
+	// assert only on the KEY name, so nothing here catches the values drifting —
+	// which is how the window text went on claiming 5 minutes after it became 10,
+	// telling every operator with a legacy config something false. The guard is
+	// daemon.TestDedupConstantsMatchTheRetiredKeyWarnings, which sits on the side
+	// that can see both: it drives Load and fails if either sentence stops
+	// matching its constant. Keep the phrasing it greps for ("fixed N minutes" /
+	// "fixed N%") intact when rewording.
 	if legacy.Limits.EscalationDedupWindowSeconds != nil {
-		warnOnce("config key `limits.escalation_dedup_window_seconds` is no longer supported and is ignored; the escalation dedup window is a fixed 5 minutes",
+		warnOnce("config key `limits.escalation_dedup_window_seconds` is no longer supported and is ignored; the escalation dedup window is a fixed 10 minutes",
 			"path", path, "configured_value", *legacy.Limits.EscalationDedupWindowSeconds)
 	}
 	if legacy.Limits.EscalationDedupJitterPercent != nil {
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -1993,6 +1993,22 @@ func (d *Daemon) decideAndAct(ctx context.Context, situation domain.Situation,
 		// escalate()'s guard is what makes this safe (and why the obvious weaker
 		// justification is wrong). escalate() still re-checks, which covers the
 		// async outcome landing after the queue moved on.
+		//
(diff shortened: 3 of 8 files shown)


0xGosu:
@CharlieHelps please review

0xGosu:
Thanks — I took the two documentation findings, and I'm declining the atomicity change. Reasoning for each below.

## Fixed: `internal/config/config.go` (valid, and worse than a comment)

This one was a real operator-facing defect, not just rot. The retired-key warning literally told anyone with a legacy config that "the escalation dedup window is a fixed 5 minutes" — false as of this PR. `internal/config`'s own tests assert only on the *key name*, so nothing caught it.

Fixed the string, and added a guard so it cannot happen again: **`daemon.TestDedupConstantsMatchTheRetiredKeyWarnings`**. `config` cannot import `daemon` (daemon imports config), so the values can never be derived — but a test in `daemon` can see both sides. It drives `config.Load` with a legacy file and fails if either sentence stops matching its constant. Mutation-checked: reverting the string to "5 minutes" fails it with the exact diagnostic.

Also added a note at the warning block pointing at that guard and at the phrasing it greps for.

## Fixed: `internal/store/store_test.go:391` (clarified)

Taking this one, though I read the original slightly differently: it said *"The daemon derives this as now-window; here a 5-minute window"*, which attributes `now - window` to the daemon and frames the 5 minutes as this test's own pick — so it wasn't strictly claiming the daemon's value. But it's ambiguous enough to invite exactly the "fix" that would be wrong, so it now says explicitly that the span is the test's own and deliberately does **not** track `daemon.escalationDedupWindow`. Matches the wording used for the sibling case in `retention_test.go`.

## Declining: the atomicity change on `daemon.go:1996`

The race is real, but three things argue against the remedy:

**1. It is pre-existing, not introduced here.** `escalate()` does the same thing — `escalationAutoDismissReason` and `duplicatePendingEscalation` are reads, `ignoreDuplicate` is a separate write (daemon.go:2716–2721). This PR reuses that pattern rather than adding it.

**2. This call site narrows the window rather than widening it.** Before, the taskgen path ran the LLM subprocess *first* and only then reached escalate()'s read-then-write; the exposure was "a multi-second-to-multi-minute generation, plus the write". Now it is just the write. Moving the read earlier changes when the snapshot is taken, not whether one is taken.

**3. The cost is bounded to one skipped cycle, and it self-heals.** Nothing is staged before the gate — `StageLLMRequest` lives inside `generateTask` — so no pending request row, reservation, or file write is stranded, and I drop the auto-task claim on that path. The next attention event, idle re-drive, or `hap capture` raises the situation again. Nothing is delivered wrongly and no answer is lost.

Against that, making it atomic means a new store method fusing the dedup query with the audit write, and applying it to `escalate()` too — otherwise the two verdicts diverge and the equivalence argument this change rests on stops holding. That is a large change to the escalate path's contract, against a race whose worst outcome is a self-healing missed cycle. And the trigger is specifically *the operator retiring this very ask* in that window, which is arguably the case where skipping the generation is the right answer anyway.

I've documented the window, its bound, and this reasoning inline at the call site rather than leaving it implicit. Happy to revisit as its own change if you'd rather make both paths atomic together — it seems worth doing on its own terms, not folded into this fix.

PR #365, opened 2026-08-18
#1 fix: stop a truncated multi-tab capture stalling auto-accept forever, and loosen full self-prompting

0xGosu:
## The bug

Audit row #1092 (`choice`, a 4-tab Claude AskUserQuestion form with preview boxes, suggestion `1 1 1 1`, full self-prompting ON) sat `escalated` for ~11 minutes across **both** the FSP sweep and the 5-minute timed threshold. It was never delivered, never dismissed, and nothing said why. The operator cleared it by hand.

Root cause, proved from the stored row: `pane_excerpt` measured **4001 runes, leading `…`, only 3 of its 4 `[question N/4]` markers surviving**.

Every audit row stores `truncateTailRunes(s.Content, snapshotMaxRunes)` with a 4000-rune cap, which keeps the **tail**. That is right for every other capture — the prompt or error is at the bottom — and exactly wrong for a swept multi-tab aggregate, whose **head** is load-bearing: `AggregatedMCQFrames` reads the tab total out of the first `[question 1/N]` marker and requires N markers to follow. Shearing the head does not degrade that capture, it destroys it, and `mcqFormHeldStill` then answers `heldStillUnevaluable` on every sweep, forever, at Debug level only.

Guard 3 behaved correctly — refusing to judge a mangled capture is the designed fail-safe, and `CLAUDE.md` predicted this case in writing ("One more tab and the `[question 1/N]` head is gone"). The defect is that the capture was mangled at write time. The real 3-tab preview fixture in this repo is already 3606 runes, so 4000 never fit four tabs: this is every large form, not an edge case. The same truncation silently disarms `deliver/codex.go`'s staleness guard, which tests for `[question 1/`.

## What this changes

**The fix.** An aggregate capture gets its own 12000-rune budget (`truncateExcerpt`/`excerptBudget`), gated on a **strict parse** rather than marker-presence — at write time content is untruncated, so a real aggregate always parses, while the marker test would hand the big budget to any pane that merely printed `[question 1/4]`. Two deliberate exclusions: `SaveSignatureSnapshot` stays at 4000 (`signature_snapshots` has no retention path at all and `hap signatures show` prints it unpaged), and `duplicatePendingEscalation` truncates with the new helper but still passes `snapshotMaxRunes` as its `snapshotCap` **argument** — there it is the "is this a large tail window" threshold gating the two fuzzy dedup paths, not a storage budget, and passing the bigger one would disarm them for exactly the captures most likely to need them.

**Three loosenings, full-self-prompting only, all on the pending side.** Each is reached only after the ordinary comparison has already refused, each is gated on the `fsp` bool already threaded through `autoAcceptOne`, and each resolves to `heldStillUnevaluable` when it declines — none of them can turn a pending row into a dismissal.

- A row whose capture was cut short can still be answered when the live frame equals a **surviving block** of that row's own capture (`domain.SurvivingMCQFrames`). Truncation cuts from the top, so every block after the first surviving marker is byte-intact, and comparing against those is the same frame-wise relation the intact path uses.
- A pane-tail row is compared with `domain.TailSimilarWithin`, which cuts both salients to the shorter one's length **from the tail** before the trigram compare. The mismatch it exists for is structural: the baseline is minted from a `--source recent` consuming delta while every re-read is `--source visible`, so symmetric Jaccard refuses however little moved — the documented reason idle and generated-task rows never auto-accept.
- A `@noop` row is retired (`auto_dismiss_noop`) instead of queued forever. Nothing is ever typed for one and it can never become deliverable, so under a mode whose premise is that nobody is watching the queue it would sit indefinitely.

**Visibility.** Every pending refusal now names itself at info level, once per (row, reason). Two of them — Guard 1b suppression and the pane-busy skip — previously produced no output at **any** log level, which is why diagnosing #1092 took a five-round investigation and a live database query.

## Not included

**More than one delivery per agent per tick**, which was on the original plan. Investigating it turned up a hazard: the per-agent cap is not only a blast-radius bound. The sweep's caller withholds a just-answered agent because the pane "has input in flight that herdr has not reported acted on yet" — the TUI has not repainted. A second delivery is safe only while the second row's baseline cannot match the un-repainted screen, and it can: duplicate pending escalations for one screen do occur (16 near-identical rows observed on one live install), and on a plain menu the second send types a digit into a composer whose menu has already closed. Invalidating the pane cache does not help — a fresh read still returns the stale screen. The safe form is a per-**signature** cap and needs its own design pass.

## Accepted limits

- A form sitting on a tab whose block the truncation removed stays pending: there is nothing left to tell it apart from another form of the same size. After the budget fix no new row is ever truncated, so this only ever affects the existing backlog.
- For one retention window after upgrading, a multi-tab escalation that was already pending may be raised a second time — the stored capture was cut to the old size and no longer compares equal to a fresh one. Self-healing, and called out in the changelog.

## Review history

Two independent review rounds ran against this branch, and both found real defects that are fixed here.

1. **Critical.** LS-1 originally accepted "the live options are a subset of the row's stored option set". That is satisfiable by a **different** form of the same tab count, because every AskUserQuestion form ends in a generated `Submit answers`/`Cancel` tab whose labels sit in every stored union — one form's answer series could have been typed into another, violating this repo's own stated invariant. Replaced with the surviving-block comparison above; the option test is kept only as a drift-catching conjunct. Pinned by `TestFSPTruncatedAggregateRefusesADifferentFormOfTheSameSize`.
2. **Major.** `LiveMCQMatchesSalient` compared a masked stored salient against an unmasked live side, so any option label carrying a path or a large number could never match — a silent, one-directional no-op. Both sides now derive through `NormalizedOptionSet` then `MaskVolatile`.
3. **Major.** The pane-tail fallback ran after `SignatureHeldStill` refused but re-checked only one of its three refusal conditions. It now re-asks the other two (either side over-masked, or a fresh salient that has become structured) — two over-masked salients are mostly repeated placeholders and would clear any tolerance over any window.
4. **Major (tests).** Two of the new tests could not fail, and two comments had gone stale in the dangerous direction (one still prescribed the exact approach that caused finding 2). Both tests were rebuilt to discriminate and **verified by mutation**, as was the frame requirement and the mode re-check.
5. The `@noop` retirement did not re-ask the mode before dismissing, contradicting the doctrine `CLAUDE.md` states verbatim. Since the sweep walks a whole candidate list, an operator switching the mode off part-way through would still have the rest of their queue retired — and unlike a send, a dismissal cannot be undone.

## Verification

- `go build -tags "vectors cpu" ./...` clean; `go test -tags "vectors cpu" ./... -count=1` green across all 34 packages (`internal/daemon` included; its one observed failure was a known load flake that passes in isolation).
- `gofmt` clean, `go vet` clean on both tag sets, `golangci-lint run --build-tags "vectors,cpu"` reports 0 issues.
- Local integration suite against a **real herdr**: 10 passed, 6 skipped. Every skip requires `HAP_ITEST_CLAUDE=1`, which drives a real Claude and spends tokens, so it was left to the operator. This branch touches no herdr CLI interaction — `internal/deliver` and `internal/mcqdeliver` are unchanged — so a clean base run is the right evidence.
- Every safety-invariant test named in `CLAUDE.md` passes **unmodified**. The only changes to existing tests were mechanical (one new argument at two call sites), the `@noop` row moving out of `TestFSPExclusionsStillWait` with its behaviour re-pinned by three new tests, and the swept-escalation seeder now truncating like production — it was the only writer of that column anywhere that skipped truncation, which is precisely why every existing test was blind to this bug.
- New invariants and their paired test names are recorded in `CLAUDE.md`; a `changelog.d/` fragment is included.

**Still to do by the operator:** drive a real 4-tab preview form to an escalation on a live agent and confirm the stored excerpt keeps all four `[question N/4]` markers and that FSP delivers it — the exact #1092 shape, which only exists on a live install.


0xGosu:
@CharlieHelps please review

0xGosu:
All three findings verified against the code and fixed in `7fbe22c`. I reproduced each one before changing anything, and each fix is pinned by a test verified by mutation (I broke the fix and confirmed the test fails).

**1. Safety refusals reaching `auto_accept_failed` — confirmed, and the worst of the three.** `screenOutbound`'s error propagated out of the seam into `autoAcceptDeliveryFailed`, which burns an attempt and dismisses at the ceiling. So a never-auto match would be refused three times and then deleted from the queue — the inverse of FR-015. It is now tagged `errOutboundRefused` and handled like `errAgentDisabled`: claim reverted, no attempt consumed, row stays escalated. Worth noting the refusal is *permanent by nature* (the same text screens the same way every sweep), so the budget could only ever have ended in that dismissal. `TestFSPGeneratedTaskSafetyRefusalIsNeverDismissed` sweeps past the attempt budget — the existing test ran one sweep, which is why it could not see this.

**2. `accept_generated_task` not rechecked — confirmed.** It is resolved once per sweep into `allowGenerated`, and `fspStillOn` re-reads only `Enabled` and the ceiling latch. Now re-read immediately before the claim, for generated-task rows only. `TestFSPRechecksAcceptGeneratedTaskBeforeClaiming`.

**3. Kill switch not rechecked — confirmed.** Read before the loop and before the guard phase, never before the claim, and Guard 3's pane re-read is a herdr shell-out with a budget in seconds. Now re-read there, failing closed on a read error, and **not** gated on FSP since FR-017 stands the whole daemon down — `TestAutoAcceptRechecksTheKillSwitchBeforeClaiming` covers both flavours.

All three now go through one `claimBlockedBy` helper so the checks cannot drift apart. I also applied it to the `@noop` retirement, which your findings do not mention but has the same exposure: it is a dismissal, and Guard 1a returns before any dismissal precisely so pausing the herd never destroys the queue it protects — a pause landing mid-sweep should get the same answer as one landing a second earlier.

On the framing: you are right that these predate the branch and that it expands their reachability. Concretely, a generated-task escalation is `idle`-typed with an unstructured pane-tail salient, and `CLAUDE.md` records that Guard 3 answers `heldStillUnevaluable` on nearly all of those — the new tail-aligned comparison is what makes that population reach the claim at all. So they belong in this PR.

Full suite green, lint clean, and the new invariants are recorded in `CLAUDE.md` with their test names.

0xGosu:
@CharlieHelps review again

0xGosu:
Valid, and fixed in `6d258d0`. I verified every premise before changing anything:

- `ReclaimAbandonedAutoAccepts` has exactly one caller, at startup (`daemon.go:1103`).
- The candidate query is `WHERE status = 'escalated'`, and the daemon's own comment above the startup reclaim already says a stranded row is *"invisible to BOTH the operator's queue and the auto-accept query (each filters on 'escalated'), silently losing the escalation"*.

So a failed revert reproduces that exact loss through the error path instead of a crash. Your point about `false, nil` is right and I took it: only a non-nil error is treated as failure — no error with `ok == false` means another writer moved the row, which is legitimate.

**Fixed wider than reported.** The same exposure was already on the three pre-existing release sites (the generated-task `unhandled` branch, the `errAgentDisabled` branch, and the ordinary retry tail) — the safety-refusal path I added in `7fbe22c` was simply a fourth instance. Fixing only the reported one would have left three identical holes, so all four now go through a shared `revertClaim`, which keeps the retry obligation.

The retry (`retryAutoAcceptRevert`) sits beside `retryAutoAcceptFinalize` and is deliberately **not** gated on the kill switch or on the feature being enabled, for the reason that one is not: it is bookkeeping about a claim that has already been abandoned, and an operator pausing the herd must not be why an escalation stays hidden. `TestAStrandedClaimIsReleasedEvenWhilePaused` pins that half.

Two tests, both verified by mutation (removed the retry, confirmed both fail): `TestAStrandedClaimIsRetriedUntilItIsReleased` drives the safety-refusal path with a store that fails exactly the compensating write, asserts the row is genuinely stranded in `auto_accepting`, then asserts the next sweep releases it once the store recovers — no restart, no operator action.

Full suite green, lint clean. One `internal/daemon` test (`TestAutoSendIdleSkipsBusyAndBlockedAgents`, untouched by this branch) failed once in a full run and passed 3/3 in isolation plus two clean full-package runs — that package's documented load flakiness, not a regression.

PR #369, opened 2026-08-25
#1 feat: the daemon performs an operator's confirmed reply

0xGosu:
## Why

`hap` runs three processes against one state dir — the daemon, the TUI, and the
CLI (the last two sharing one `frontend.App`) — and responsibility for acting on
an agent was split between them. Some operator actions were queued for the
daemon; others were performed by the front-end process itself, shelling out to
herdr. Confirm-and-send typed keystrokes into a live pane from whichever `hap`
process the operator happened to be in.

Two processes driving the same pane is the problem, and it had concrete
consequences: the controls guarding the daemon's own sends did not apply to the
operator's at all.

This PR is the first part of moving that work behind the daemon. The front ends
write SQLite and nudge; the daemon acts.

## What changed

**A work queue with outcomes** (`agent_actions`). The front ends write a row and
nudge; the daemon claims it, executes it, and writes the result. The row carries
its own `status`/`error`/`result` because the control socket has never had a
reply channel — that is the only way a surface that queued an action can learn
whether it landed.

**Confirm-and-send moved.** `App.Resolve` records the correction and queues the
delivery in one transaction, then waits for the daemon's verdict, so the
operator still learns on the spot whether their answer reached the agent.
`internal/frontend` no longer imports `internal/deliver`.

Three things the move had to **add** rather than inherit:

- **`screenOutbound` on the reply.** The daemon's own sends are screened at
  `Decide` time; an operator authors their answer *after* the decision that
  raised the escalation, so nothing had ever screened it.
- **`WithAgentAutomation` around the send** — the cross-process per-agent
  barrier the operator path skipped entirely, so an operator disabling an agent
  could not stop a delivery mid-flight.
- **`ReadLines` from the daemon's own config.** The front end passed none, so an
  operator's reply verified forms against a different pane window than the
  daemon used.

`hap confirm --send` / `hap resolve --send` now require a running daemon and say
so with the `hap daemon --ensure` remedy, rather than recording a correction for
a reply nothing will deliver. Confirming *without* `--send`, dismissing, and
every config command still work with no daemon — the ordinary first-run order.

## Two bugs found on the way

**A vetoed reply cleared its escalation.** `applyCorrection` ends by flipping its
audit row to `resolved` whatever the delivery did. So a never-auto match left the
escalation gone, nothing typed, and the agent still blocked — with no row left
for the operator to look at. That is exactly what FR-015 says must not happen.
The branch is new: an operator's reply had never been screened before. The
refusal now withdraws its own correction, so nothing resolves a row that was
never answered.

**`store.Open` could open the wrong database.** The path was pasted into a URI
DSN unescaped, so a path containing `?` or `#` was truncated there and a
*different* database was opened — silently, with no error, and with every caller
under that path sharing one file elsewhere. Go's own `t.TempDir()` produces such
a path for any repeated subtest name (`#01`), so those tests had been sharing one
database in `/tmp` across runs. That is how a newly added column read as "no such
column" in a supposedly fresh store.

## Scope — deliberately partial

`Resolve` dispatches generated-task suggestions before the new code, so
`acceptGeneratedTask`, `appendGeneratedTasks` and `SendTaskToAgent` still call
`ports.SendToAgent` from the front-end process, and do not go through the
liveness check. `App.Herdr` is still present. Those, plus publishing agent state
to SQLite, `set_mode`/`capture`, and the invariant test that makes the rule
enforceable, are follow-ups.

Worth flagging for whoever takes the next one: `SendTaskToAgent` writes no
reservation row, no audit row, no handout counter, and takes no
`WithAgentAutomation` lock — so an operator hand-out is invisible to
`reclaimStrandedTasks` and races the daemon's own.

`agent_actions` has no retention path yet; terminal rows are kept readable on
purpose, for outcome polling.

## Testing

Full suite with the CI tags passes: `go test -tags "vectors cpu" ./... -count=1`.

Every gate here is proved by **mutation** — removed, watched the test fail,
restored:

- the `pending` guard on the claim, and the `running` guard on the outcome
- the correction+delivery transaction (split it, the rollback test fails)
- the withholding filter, in all three of its clauses
- the startup reclaim, and the drain running before `processCorrections`
- the URI escaping
- withdrawing a refused reply's correction

The delivery tests moved to `internal/daemon`, where they now run against a fake
that models what a keystroke actually does to a live form. The front-end tests
keep what is still the front end's job: that a delivery is queued, linked to its
correction, and refused when no daemon could run it.

**Not run:** `golangci-lint` — its bundled Go cannot parse `math/rand/v2` in this
environment. `queueAgentAction` has no caller until the remaining sends move, so
the `unused` linter may flag it.

The local integration suite (real herdr + Claude) has not been run;
`TestRealConfirmDeliversMenuDigit` is the direct regression for this change and
should be run before merge.

https://claude.ai/code/session_011ZScND9kmQ5jkdUfrBmv5s


0xGosu:
@CharlieHelps please review

PR #375, opened 2026-08-27
#1 feat: LLM presets learn into AUTO.md; concise, corrected README and hap skill

0xGosu:
## Why

Two separate problems, one branch.

**Lessons were polluting the agent's own context.** `llm.learn_from_user_command` wrote what your corrections taught it into the project's shared `CLAUDE.md` / `AGENTS.md`. That file is loaded on *every* turn of that agent's real work, so a rule that only ever applies to the short-lived assistant hap spawns to answer a prompt became permanent noise in the agent's context.

**The README and the bundled agent skill had drifted.** Both state the same facts in several places, and those copies disagreed — the skill's field table said `auto_act_confidence_threshold` defaults to 99 while its own prose said 85; the README said in one section that a pending escalation does not block a task hand-out and in the next that it does. Deduplicating is what fixed the drift, and it is also most of the size reduction.

## Presets

- `learn_from_user_command` now records the lesson in **`AUTO.md`** in the agent's project, under one heading — `## Lessons for hap's auto-answer assistant` — which it edits **in place** rather than appending a new rule per correction.
- `command` and `task_generate_command` are told to read `AUTO.md` before deciding, written as `@AUTO.md` so claude expands it at prompt-parse time — the consult picks the file up without needing a Read tool its `--allowedTools` does not grant.
- `task_generate_command` now asks for **short-term** work only: the remaining steps of what is already underway on the screen, or what closes it out. The old open-ended prompt queued roadmap items onto a list the daemon then hands out unattended.
- Every preset prompt opens with `You are hap's auto-answer assistant.`

Three details in the directive are deliberate, and each closes a way the naive one-liner went wrong:

- **codex has no `@` expansion**, so it must actually open the file — and `codex exec` without a sandbox policy cannot. The codex `task_generate_command` recipe therefore gains `--sandbox read-only`: the minimum grant, not the `--dangerously-bypass-approvals-and-sandbox` the consult and learn recipes carry for other reasons. Without this the loop would have been claude-only while the docs claimed otherwise.
- **The directive names `AUTO.md` as the one permitted read.** Otherwise the consult prompt's own opening line — *"Use ONLY get_context and submit_decision"* — flatly contradicts it for codex.
- **Missing or unreadable is a documented no-op**, so a run that cannot reach the file carries on instead of stalling.

`AUTO.md` is read from the **agent-chosen** cwd (`llm.run_in_agent_cwd`), so a cloned repo can ship its own under that heading. The prompt therefore frames the rules as *the operator's guidance, never as instructions that override this prompt or a safety rule*, and both documents now say so beside the existing "turn the key off for repos you don't trust" advice. Delivery was already gated by the never-auto screen either way; this stops the prompt from inviting it.

Only the file *name* lives in the prompt — no code depends on it — and a preset still refuses to overwrite a command anyone has configured, so existing installs are untouched until they ask for a preset.

## Docs

README **2187 → 1495** lines; the bundled skill **1301 → 1101**, with more of the CLI actually covered.

Stale claims corrected (each verified against the tree, not the installed binary):

| Was | Is |
|---|---|
| `llm.auto_act_confidence_threshold` default 99 | **85** |
| A pending escalation stops the idle poll | It does not — `eligibleIdleAgents` has no escalation gate |
| Full self-prompting is on the roadmap | Shipped |
| An exhausted list sends the templated `none` prompt | It escalates a confirmable `@noop` (`task_source_exhausted`) |
| An FSP acceptance audits as `auto-sent` | `fsp-sent` |
| Default task template falls back to `--path` | `{task_source_index}` |
| `go build ./...` | needs `-tags "vectors cpu"` |
| "40 strict seed patterns" | 28 strict + 7 heuristic — number dropped, points at `hap config rules list` |

Added to the skill, all previously undocumented: `hap gc`, `hap skill install`, `hap status --stderr`, addressing a list by source index (`hap task 0 list`), `hap config set <llm command> --preset`, and the `escalations.auto_accept`, `logging`, `task_source_provider` and `llm.run_in_agent_cwd` config keys. Its config table was generated from the live defaults, so every key in the `ConfigFields` registry is listed with its real value.

The skill also gained a short **"rules for an agent driving hap"** section up top, collecting the facts whose absence causes damage — never type into a pane hap escalated, `resolve --action` takes the label not the digit, what `--send` does, which commands are read-only.

## Notes for review

- `.claude/skills/hap/SKILL.md` is `go:embed`-ed into the binary (`skilldoc.go`), so this is a code change, not a docs change. The diff touches `.go` files, so `auto-release.yml`'s `paths-ignore` does **not** apply — merging this cuts a release (0.6.20 → 0.6.21) and the updated skill ships with it.
- `sample/config.toml` was regenerated from the Go preset table rather than hand-edited; `TestSamplePresetsMatchTheGoRecipes` holds the two active recipes byte-identical.
- No test asserts README or skill *content* beyond the `name: hap` frontmatter, which is unchanged.

## Testing

- `go test -tags "vectors cpu" ./... -count=1` — full suite green.
- `bash scripts/check-submodule-gitlink.sh` — OK.
- `golangci-lint` cannot run in this environment: its bundled parser rejects the installed Go stdlib's `math/rand/v2`, which is unrelated to this diff. CI uses its own pinned versions.

https://claude.ai/code/session_01HGvVrupMDWnjSbF3FWx1NK


0xGosu:
@CharlieHelps please review

PR #377, opened 2026-08-31
#1 feat!: remove the llm.command_start / task_generate_command_start family; add AUTO.md

0xGosu:
Removes the first-interaction command family — nothing in the shipped recipes ever set it — and adds the `AUTO.md` that hap's own runs read and write.

## AUTO.md

`llm.learn_from_user_command` writes lessons to `AUTO.md` in the agent's project, and the shipped `command` / `task_generate_command` prompts read it back via `@AUTO.md`. This adds the repo's own, seeded with one rule migrated out of `CLAUDE.md` so it steers hap's runs instead of loading into every turn of an agent's real work. The heading is byte-identical to the one the shipped prompts edit in place.

## The removal

Gone: `llm.command_start`, `llm.task_generate_command_start` and their `_env` / `_env_file` companions, plus the adapter's start templates, `domain`'s `First` flags, the daemon's `firstConsult` / `firstTaskGen` maps, the frontend/TUI registry rows, and `LLMPresetFollowUp` (whose only job was warning that `task_generate_command_start` was still unset).

Two behaviors existed **only** to serve those keys and went with them:

- **The fast-fail retry** in `ConsultWithSession`. Its "alternate template" was always the start template, so with one template there is nothing to retry. It was there for claude's `--resume` rejecting a not-yet-created session — and hap now injects `--session-id` and never resumes, so the original motivation is gone too.
- **Refilling an exhausted declared task source**, gated on `GenerateTaskConfigured && GenerateTaskStartConfigured`. That gate can no longer be satisfied. Since no shipped preset ever set the second key, the branch was already dead for every install, so an exhausted source now always escalates `task_source_exhausted` with a confirmable `@noop` — which is what everyone already gets. It is one boolean key away if we want it back.

## Upgrade path

Follows the existing `rewrite_command_start` precedent: all six keys are probed in `legacyKeys`, a config carrying any of them loads with one warning and is rewritten without it on the next save. Nothing breaks at runtime.

## Versioning

`feat!:` with three `Breaking.` changelog lines, so this takes CLAUDE.md's **manual minor path**: `herdr-plugin.toml` → `0.7.0` and `scripts/assemble-changelog.sh 0.7.0` run in-branch, because no bump commit is created for the fragment to assemble into. Kept as its own commit (`3475a25`) so it can be dropped alone if this should ship as a patch instead — nothing in it touches code.

## Notes for review

- The `sourceExhausted` parameter of `daemon.generateTask` is now **dormant**, not deleted. It is kept with its `max_tasks` cap guard, commented as such: a refill path returning without the cap would bury the operator silently. Only tests reach the true branch today, and the `max_tasks` docs are qualified to match.
- `TestIdleDeclaredTaskExhaustedGeneratesMoreTasks` is repointed rather than deleted — it now asserts **zero** generator calls *with* `task_generate_command` configured, which is what makes it a guard against a half-restored refill path.
- `TestRemovedStartCommandKeysWarnAndDrop` is one config **per key**: the Load check is a single `||` chain, so one config carrying all six passed even with five toml tags misspelled. Mutation-checked — misspelling one tag fails exactly that subtest.
- Added `TestConsultStagedDecisionSurvivesANonzeroExit`. Deleting `fastfail_test.go` dropped the only coverage of a behavior that still exists: a CLI that stages a decision and *then* exits nonzero must still yield it. Every other consult test exits 0.

## Verification

`gofmt` clean, `go build -tags "vectors cpu" ./...` ok, `go test -tags "vectors cpu" ./... -count=1` → exit 0, 35 packages, no failures. `golangci-lint` reports 2 `typecheck` errors inside Go's own stdlib (`crypto/internal/randutil` → `math/rand/v2`) — a linter/toolchain version mismatch in the dev container, unrelated to this diff.


0xGosu:
@CharlieHelps please review

PR #378, opened 2026-08-31
#1 fix: a [no_task_source] confirm recommends enabling task generation

0xGosu:
## What

An escalation tagged `[no_task_source]` **with an empty suggestion** is not an answerable prompt — it is a notice. `domain.Decide` raises it when an agent goes idle, no `[[task_sources]]` entry matches it, no native todo is inferable, and `llm.task_generate_command` is unset, so FR-011's relaxation never fires and hap invents nothing.

Confirming one (`hap confirm <id>`, `--send`, TUI `Enter`, TUI `y`) died with:

```
error: audit record 42 has no suggestion to confirm: no action could be resolved for it.
Answer it explicitly with `hap resolve 42 --action TEXT --send`, or drop it with `hap dismiss 42`
```

That reads as a broken plugin, and neither escape it names addresses the cause. It now shows what to configure instead:

```
audit #42 [no_task_source]: the agent is idle and no task source matches it, so there is
nothing to confirm — this is a notice, not an answerable prompt.

Let hap propose the next task with an LLM (install one recipe):
  hap config set llm.task_generate_command --preset claude
  hap config set llm.task_generate_command --preset codex

Or give the agent its own checklist:
  hap config task-source add --agent <name> ./docs/tasks.md

Nothing to do about this one? Drop it with:
  hap dismiss 42
```

CLI exits 0; the TUI opens it in the scrollable detail overlay. The row stays **pending** — nothing is learned, resolved or dismissed.

## Why the branch is scoped by construction

`[no_task_source]` does *not* always mean an empty suggestion — the idle-handout and learned-rule paths in `decide.go`, and the task-gen success / `@noop`-decline outcomes in `handleTaskGenOutcome`, all stamp the same tag with a **confirmable** suggestion. The new branch lives *inside* `App.Confirm`'s `action == ""` guard, so those never reach it.

The one other empty-suggestion producer — the daemon's "task source is configured but unusable" escalate — passes a `Decision` whose `Reason` is unset, so its rationale is `"[] idle with no task source…"` and `EscalationReasonTag` returns `""`. It keeps the ordinary error, correctly: the fix there is the broken source.

That is why the guidance can name `llm.task_generate_command` without reading config. The precise claim is that the key was unset when the row was **raised** — an operator holding several notices will see `ApplyLLMPreset` refuse the second (`"already configured — a preset only bootstraps an unset command"`), which is correct and self-explaining.

## Changes

- **`internal/frontend/notasksource.go`** (new) — `NoTaskSourceNotice`, a typed error with a one-line `Error()` and a multi-line `Guidance()`. The preset lines are built from `LLMTaskGenerateCommandKey` and `LLMPresetNames`, so renaming the key or adding a third recipe cannot leave stale advice. The `--preset claude` spacing is load-bearing: `config set` matches the flag by equality on its own argument, so `--preset=claude` falls through and is stored as a literal one-word argv, silently.
- **`internal/frontend/frontend.go`** — the branch inside `Confirm`'s existing guard.
- **`internal/cli/{cli,help}.go`** — the handler intercepts with `errors.As`, prints `Guidance()` and returns nil. `confirm` became `SelfHints`: its static footer points at a recorded action and a graduated rule, neither of which a notice produces.
- **`internal/tui/tui.go`** — the notice opens the detail overlay (see below); the batch `y` keeps it out of its `%w` error wrapper, and reports a count only when something actually confirmed.

## Two things worth a reviewer's attention

**The TUI surface changed during review.** Routed to the ephemeral message area, the guidance overflowed the pane: `chromeRows` budgets a message as a flat 2 rows, and this is the first multi-line message in the TUI — 23 rows drawn into a 12-row pane, pushing the help line and the escalation rows off screen. Widening the budget cannot fix it either, since the block is taller than a short pane. It opens the **scrollable detail overlay** instead, carrying no `confirmID`/`retryID` — there is nothing on it to confirm, which is the message.

**`detailPageSize` had a pre-existing bug, fixed here and separable from the feature.** It was a flat `chrome := 8` while `renderDetail` draws the message *and* status areas, so any overlay opened with either showing ran two rows over. Invisible while nothing set one before opening an overlay; this notice sets both at once. Now mirrored from `chromeRows`, as AR-002 says it should be. It shipped green because there is no generic detail-overlay pane-fit test — the notice test is now the only guard.

## Accepted limitations

- A mixed batch where a **real** error sorts before the notice returns the error branch and drops the guidance. The failure is the more urgent thing, and the notice row is still pending, so confirming it alone shows the guidance.
- `hap escalations`' footer still offers `hap confirm <id> --send` for the newest pending row, including these notices. That is how an operator lands here; out of scope for this fix.

## Testing

Full suite green under `-tags "vectors cpu"`; `gofmt` and `go vet` clean. Six mutation checks confirm the tests are not vacuous — the scoping guard, the TUI routing, the overlay, the `detailPageSize` budget, the zero-count status, and the footer assertion each fail their test when reverted.

New: `internal/frontend/notasksource_test.go` (there was previously **no** test of `errNoSuggestion` at all), additions to `internal/cli/confirm_test.go`, and `internal/tui/notasksource_test.go`. Controls included: every other unconfirmable reason still reports the old refusal, and a `[no_task_source]` row *with* a suggestion still confirms normally.

The local integration suite is not applicable: this change adds no herdr call, no delivery and no pane read, so nothing in `test/integration/` exercises the new path.


0xGosu:
@CharlieHelps please review

PR #384, opened 2026-09-02
#1 fix: sync every live claude agent when the session-name setting is turned on

0xGosu:
## The bug

Flipping `[agents] sync_claude_session_name` from `false` to `true` did nothing visible to an already-running herd.

The sync is a side effect of a pane **capture** (`daemon.handleAttention` → `syncClaudeSessionName`), and nothing re-captures on a config change:

- `reloadWith` swaps the config and schedules no captures.
- The `reconcileAttention` that follows a `KindReload` nudge skips every pane already in `episodeHandled` (set on the first sweep after start, cleared only by a `working` transition or a pane recycle), skips `working` agents outright, and skips any agent with an open escalation.

So on a settled herd the flip reached nothing. Each agent only synced on its next `working → parked` round-trip — hours for a parked herd, and never for one sitting on an approval.

## The fix

A false→true flip drives its own one-shot pass (`daemon.syncClaudeSessionNamesNow`) over the live agents: `ListAgents`, skip non-claude on the listed type, `EnsureAgentName`, read the pane, run the shared decision half.

Four bounds are load-bearing and each is tested:

- **`--source visible`, never the consuming delta.** Non-consuming is required, not merely preferable: a `--source recent` read here would swallow the delta a pending classification capture is about to take. It is also the only reason the flip sees a composer on a quiescent pane at all.
- **Not `episodeHandled`.** Clearing it would have re-driven the whole herd through classify → decide → act, raising escalations and spending LLM consults for what is a naming feature.
- **`!first`.** `reloadWith` also runs inside `New()`, before `Run` exists; the startup reconcile already re-drives every parked agent, which is the only set Path 2 can push to.
- **The latch is released by the goroutine's own defer *and* by hand when `spawn` refuses**, or one shutdown-race flip would disable the pass for the process lifetime.

Both entry points now share `applyClaudeSession`, so a gate added to either is added to both.

## Safety

Unchanged. Path 1 only writes the store. Path 2 still goes through `startSessionRename`: parked (idle/done) agents only, `acquirePane`, kill switch and per-agent disable re-asked inside the goroutine, the never-auto screen over the exact text, a proven-empty composer, a re-read before and after the send, and the 3-push ceiling per (agent, terminal, name).

## Scope — deliberate narrowing

This fixes the **flip** path. `handleAttention`'s classification read keeps its `--source recent` blind spot: an ordinary capture whose delta shows no composer still returns `ok=false` ("UNKNOWN, never unnamed") and waits for the next one. That is existing, intended behaviour and is not addressed here.

## Tests

`internal/daemon/sessionname_test.go`, all mutation-proved:

- `TestFlippingSessionSyncOnRenamesTheLiveHerdWithoutACapture` — the claim: no transition is ever pushed through `h.events`.
- `TestFlippingSessionSyncOnAdoptsANamedSession` — Path 1 through the same pass.
- `TestReloadWithoutAFlipRunsNoSessionSyncPass` / `TestDaemonStartWithTheSettingOnRunsNoSessionSyncPass` — the `!first` and no-flip controls. Asserted on the absence of a keystroke, not on `listAgentsCalls`, which the nudge's own reconcile contaminates.
- `TestSessionSyncPassSkipsNonClaudeAgents` / `…DoesNotRunTwiceAtOnce` / `…ReleasesItsLatch`.

The tests wrap the fake so the composer is visible **only** through `--source visible`. Without that the capture path could produce the same rename and none of them would discriminate.

Verified live that `herdr agent list` carries the `agent` field the type gate reads — otherwise the whole pass would be a silent no-op that every fake-driven test still passes.

## Checks

- `go build -tags "vectors cpu" ./...` — clean
- `go test -tags "vectors cpu" ./... -count=1` — full suite green
- `gofmt` / `go vet -tags "vectors cpu" ./...` — clean
- `golangci-lint` **could not run locally**: it fails typechecking the Go stdlib (`math/rand/v2: method must have no type parameters`) before reaching this repo's packages, identically on a clean `main` checkout. Toolchain skew in this devcontainer — CI is the gate.
- Integration suite: `./test/integration -run SessionName` compiles and skips cleanly (needs `HAP_ITEST_CLAUDE=1`, which spends tokens). Not run — those cases pin the composer *parse shape*, which this change does not touch.


0xGosu:
@CharlieHelps review

PR #386, opened 2026-09-05
#1 feat: the daemon drives herdr, and the front ends talk to it through SQLite

0xGosu:
Stages 1–3 of the **TUI ⇄ SQLite ⇄ daemon ⇄ herdr** migration. No front end drives a live pane or reads live herdr state any more, except at the call sites the new guard still lists by name.

Two processes driving the same panes is what lets an operator's hand-out race the daemon's own, and none of the daemon's delivery guards (never-auto screen, kill switch, per-agent disable, `WithAgentAutomation`, Guard 3's re-read) apply on the front-end path. PR #369 landed the `agent_actions` substrate; this finishes the stages that ride it and closes the read side, which had no DB channel at all.

## What changes for an operator

`hap capture`, agent focus, and (in the TUI) anything that answers "what is running" now need a live, non-stale daemon. Each says so with `no healthy hap daemon is running` instead of failing obscurely. Config writes, dismissals and record-only confirms still work with the daemon down — that is the ordinary first-run order and it must not regress.

Two behaviours are now honest where they used to be silent:

- `hap capture reviewer` naming an agent that does not exist, or catching one mid-work, used to print `capture queued` and do nothing. Both now fail with the reason.
- An empty agent list used to read as `no agents detected (is herdr running?)`. Since the herd is now what the daemon published, the likelier cause is that nothing has looked — so `hap agents`, `hap status` and the TUI say which of *"no daemon has reported yet"* and *"the last report is N old"* it is. The wording deliberately blames neither herdr nor the daemon by name: a healthy daemon that cannot reach herdr publishes nothing either, and sending an operator after the wrong process is the failure this replaces.

One consequence is documented in the changelog rather than fixed: an agent that **disappears** produces no event, so with no TUI open a closed pane can still be listed for up to a minute, until the sweep reconciles.

## Stage 1 — `focus`

New `focus` action kind. Enqueue and nudge, **no await**: focusing moves the herdr UI away from the TUI pane, so the operator is not looking at the surface that would render a verdict. The status line says it *asked* rather than claiming the pane moved. Bounded by staleness — a pending row that outlives a dead daemon must not be replayed hours later into whatever pane you had moved on to.

## Stage 2 — `capture`, and the last payload-carrying nudge

The declared `capture` kind gets its executor, and the daemon resolves the target (short names included — the old front end resolved them first, and comparing ids only would have kept every pane-id test green while `hap capture reviewer` matched nothing). `control.CaptureKind`/`CaptureTarget`/`NudgeCapture` are deleted, restoring `internal/control`'s own "nudges carry no domain payload" contract without a carve-out.

## Stage 3 — the `agent_roster` snapshot

Three new tables written only by the daemon. The publisher reuses the `ListAgents` calls the daemon already makes plus the transitions it already receives, so a quiescent herd costs nothing; a 2 s tick runs **only** while a `hap tui` process is registered in the existing `tuisession` flock directory, because an unconditional ticker would be new permanent idle polling on every install.

Decisions worth reviewing, each mutation-proved:

| Rule | Why it is not optional |
|---|---|
| A roster's **timestamp** decides freshness, not its row count | An empty herd and an absent daemon are otherwise the same answer |
| `gone_at`, never `DELETE`; a changed `terminal_id` **replaces** | Herdr recycles pane ids and an agent id IS a pane id |
| Unobserved is never evidence — including on the **write** | Herdr's status events carry no terminal id, so writing one through would erase what a publish recorded and leave the recycle check *disarmed*, not merely uninformed |
| Order is herdr's own listing (`list_seq`), not `agent_id` | `w1:p10` sorts before `w1:p2` — invisible in every fixture, visible to anyone running ten panes |
| A **discovery** event is refused | `AgentBusy("detected")` is TRUE, so recording herdr's replayed `pane.agent_detected` would make `hap task send` refuse a perfectly idle agent, herd-wide, on every reconnect. An empty status is worse — it reads as not-busy, so the gate fails *open* |
| The shell-out half runs off the select loop, **once at a time** | `store.Open` hands out two connections; inline, this made an unrelated rename miss its 3 s deadline. The latch closes the same starvation by the second door — a pass outliving the 2 s tick would stack passes on that pool |
| TTLs know whether anyone is looking | 20 s/60 s watched, 5 min unwatched. Refusing to shell out at all when unwatched would leave `hap agents`' cwd column blank forever on a CLI-only install, since reads never nudge |
| A cwd write is scoped by `terminal_id` | The loop can publish a recycled pane mid-pass, and an unscoped write paints a dead process's directory onto its successor |

## Stage 0 — the guard that makes the rest self-verifying

`TestFrontendNeverTouchesHerdr` walks `internal/frontend`, `internal/tui` and `internal/cli` and bans herdr-backed `ports` symbols, the `Herdr` field selector and herdr binary literals. Exemptions are **per symbol**, each carrying its reason and the stage that removes it, and a stale one **fails** — so the list can only shrink. Banning the selector matters as much as the symbols: `app.Herdr.Send(ctx, pane, text)` names no banned symbol, imports no adapter and contains no herdr literal, yet it is exactly the pane-driving call the rule exists to stop.

## Two traps recorded in the code

- The store's `unix()` is `UnixMilli`. Reading a timestamp back as **seconds** yields the year 58648, so `now.Sub(publishedAt)` is negative and *every* freshness check passes — including the tests written to catch staleness. `TestARosterTimestampSurvivesTheRoundTrip` pins it.
- A backtick inside the raw-string schema terminates the literal, which is why the `list_seq` comment says "agent listing" rather than naming the command.

`list_seq` also gets its `migrate()` ALTER, or every machine that ran an earlier build of this branch fails with `no such column`.

## Verification

- `go build`, `gofmt`, `go vet` clean; `golangci-lint` reports only the two pre-existing `typecheck` errors in `crypto/internal/randutil` that `main` also reports (devcontainer toolchain mismatch).
- Full suite `go test -tags "vectors cpu" ./... -count=1` — **two consecutive clean runs**.
- Every new assertion proved by mutation (11 mutations, all caught): roster ordering, event-preserves-position, unknown-sorts-last, terminal preservation, terminal-scoped cwd write, nil-vs-empty locations, the pass latch, the location TTL, the negative cwd cache, the `detected` refusal, and the `agentCwd` precedence.
- Local integration suite against real herdr 0.8.2: `TestRealPaneInfo`, the notification cases, `TestRealConsultContextTaskSourceSummary` and `TestRealIdleUnlearnedSignatureConsultsInsteadOfReviewing` pass. `TestRealConfirmDeliversMenuDigit` / `…RemoteEnvSelection` fail with `no healthy hap daemon is running` — **verified identical on `main`**, so pre-existing since #369 made confirms daemon-mediated; those two cases never start a daemon.
- e2e harness runs the pipeline to an escalation. It cannot exercise the publisher: its fake herdr does not implement `agent list` at all (`parse agent list output: unexpected end of JSON input`, on `main` too), so nothing publishes and `hap agents` correctly reports "no daemon has reported yet" rather than an empty herd.
- **A live-daemon check against the running herd was deliberately skipped**: a second daemon takes consuming `pane read --source recent` deltas from the real one and could act on live agents. Worth running by hand via the `hap-development-local` skill before merge.

## Still to come

Stage 4 (`set_mode` + `read_mode`, moving the whole Shift+Tab rotation into the daemon), Stage 5 (`send_task`, closing the `Resolve` short-circuit and introducing the `TaskSender` seam for the `InDaemon` split), Stage 6 (drop `App.Herdr` from the operator App and empty the guard's exemption list).


0xGosu:
@CharlieHelps please review

PR #388, opened 2026-09-05
#1 fix: keep the roster publisher off the daemon's loop and out of its writes

0xGosu:
Fixes a regression from #386. On a two-core runner the `internal/daemon` package went from passing to **failing 2 runs in 3**, a different test each time — and CI's ubuntu job failed twice on this pattern. One failing run named the real mechanism:

```
reconcile: terminal-id sync failed agent=pA error="database is locked (517)"
```

## What was actually wrong

The roster publisher landed as a new, **frequent writer** on a store that handed out two connections and began every transaction as a reader.

`517` is `SQLITE_BUSY_SNAPSHOT`. A deferred transaction reads first and asks for the write lock at its first write; if another connection committed in between, its snapshot is stale and SQLite refuses **immediately**. `busy_timeout` deliberately does not retry — waiting cannot make an outdated snapshot current. Every caller in this repo logs a store error and carries on, so the write simply never happened. That one left an agent's `FirstSeen` unreset after its pane had been recycled: a silent, wrong result, not a crash.

This was not a flaky test. It is a lost write, and it is reachable in production wherever two goroutines write at once.

## Three changes, each measured

**1. The publish runs off the select loop.** #386 moved the *shell-outs* off it but left the store write on the loop that handles every agent's transitions, nudges and timers. A listing arriving while a pass runs now **replaces** the pending one — latest-wins, because an older snapshot can only describe a herd that has moved on, and dropping it outright would lose the vanish reconciliation only a full listing can perform.

**2. Transactions take their lock at BEGIN** (`_txlock=immediate`). All 51 transactions this package opens write, so starting as a reader buys nothing and exposes every one of them to 517. It costs no concurrency — WAL admits a single writer either way — it only moves *where* the wait happens, turning an unretried failure into a bounded one.

**3. The pool is sized for readers** (2 → 8). WAL allows many concurrent readers beside its one writer. At 2, the whole process had one writer and **one reader**, so an unrelated read queued behind a write inside `database/sql` for no reason SQLite imposed. The previous comment justified the cap as avoiding "needless SQLITE_BUSY", which (2) now handles properly.

(2) and (3) are a pair, and the measurement says so: **deferred at pool 8 is worse than the status quo** — 1 of 4 runs passed.

## And one frequency change

Publishing no longer rides the **nudge** path. That is the daemon's busiest one — every operator action and every self-nudge — and no *read* ever nudges: `hap agents` and `hap status` are pure reads that never woke the daemon. So those publishes bought no reader anything while costing the daemon's own bookkeeping its turn at the lock. Startup, the once-a-minute sweep and the 2 s TUI tick still publish, which covers every path a reader can observe.

## Measurements

`internal/daemon`, `taskset -c 0,1` (matching CI's two cores), whole package, `-count=1`:

| Tree | Passing |
|---|---|
| pre-merge main (`8fee60b`) | **5/5** |
| merged main (`cb9e5e7`) | **1/3** |
| + off-loop publish only | 3/5 |
| + immediate transactions, pool 8, no nudge publish | **9/10** |

Targeted, in isolation: `TestReconcileSyncsTerminalIDs` went from ~2 failures in 6 to **25/25**; `TestAutoAcceptEndToEndThroughTheRealSweep` from ~1 in 18 to **30/30**; `TestEscalationAutoDismissesUnavailableAgentWithAudit` from ~1 in 32 to **40/40**.

The isolating experiment that identified the cause: with `publishRoster` stubbed to return immediately, the same test ran **40/40** clean.

## Verification

- Full suite `go test -tags "vectors cpu" ./... -count=1`: clean.
- `GOTOOLCHAIN=go1.26.2 golangci-lint run --build-tags "vectors,cpu"`: 0 issues. (The pin matters — an unpinned run in this devcontainer cannot typecheck the go1.27 stdlib, which silently disables `unused`. That is how #386's lint failure got through locally.)
- New test `TestAListingArrivingMidPassIsStillPublished` pins latest-wins, proved by mutation: dropping the mid-pass listing fails it.
- Roster tests now drive the publish through the sweep path (`publishNow`) rather than a nudge, since the nudge deliberately no longer publishes.

## One thing deliberately not tested

`_txlock=immediate` ships without a dedicated regression test, and the code says so. Every version I wrote that reproduced the 517 did it by saturating the write lock, which then times the other writer out at `busy_timeout` under **both** modes — so the test passed either way and would have been a guard proving nothing. It stands on the observed failure above and on the pool measurement, not on a test I could not make discriminate.

## Residual

1 run in 10 still fails on `TestAutoSendIdleCapsAtExactlyMaxHandouts`, whose `waitFor` deadline is 5 s — exactly `busy_timeout(5000)`. That equality is pre-existing and makes the test marginal under any write contention; worth widening separately.

Note the baselines are not the same sample size: the pre-merge figure is 5 runs, not 10 — a longer baseline run was killed by memory pressure on this machine. So "9/10 vs 5/5" is the honest comparison, and it does not rule out a residual difference of that order. What it does rule out is the failure rate this started at, which was 2 runs in 3.

### DIFF
--- a/changelog.d/fix-roster-publish-connection-pressure.md
+++ b/changelog.d/fix-roster-publish-connection-pressure.md
@@ -1,0 +1,14 @@
+- Fixed the daemon losing bookkeeping writes while it published the running
+  agents. Publishing took a database transaction on the loop that handles every
+  agent, on the same paths as the daemon's own work — and because those
+  transactions began as readers, one that committed in between made another
+  fail outright with `database is locked (517)`, which is not retried. The
+  daemon logs such a failure and carries on, so the write simply never
+  happened: an agent's first-seen time was left unreset after its pane had been
+  recycled. Publishing now runs off that loop entirely, transactions take their
+  lock up front, and the database hands out enough connections that a read no
+  longer waits behind an unrelated write
+- Changed when the herd is republished. It happens at startup, on the
+  once-a-minute sweep, and every two seconds while a TUI is open — no longer on
+  every operator action. Nothing reads any fresher for it: `hap agents` and
+  `hap status` are pure reads and never woke the daemon in the first place
--- a/changelog.d/fix-roster-publish-connection-pressure.md
+++ b/changelog.d/fix-roster-publish-connection-pressure.md
@@ -12,3 +12,8 @@
   once-a-minute sweep, and every two seconds while a TUI is open — no longer on
   every operator action. Nothing reads any fresher for it: `hap agents` and
   `hap status` are pure reads and never woke the daemon in the first place
+- Fixed an agent's status briefly reverting after it started working. The daemon
+  lists its agents and records that listing a moment later, so a status change
+  arriving in between was overwritten by the older reading — and with no TUI
+  open the next listing is a minute away. An agent that had just gone to work
+  could read as free for that minute, including to `hap task send`
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -285,6 +285,12 @@ type Daemon struct {
 	// the life of the process.
 	rosterPassRunning bool
 
+	// rosterPending is the newest listing waiting to be published (guarded by
+	// mu). Latest-wins: a listing arriving while the pass runs replaces this
+	// rather than queueing, since an older snapshot can only describe a herd
+	// that has moved on.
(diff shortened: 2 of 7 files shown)


0xGosu:
@CharlieHelps please review

PR #402, opened 2026-09-07
#0 feat: push the shared database at once for a remote focus

0xGosu:
## What

A focus filed for **another node's** agent (`f` in the TUI on a remote row) used to wait out the fleet sync loop's 2 s write debounce before the row even left this machine. This nudges the local daemon to push the shared database immediately instead, so the request is in Turso Cloud in time for the owning node's next pull.

## Why the number is what it is

Latency for a remote focus is `~2 s push debounce + 0–15 s until the owning node's next pull + drain`. `focusAgent` itself is a single `FocusPane` shell-out — all of it is transport.

| | avg |
|---|---|
| before | ~9.5 s |
| this PR | ~7.5 s |
| `turso_sync_interval_seconds = 5` on the owning node | ~4.5 s |
| both | ~2.5 s |

**This PR removes the smaller half.** The pull tick lives on the other machine and is only reachable from its own config (`database.turso_sync_interval_seconds`, default 15, min 5, read once at daemon start — a reload does not re-pace the loop). Worth saying out loud so nobody reads this as "focus is fast now".

## Scope

Focus only, as requested. The four awaiting kinds (`rename`, `set_enabled`, `capture`, `deliver_reply`) keep the debounce: each blocks on `AwaitAgentAction`, so the operator is already waiting out the remote's pull *and* the answer's return trip, and 2 s is a rounding error against that. Focus is the one kind that returns immediately, which is what makes its two seconds worth removing.

## The design bit worth reviewing

`control.KindFleetPush` is the one nudge that is **not** about this node's own queues, and it drains none of them — the early return in the nudge handler is the point, not an optimization:

- every local drain is node-scoped, so it could only ever find nothing for a row filed under another node;
- the attention reconcile a `KindReload` carries would re-drive **this** herd's parked episodes.

So without the early return, a keypress that moves somebody else's view could raise an escalation here.

Other bounds:
- `fleetPushNow` is created only when `FleetSync` is wired, mirroring `syncEvents` — under the local engine no loop drains it, and a buffered channel nobody reads would fill on the first nudge and stay full for the process lifetime.
- The push goes through the existing `fleetRun`, so it inherits off-loop execution, the no-cancellation rule and shutdown safety for free.
- A debounce timer armed by the same write is dropped; the reverse order is harmless (one wasted round trip, no correctness question).
- A no-op under the local engine and on daemons predating the kind, which log and ignore it — in both cases the row still travels on the ordinary debounce, so a miss costs latency only.

## Also

Corrects `runRemoteAction`'s doc comment, which claimed the reload nudge "makes THIS node push sooner". It never did — the push is driven by the store's `OnWrite` signal and its debounce, never by a nudge. Left as-is it would invite exactly the wrong follow-up now that `KindFleetPush` exists.

## Tests

`TestFleetPushNudgePushesAheadOfTheDebounce` asserts the push lands *inside* the debounce window, that exactly **one** push happens (proving the armed timer was dropped), and that **zero** local drains ran. `TestRemoteFocusNudgesAFleetPushNotAReload` pins the kind at the call site.

Both non-obvious assertions proved by mutation:
- deleting the early return → `the fleet-push nudge drained the local action queue 1 times`
- deleting the timer clear → `pushes = 2, want exactly one`

Full suite, `go vet`, `gofmt` and `golangci-lint` (0 issues, pinned `GOTOOLCHAIN=go1.26.2`) all green.


0xGosu:
Pushed `c31b698` — remote focus on the **Escalations** tab, and a wrong-agent bug it uncovered.

**The bug:** the Escalations *list* refused a remote agent outright ("focus is local-only"). The *detail overlay* did not refuse — it resolved the bare agent id against this node's agents. An agent id IS a herdr pane id and repeats on every machine, so `f` on a remote escalation for pane `1` moved **this** herdr to its own pane `1`: a different agent, silently, under a success banner.

**The fix:** `agentRowOn(nodeID, agentID)` resolves the **pair** against `agentRows()`. A remote row arrives with the coordinates its own node published — and with `Stale`, so `remoteActionBlocked`'s "that machine has stopped reporting" gate now covers the Escalations path. It was unreachable there before, because a remote agent was refused a step earlier. A vanished agent and a down node report separately: only one is something the operator can go and fix.

**Tasks is unchanged, and that is the right answer.** `agentTaskSourceMatches` returns `nil` for a remote row and `focusTaskGroupAgent` walks local agents only, because `[[task_sources]]` is this node's config and never enters the shared database — a source configured here feeds an agent here. There is nothing remote to focus on that tab by construction.

`TestFocusRefusesARemoteEscalation` is rewritten rather than deleted: its intent (never focus the local namesake) still holds, only the remedy changed. Its local-namesake fixture is what makes the assertion discriminating, and it now asserts the queued action landed on the *other* node carrying that node's tab id. Proved by mutation — a node-blind resolver fails it with `the owning node's queue = []`.

Full suite, vet, gofmt and golangci-lint (0 issues) green.

PR #409, opened 2026-09-07
#0 fix: resolve the audit agent name per node, and name it in `hap audit`

0xGosu:
## The bug

The Audit tab resolved an agent through `Status.AgentName(r.AgentID)` — a map keyed by the agent id **alone**. An agent id IS a herdr pane id, which repeats on every machine sharing the store, and `Store.AuditLog` spans the fleet. So another node's audit row either:

- fell back to a bare pane id (`1`, `w1:p2`), or
- when that pane id collided with a local one, rendered **this node's agent name**, with nothing marking it as another machine's.

The Escalations tab has used the node-aware `Status.EscalationAgent` since the 0.8.0 fleet migration (`1eae7d6`), which wired it into `renderEscalations` and `hap escalations` and left `renderAudit` untouched. This is that oversight, not a deliberate asymmetry.

## The fix

`Status.EscalationAgent` → `Status.RecordAgent`. It always took a `domain.AuditRecord`; only its name was escalation-specific. Every audit-record lookup now goes through it:

| Site | Was |
|---|---|
| `renderAudit` | `AgentName(r.AgentID)` + bare-id fallback |
| `auditDetailLines` | `AgentName(r.AgentID)` |
| `filterAudit` (the `/` search) | `AgentName(r.AgentID)` — `/name` matched a row the column did not show |
| `escalationAlertText` (toast) | `AgentName(...)` |
| `renderEscalations`, `hap escalations` | rename only |

`agentTypeFor` gets the same guard: `MonitoredAgents` is this node's herd, so a remote row now reads `-` rather than borrowing a local agent's type. The audit/escalation detail overlay gains a `Node` line for a remote row, matching what `agentDetailLines` already did.

## `hap audit`

It printed **no agent field at all** — never implemented; the help's column list agreed. It now carries `agent=`, **appended** beside `node=`, so nothing parsing the existing tab-separated fields moves. `agent=` reads `name@node` for another machine's row, exactly as `hap escalations` does.

## Tests

The existing `TestAuditRowsRenderAgentName` uses a single-node status and passes with or without the fix, so both new cases are **two-node with a colliding pane id**: `TestAuditRowFromAnotherNodeIsNotLabelledWithALocalName` (TUI — also asserts the type is not borrowed) and `TestAuditNamesTheAgentAndNeverBorrowsALocalName` (CLI, driving a real second-node store).

Full suite green: `go test -tags "vectors cpu" ./... -count=1`.

https://claude.ai/code/session_01QuzmZBxJZx86FAgdTWjvug

### DIFF
--- a/changelog.d/fix-audit-agent-name.md
+++ b/changelog.d/fix-audit-agent-name.md
@@ -1,0 +1,7 @@
+- Fixed the TUI Audit tab and `hap audit` labelling another machine's rows with a local
+  agent's name: a herdr pane id repeats on every node sharing the store, so audit rows
+  now resolve the same `name@node` identity the Escalations tab already used — and an
+  audit row from another node no longer borrows a local agent's type
+- Added an `agent=` column to `hap audit`, which never had one; it is appended beside
+  `node=` so nothing parsing the existing tab-separated fields moves
+- Added the node to an audit/escalation detail view when the row came from another machine
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -1172,7 +1172,7 @@ func escalations(ctx context.Context, app *frontend.App, out io.Writer, args []s
 	}
 	rules, gradN := ruleIndex(ctx, app)
 	for _, e := range esc {
-		agent := st.EscalationAgent(e)
+		agent := st.RecordAgent(e)
 		rule := "none yet"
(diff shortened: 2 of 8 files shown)


0xGosu:
@CharlieHelps please review

PR #410, opened 2026-09-07
#0 feat!: default new installs to the sqlite task-list backend

0xGosu:
Changes `[task_source_provider] provider` from `local_fs` to `sqlite` — for **new installs only**.

## What a fresh install gets

A checklist is a row in hap's own database (`task_lists`) rather than a markdown file: no file lock behind every read-modify-write, one list derived per matched agent instead of a hand-written path, and — under the turso engine — lists visible fleet-wide. It still makes no outbound call; `github_gist` remains the only provider that does.

- `hap config task-source add --agent x` no longer needs the positional `<checklist.md>`.
- A filesystem path is now **refused** under the default, and the message names both ways out: `--provider local_fs` for a file on disk, or omit the path.
- Locators render as `db_list=` rather than `path=`.

## Existing installs do not move

`Default()` names `sqlite`; `Load` pins `local_fs` the moment the config **file** exists — **before the decode**, so an explicit key still wins and *every* return carries the pin. That covers three exits, not two: the read error, the parse error, and the `AutoAccept.validate()` branch that keeps the partially decoded config rather than rebuilding it. A post-decode pin misses that third one silently — a typo in the config would move an install's storage backend.

`ResolveProvider`'s literal fallback stays `local_fs`: it is reached only by a `Config` assembled in memory, where the posture needing no store handle and no node id is the right answer.

## `AnyNonDefaultProvider`

It gates every "print the provider" branch in the CLI and TUI, and it could no longer be written as "anything but local_fs" — there are now **two** postures an operator reaches without touching the setting, and neither may grow a provider column. It now asks whether the config is **mixed**, and otherwise stays quiet on either local backend. `github_gist` is still always notable.

## Tests — and why `testApp` was left alone

88 `cli`/`frontend` tests failed on the flip. Every one of them means "a file on disk", so they now **declare** it (`localFSApp` / `localFSCfg` / `seedLocalFSConfigIn`), plus ~50 in `tui`.

`testApp` deliberately stays on the **default**. Seeding `local_fs` there would have made all 88 green by declaring that no CLI or frontend test exercises the new default — which is the failure `CLAUDE.md` names three times over ("every frontend test before it ran on a local file, which is the single reason all of this shipped green").

New coverage for the default path instead:

- `TestFreshInstallAddsAPathlessSourceAndRoundTripsItsTasks` — pathless add, list created on demand, full `hap task` round trip, inheritance left unmaterialized.
- `TestFreshInstallRefusalNamesTheWayOut`.
- `TestTaskSourceListIsQuietOnTheSQLiteDefault` — the other half of the byte-identical-output promise.
- A `sqlite` subtest of `TestBootstrapWritesWhereItRegisters` — the generated-task bootstrap is where the locator-is-not-a-path class historically bit; it asserts no file is written, the registered source is a store name, and the tasks read back through the store.
- `TestAFreshConfigDefaultsToSQLite` / `TestAnExistingConfigWithoutTheKeyStaysLocalFS` (writes a real file — a config built in memory cannot tell "no file yet" from "a file omitting the key", so it is the only shape that discriminates; covers all three exits) / `TestAZeroConfigStillResolvesLocally`.

## Docs

README, `sample/config.toml`, the in-binary `hap` skill, the architecture doc's NFR-007, the `privacy` package doc, `CLAUDE.md`, and `hap config task-source add --help` — whose `--provider` enum was already stale (it never listed `sqlite`).

## Release

Minor, so per `CLAUDE.md` the version is hand-written into `herdr-plugin.toml` (`0.9.0`) and `scripts/assemble-changelog.sh 0.9.0` was run in this PR.

**If another PR merges before this one**, rebase and re-run `bash scripts/assemble-changelog.sh 0.9.0` so its fragment is folded in too — the release refuses to tag while unassembled fragments remain.

## Verification

`go build`, `go vet`, `go test -tags "vectors cpu" ./... -count=1` and `golangci-lint run --build-tags "vectors,cpu"` (0 issues). The only failing test is `TestALockedMigrationStepLosesTheLeaseAndFailsClosed` in `internal/store/turso`, which fails identically on an untouched branch in this environment.

https://claude.ai/code/session_01QuzmZBxJZx86FAgdTWjvug

### DIFF
--- a/.claude/skills/hap/SKILL.md
+++ b/.claude/skills/hap/SKILL.md
@@ -534,7 +534,7 @@ tab-separated stdout is unaffected.
 | `tui.disable_check_for_update` | false | turn off the GitHub release check (TUI only, at most every 6h) |
 | `tui.max_instances` | 1 | how many `hap tui` processes may run; starting one closes the oldest past this cap. `0` = no limit |
 | `cli.ai_agent_friendly_output` | true | append the "Next steps" footer to command output |
-| `task_source_provider.provider` | `local_fs` | default storage for every task list: `local_fs`, `github_gist`, or `sqlite` (inside hap's database; syncs under the turso engine) |
+| `task_source_provider.provider` | `sqlite` | default storage for every task list: `sqlite` (inside hap's database; syncs under the turso engine), `local_fs` (a markdown file on disk — the only provider where `path` is a filesystem path), or `github_gist`. The default applies to a config file that does not exist yet; an install that already has one is pinned to `local_fs`. |
 | `task_source_provider.env_file` | (none) | file holding `GITHUB_TOKEN` for `github_gist`; read at use time |
 | `task_source_provider.timeout_seconds` | 20 | per remote store call |
 | `task_source_provider.refresh_seconds` | 30 | how long a remote list is cached |
--- a/.claude/skills/hap/SKILL.md
+++ b/.claude/skills/hap/SKILL.md
@@ -665,15 +665,24 @@ Deprecated but still loading, migrated on the next config save:
 
 ## task sources
 
-A task source points an agent at a checklist file so idle agents get the next
-unchecked item. `hap config task-source` manages **which file**; `hap task`
+A task source points an agent at a checklist so idle agents get the next
+unchecked item. `hap config task-source` manages **which list**; `hap task`
 manages the **items inside it**.
+
+Under the default `sqlite` provider the checklist argument is OPTIONAL: leave it
+out and each matched agent gets its own list inside hap's database. Give it only
+to share ONE list across the agents a source matches — and then it is a list
+NAME, never a filesystem path. For a markdown file on disk, pass
+`--provider local_fs` (an install that predates the default is pinned there, and
+its sources keep taking a path).
 
 ```bash
 hap config task-source list                                   # every source, with its index
-hap config task-source add --agent backend-dev ./docs/tasks.md
-hap config task-source add --workspace "codex-*" ./docs/tasks.md   # "*" wildcards
-hap config task-source add ./docs/tasks.md                    # any agent, any workspace
+hap config task-source add --agent backend-dev                # its own list, derived per agent
+hap config task-source add --workspace "codex-*"              # "*" wildcards
+hap config task-source add --agent backend-dev shared.md      # share ONE list by name
+hap config task-source add --agent legacy-fox --provider local_fs ./docs/tasks.md
+hap config task-source add                                    # any agent, any workspace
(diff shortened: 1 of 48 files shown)


0xGosu:
@CharlieHelps please review

0xGosu:
Both findings addressed in e9251f8 — thanks, they were real.

**1. The TUI could not express the pathless SQLite source.** Correct, and worse than it looked: the prompt took the checklist as its **first positional field**, so on a fresh install there was nothing to type in its place — an empty input was refused for having no path, and an actual filesystem path is refused by the sqlite provider.

`addTaskSourcePrompt` now takes `--agent A`, `--workspace W` and `--provider P`, mirroring the CLI's own flag spellings (both `--flag value` and `--flag=value`, in any position, as the existing flags already were). Positionals still work, and a flag and a positional naming the same field is **refused** rather than silently resolved — whichever one lost would be a selector the operator believes this source carries.

Whether the checklist is required now follows the same rule as the CLI: `cfg.ResolveProvider(...).Remote()`, so it stays **required under `local_fs`**, where there is nothing to derive. This is a provider-conditional relaxation, not a blanket "path is optional now".

I also added `--provider`, which you didn't ask for but is the symmetric hole this PR opened: without it the TUI could no longer create a **file-backed** source at all on a fresh install.

New coverage against a config file that does not exist (the existing `sourcePromptModel` seeds `local_fs`, which is exactly why it couldn't see any of this): `TestTUIAddPathlessSourceOnAFreshInstall` (asserts the inheritance is left unmaterialized, too), `TestTUIAddLocalFileSourceOnAFreshInstall` (the refusal names `local_fs`, then `--provider local_fs` succeeds), `TestTUIPathlessAddIsStillRefusedUnderLocalFS`.

**2. Guidance emitted commands that fail under the new default.** Fixed at every site you listed — `internal/cli/cli.go:1999,2241`, `internal/frontend/notasksource.go:81`, `internal/frontend/frontend.go:3942`, `internal/cli/help.go:786,1001,1170`, `.claude/skills/hap/SKILL.md` — plus two you didn't: `README.md:450` and the SKILL's "where task lists are stored" paragraph, which still said a checklist is a file by default.

Generic guidance is now the pathless form, with `--provider local_fs ./docs/tasks.md` shown beside it for the filesystem case. The `hap task --path` examples are untouched, as you asked — those really do read a local file.

Changelog re-assembled with two entries for the above. `go test -tags "vectors cpu" ./... -count=1` and `golangci-lint --build-tags "vectors,cpu"` are both clean (the earlier lint failure was the 504 you spotted).

PR #424, opened 2026-09-09
#0 fix: repair a wedged fleet sync, and say when this node is isolated

0xGosu:
## The incident

A two-machine turso fleet appeared **isolated**: each TUI showed only its own escalations and agents despite both nodes pointing at the same Turso Cloud URL. The read path was never the problem — `store.PendingEscalations` has no `node_id` filter and neither does `LiveRoster`.

`hap status` on the Mac had the answer, and it was the only place that did:

```
pending escalations: 6 (6 on this node)
other nodes:         1 (1 stale, 4 agents)
fleet sync:          turso — DEGRADED: turso: error: sync engine operation failed:
                     database sync engine error: Post ".../v2/pipeline": tls: failed to
                     verify certificate: SecPolicyCreateSSL error: 0
                     (last pull never, last push never, 0 unpushed)
```

That string comes from **Go's standard library** — `crypto/x509/internal/macos/security.go:71` — and `error: 0` is not a status code, it is the NULL return echoed back. It means `SecPolicyCreateSSL()` returned NULL: a Security-framework allocation failure that says nothing about the certificate. The tursogo SDK builds its own `http.Client` with no `TLSClientConfig` (`driver_sync.go:193`), so it lands on macOS's platform verifier with no way to inject roots.

`hap daemon --restart` cleared it immediately — the evidence base for everything below.

## Why it was invisible

`fleetPush`/`fleetPull` were `logging.Guard` + one `Warn` + `return nil`. `FleetSyncLine` was computed in `frontend.AssessDaemonHealth` and then **never reached `Severity()` or `Banner()`**. So a node could be cut off for hours while the daemon looked healthy and the herd looked quiet.

## What this changes

**Report** — failures now count across *both* directions (a node that pulls fine and cannot push is just as cut off) and isolation is measured from the last **success**, falling back to `FirstFailureAt` — the only anchor a daemon reporting `last pull never, last push never` has.

| State | Severity | Banner |
|---|---|---|
| one failed tick | `DaemonWarn` | `⚠ fleet sync failing…` |
| >5 min no successful sync | **`DaemonError`** | `⚠ FLEET SYNC ISOLATED for 42m — this machine is NOT exchanging rows with the other nodes; their escalations and agents are not shown here and this node's are not reaching them` |
| bootstrap stuck >5 min | **`DaemonError`** | `⚠ DAEMON NOT MONITORING — still waiting 25m for the shared database to bootstrap…` |

It is a clock, not a boolean on `LastError`: a banner that fires on the first blip is a banner nobody reads by Friday. The banner leads with the **consequence**; the TLS text and the evidence are detail lines.

**Repair** — after 5 consecutive failures spanning ≥5 minutes, and only for a fault a fresh process is known to clear, the daemon orders its own restart. Six gates, each closing something the others do not:

- a wired seam + a sync engine at all (nil under sqlite and in every front end)
- not handing off to a replacement **binary** (a newer build outranks a same-binary restart)
- not already ordered
- ≥5 consecutive failures **and** ≥5 min isolated — a count alone fires on a fast interval, a time alone on one old failure
- `domain.SyncFailureProcessLocal` — **unknown means NO**
- the recovery latch is free

**Instrumentation** — `internal/fdprobe` records the descriptor budget at each failure. `RLIMIT_NOFILE` is deliberately **not** raised: exhaustion was the leading hypothesis and was never measured, so raising it would be a fix for an unconfirmed cause that also hides the real one.

## Three defects caught during the build

1. **`--ensure` would have done nothing.** The first version reused the upgrade handoff's spawn. `daemonlock.EnsureFresh` returns doing nothing when the running holder already matches the version and path it would start — precisely a restart as the same binary. The successor would bow out, this daemon would step aside behind it, and the herd would be left **unmonitored**. Now `--restart` (force), and the daemon *latches and keeps running* rather than taking `handedOff`, because the spawned command is what stops it. Pinned by `TestSyncRecoveryOrdersARestartNotAnEnsure`.

2. **The unbootstrapped state was invisible.** `daemonlock.Acquire` runs *before* `openTurso`, whose retry is unbounded — so a wrong URL or rejected token leaves `hap status` reporting a running daemon that has never begun monitoring anything. Exactly the failure class this PR is about.

3. **Fixing (2) introduced a contradiction.** With `FirstFailureAt` populated, an unbootstrapped node also read as `Isolated`, and `hap status` (a bare `if`, not an ordered switch) would tell an operator with no database that their peers' escalations weren't reaching them — pointing at the sync engine when the answer is the URL and token. `Isolated()` now requires `Bootstrapped`.

## Validation

- `go build` / `gofmt` / `go vet` / `golangci-lint --build-tags "vectors,cpu"` — **0 issues**
- Full suite green except one **pre-existing, load-dependent** flake, proven not to be from this branch by running both under identical concurrent load, 5 iterations each:

  ```
  main:   --- FAIL: TestALockedMigrationStepLosesTheLeaseAndFailsClosed
                    A's migration = <nil>, want ErrSchemaLeaseLost
  branch: same failure
  ```

  `internal/store/turso/schema_lease_test.go` is untouched since 0.8.0 (`1eae7d6`); this diff touches **zero files** under `internal/store`. Its 1500 ms lease TTL races a `lockingOwner` holding the real write lock, with no ordering enforced between "A re-proves and finishes" and "B acquires".

- **Five gates verified by mutation** (each fails its test when removed): the process-local check, `Isolated` vs `Degraded`, the retry throttle, the `Bootstrapped` requirement, and the recovery latch.

Every new invariant has a paired control test — `TestARemoteOutageNeverRestartsTheDaemon`, `TestAFreshSyncFailureIsOnlyAWarning`, `TestAColdBootstrapIsOnlyAWarning` — because each of them passes on code that answers one way for everything.

## Not addressed

The macOS root cause itself. `SSL_CERT_FILE` bypasses the platform verifier only from Go 1.27 (`go.mod` says `go 1.25.0`); the alternative is `x509.SetFallbackRoots` + `GODEBUG=x509usefallbackroots=1`. Worth a follow-up once the fd evidence from this PR says whether exhaustion is the trigger.

https://claude.ai/code/session_0129RvYcSKmb1isHsQ78SbY9


0xGosu:
@CharlieHelps please review

PR #428, opened 2026-09-10
#396 fix: make the integration confirm-delivery and task-summary cases assert again

### LINKED ISSUE #396: Integration: the confirm-delivery tests cannot pass since 0.8.0, silently disabling the send-content guard
## Summary

`TestRealConfirmDeliversMenuDigit` and `TestRealConfirmDeliversRemoteEnvSelection` in `test/integration/` **cannot pass as written** and have been failing since v0.8.0. They are not flaky — they fail deterministically, on `main`, on every run.

This matters more than a dead test usually would: `CLAUDE.md` names `TestRealConfirmDeliversMenuDigit` as the guard on **the send-content regression** — that confirming a label reply selects the *numbered menu digit* rather than pasting the label. That is the regression where, verified live, `agent prompt "2"` pastes the `2` as text and its Enter commits whichever option the caret was on: it answered "Apple" while hap had chosen "Banana", silently, with a success exit code. That guard has been off since 0.8.0.

## Reproduce

```sh
go test -tags integration ./test/integration/ -run 'TestRealConfirmDelivers' -v
```

```
=== RUN   TestRealConfirmDeliversMenuDigit
    integration_test.go:287: no healthy hap daemon is running; start one with `hap daemon --ensure`
--- FAIL: TestRealConfirmDeliversMenuDigit (0.34s)
=== RUN   TestRealConfirmDeliversRemoteEnvSelection
    integration_test.go:388: no healthy hap daemon is running; start one with `hap daemon --ensure`
--- FAIL: TestRealConfirmDeliversRemoteEnvSelection (0.17s)
```

Reproduced on `main` (`71b05fb`) with a healthy daemon actually running (`hap status` → `daemon: running v0.8.2`), so the message is misleading — see below.

## Cause

Two layers, and the second is why the obvious fix is not enough.

**1. The App has no `DaemonInfo`.** Both tests build:

```go
app := &frontend.App{Store: st, Herdr: cli, Author: "itest"}
```

`integration_test.go:275` and `:376`. `App.AssessDaemonHealth` (`internal/frontend/daemonhealth.go:84`) derives `h.Running` from `a.DaemonInfo`, which is nil here — so `h.Running` stays false whatever daemon is actually up, and `requireLiveDaemon` (`internal/frontend/agentactions.go:56`) returns `ErrDaemonUnavailable` before anything is delivered.

**2. Even with `DaemonInfo` wired, the delivery could not happen.** `1eae7d6` (Turso central database, 0.8.0) changed `App.Confirm` to *queue an `agent_actions` row for the daemon to drain* rather than sending inline, and added the `requireLiveDaemonFor` gate in front of it. But both tests open their own throwaway store:

```go
st, err := store.Open(filepath.Join(t.TempDir(), "hap.db"))
```

The real daemon reads its own state-dir store, so an action queued into a temp store is never drained. Satisfying the gate alone would move the failure from "no healthy daemon" to a 5-second timeout on the marker file.

These tests were last touched in `b48000c`, before `1eae7d6` — they were simply never revisited when the delivery path became asynchronous.

## Blast radius is three tests, not two

`TestRealClaudeConsult` (`integration_test.go:546`) has the identical `frontend.App{...}` construction and also calls `app.Confirm`. It currently **skips** unless `HAP_ITEST_CLAUDE=1`, so it hides the same breakage — anyone enabling the real-Claude cases to check something else will hit it with no clue why.

## Suggested direction

Not obvious which is right, so leaving it open:

- Point the test App at the **daemon's real store and state dir** (and set `DaemonInfo`), so the running daemon actually drains the queued action. Closest to what the test is trying to prove — it exercises the true production path end to end — but it writes to the operator's live store.
- Or have the test **run its own daemon** against the temp store, as `task_summary_test.go` already does for its consult cases. Isolated, but more harness.
- A `frontend.App` built without `DaemonInfo` silently meaning "no daemon" is itself a sharp edge worth a second look, given the failure mode is a misleading error rather than a compile-time or obvious runtime complaint.

Whatever the fix, it is worth adding something that fails when these tests stop *actually asserting delivery*, since the current failure mode is one that could just as easily have been "quietly skipped" instead of "loudly failed".

## Environment

- `main` @ `71b05fb`, and on `feat/reduce-db-writes-and-retention` (#395) — unrelated to that change
- herdr 0.7.x, `HERDR_ENV=1`, Linux
- Found while running the integration suite before #395 per `CLAUDE.md`'s pre-PR recommendation

### LINKED ISSUE #397: Integration: task_summary_test.go flakes in full-suite runs — scratch pane classified before it paints
## Summary

The two tests in `test/integration/task_summary_test.go` — `TestRealConsultContextTaskSourceSummary` and `TestRealIdleUnlearnedSignatureConsultsInsteadOfReviewing` — flake when the integration suite runs as a whole. Each **passes reliably in isolation**.

Observed on `main` (`71b05fb`): **2 of 3 full-suite runs failed**. Which of the pair loses varies between runs, so it is not a fixed ordering dependency.

```
--- main run 1 ---
--- PASS: TestRealConsultContextTaskSourceSummary (0.70s)
--- FAIL: TestRealIdleUnlearnedSignatureConsultsInsteadOfReviewing (16.49s)
--- main run 2 ---
--- PASS: TestRealConsultContextTaskSourceSummary (0.78s)
--- FAIL: TestRealIdleUnlearnedSignatureConsultsInsteadOfReviewing (16.39s)
--- main run 3 ---
--- PASS: TestRealConsultContextTaskSourceSummary (0.51s)
--- PASS: TestRealIdleUnlearnedSignatureConsultsInsteadOfReviewing (1.46s)
```

In isolation, either test passes in well under a second:

```sh
go test -tags integration ./test/integration/ -run TestRealIdleUnlearnedSignatureConsultsInsteadOfReviewing -v
# --- PASS (1.46s)
```

The ~16s duration is `waitForConsult`'s 15-second deadline (`task_summary_test.go:158`) expiring:

```
task_summary_test.go:211: no LLM consult captured for the scratch pane within timeout
```

## Cause

The scratch pane is being **classified before its shell has painted**, so it never becomes a consult at all.

Daemon logs from a failing run — the scratch pane is `w1:p2K`:

```
INFO reconcile: re-driving parked agent agent=w1:p2K pane=w1:p2K status=blocked
INFO escalated agent=w1:p2K situation=unclassifiable reason=unclassifiable
```

versus the same test passing, where the scratch pane is `w1:p2M`:

```
INFO escalated agent=w1:p2M situation=idle reason=llm_no_submit
```

`llm_no_submit` means the consult ran and the stub LLM declined to submit — the expected path. `unclassifiable` means the capture read an empty or half-painted pane, so `Decide` never reached the consult and `waitForConsult` waits out its full deadline for something that will never arrive.

The synchronisation is a fixed sleep (`task_summary_test.go:258`):

```go
pane := startIdleAgent(t)
// Give the scratch shell a moment to print its line before the daemon's
// classification read.
time.Sleep(1 * time.Second)
```

against a 50ms capture delay (`tinyCaptureDelayTOML`). One second is enough on an idle machine and not enough under full-suite load — which is exactly the difference between the isolated and whole-suite runs.

Load is plausibly higher than it looks: `newTestDaemon` wires a **real** `herdr.CLI`, so each test's daemon reconciles every real pane on the shared herdr instance, not just its own scratch pane. The failing run shows it escalating two unrelated panes in the same pass:

```
INFO escalated agent=w1:p1K situation=idle reason=no_task_source
INFO escalated agent=w2:p7 situation=idle reason=no_task_source
```

`runDaemon`'s own comment already acknowledges this ("a slower reconcile-driven goroutine (processing the OTHER real panes on the shared herdr instance)"), and per `CLAUDE.md` adding a pane also makes the events subscriber reconnect with a 1s backoff — so an earlier test's scratch panes lingering in the herd add both work and a resubscribe window to the next test's startup. I have not instrumented which of those dominates; the empty-capture symptom is the part I can evidence directly.

## Suggested direction

The sleep is the fragile part, so the fix is probably to stop timing it:

- **Poll the pane for its expected content** before injecting the transition, rather than sleeping a fixed 1s — the same shape `waitForConsult` already uses for its own wait.
- Or have `startIdleAgent`/`startMenuAgent` block until the scratch script's marker line is actually on screen (`pane read`), so the helper returns a pane that is ready by construction.

Either removes the wall-clock assumption entirely. Raising the sleep would only move the threshold.

Worth noting that the failure is at least *loud* — it fails rather than skipping — but it costs 15s per occurrence and, since it reads as "the daemon never consulted", it invites suspicion of the decision pipeline rather than of the harness.

## Environment

- `main` @ `71b05fb`; also seen on `feat/reduce-db-writes-and-retention` (#395), which does not touch this path
- herdr 0.7.x, `HERDR_ENV=1`, Linux
- Found while running the integration suite before #395 per `CLAUDE.md`'s pre-PR recommendation
- Unrelated to #396, which is a separate deterministic failure in the same suite

0xGosu:
Closes #396. Closes #397.

Both issues verified reproducible on `main`, then fixed. Both are test-only — no production code changes.

## #396 — legit, exactly as reported

Deterministic, not flaky:

```
=== RUN   TestRealConfirmDeliversMenuDigit
    integration_test.go:287: no healthy hap daemon is running; start one with `hap daemon --ensure`
--- FAIL: TestRealConfirmDeliversMenuDigit (1.28s)
--- FAIL: TestRealConfirmDeliversRemoteEnvSelection (1.20s)
```

…with `hap status` reporting `daemon: running v0.9.7`. Both layers the issue names are real:

1. `&frontend.App{Store: st, Herdr: cli, Author: "itest"}` sets no `DaemonInfo`, and `AssessDaemonHealth` derives `Running` from exactly that field — so `h.Running` is false whatever daemon is up and `requireLiveDaemonFor` refuses before anything is written.
2. Past that gate it still could not deliver: since `1eae7d6` (0.8.0) `App.Confirm` **queues** an `agent_actions` row for the owning node's daemon, and the tests opened a throwaway store nothing drains.

So the guard on the send-content regression has been off since 0.8.0 — two minor releases. `TestRealClaudeConsult` carried the identical construction behind `HAP_ITEST_CLAUDE=1` and is fixed with them, as the issue's blast-radius section asked.

**Fix — option 2 from the issue.** Each case runs its own daemon over its own store (`testDaemon.App`), never the operator's live one: under the turso engine that would push the suite's scratch rows to their cloud database. Three fields are load-bearing (the daemon's store, a `DaemonInfo`, the control socket for the wake nudge) and `StateDir` is deliberately left empty so the health-derived refusals are never weighed against a heartbeat this in-process daemon may not have written yet.

The App is left with **no herdr adapter at all**, and that is the assertion rather than an omission: a confirm that lands the keystrokes with no adapter in hand is proof the delivery really went through the daemon. Confirmed in the logs — `INFO agent action done action=1 kind=deliver_reply`.

The issue's third bullet (a nil `DaemonInfo` silently meaning "no daemon" is a sharp edge) is answered with a CLAUDE.md rule naming the trap rather than a code change; happy to take a different call on that.

## #397 — the flake is legit, but the reported cause was not what kept it red

The 1s sleep is a real wall-clock assumption and is gone (replaced by polling the pane's own content, the shape `waitForConsult` already uses). But it was not sufficient — with the sleep replaced, both cases still failed, and the daemon logs say why:

```
INFO escalated agent=w1:p5V situation=unclassifiable reason=unclassifiable
INFO escalated agent=w1:p5W situation=idle      reason=over_masked
```

Both are **content** failures, not paint races. Probing herdr directly (0.8.2):

| pane | `--source recent` |
|---|---|
| scratch pane, 5 lines of output, quiescent | **empty** |
| same pane while 200 lines stream past the viewport | content |
| real idle claude pane `w2:p7` | 7727 bytes |

`pane read --source recent` — the source the **daemon** classifies from — is a scrollback-shaped delta: it returns nothing for a pane whose whole output still fits on screen. Every real agent pane has scrolled, which is why production never meets this; a five-line scratch script had not, so the daemon captured nothing and escalated instead of consulting. That also explains why the symptom read as a broken decision pipeline.

**Fix.** `fillViewportSh` pushes the fixture past the pane's own `tput lines`, so it does not silently depend on how tall the operator's terminal is — the same class of hidden environmental assumption as the sleep. Filler is digit-free on purpose: the classifier masks numbers to placeholders, and a salient made mostly of placeholders trips the over-masking floor, which is the same failure by the other door.

## Verification

3 consecutive **uncached** full-suite runs, all green:

```
--- PASS: TestRealPaneInfo (0.13s)
--- PASS: TestRealConfirmDeliversMenuDigit (0.75s)
--- PASS: TestRealConfirmDeliversRemoteEnvSelection (1.12s)
--- PASS: TestRealHerdrNotification (0.12s)
--- PASS: TestRealHerdrNotificationSecondCallSucceeds (0.07s)
--- PASS: TestRealHerdrNotificationRejectsEmptyTitle (0.04s)
--- PASS: TestRealInHerdrDetection (0.00s)
--- PASS: TestRealConsultContextTaskSourceSummary (1.10s)
--- PASS: TestRealIdleUnlearnedSignatureConsultsInsteadOfReviewing (0.65s)
ok  github.com/0xGosu/herdr-auto-pilot/test/integration  3.995s
```

Both task-summary cases now escalate for the right reason (`reason=llm_no_submit` — the consult ran) rather than timing out at 15s. `HAP_ITEST_CLAUDE=1` cases still skip and were not exercised.

Full gate green under `GOTOOLCHAIN=go1.26.2`: `go test -tags "vectors cpu" ./... -count=1`, `gofmt`, `go vet`, `golangci-lint run --build-tags "vectors,cpu"` (0 issues).

## Two things worth knowing, neither changed here

- **Every integration run leaks scratch panes into the operator's live herd**, and the real daemon then adopts and escalates them (`pane_not_found` / `unclassifiable` rows in `hap audit`). The suite's own daemons also reconcile every real pane on the shared herdr instance — pre-existing, and `runDaemon`'s comment already acknowledges it. Safe today only because a fresh temp store has no graduated rules, so those panes can only ever be escalated, never sent to.
- **The `--source recent` semantics above are worth confirming against herdr's intent.** Production is fine (real panes scroll), so nothing is changed here, but if herdr 0.8.x narrowed that source, hap's classification read is closer to the edge than `min_herdr_version = 0.7.0` suggests. Happy to file that separately.


0xGosu:
@CharlieHelps please review

PR #430, opened 2026-09-10
#421 feat: add `signatures search --screen` to search the captured pane

### LINKED ISSUE #421: signatures search: add --screen to search the captured pane, not the masked salient
A learned rule's captured screen is stored in full and unmasked, but nothing can search it. `hap signatures search` matches over `signature_embeddings.salient` — the **masked** text — so every literal path, version, hash, command argument and numeric option has already been replaced by `<path>`/`<num>`/`<hash>` before it reaches the corpus. An operator who remembers the command but not the rule cannot find it.

```console
$ hap signatures search "npm install --force"
no rules match the keyword search "npm install --force"
```

...even when a rule for exactly that approval exists and `hap signatures show` prints the command verbatim.

`signature_snapshots` already holds what's needed — one raw excerpt per signature, written by `SaveSignatureSnapshot` on first sight, with no retention path so nothing ages out. Its only readers today are two single-signature `GetSignatureSnapshot(sig)` lookups for display (`frontend.go:5334`, `5345`); the port offers no scan.

**Proposed surface**

```
hap signatures search <query> [--screen] [--semantic] [--limit N] [--min-score S] [filters]
```

| Mode | Corpus | Matcher |
|---|---|---|
| *(default)* | rule fields + masked salient | case-insensitive substring |
| `--screen` | `signature_snapshots.pane_excerpt` (raw) | case-insensitive AND-of-terms |
| `--semantic` | stored vectors | cosine, `--min-score` floor |

- `--screen --semantic` is refused with an error: snapshots carry no vectors, so accepting it would silently ignore `--screen`.
- Existing `--type` / `--mode` / `--agent-type` / `--min-conf` filters compose unchanged.
- Name: `--screen` over `--raw` (reads as an output format) and `--snapshot` (names an internal table).

**Matching semantics**

`--screen` requires **every term, anywhere in the excerpt** — not the whole query as one substring, which is right for a 35-char masked salient and wrong for a 4000-rune screen where terms are scattered across a menu. The divergence from the default mode is deliberate and must be visible: the summary line reads `N screen match(es)`, and `Details` states it.

A phrase falls out of the shell for free — the permuting parser peels positionals one at a time, so a quoted argument arrives as a single argv entry containing a space:

```console
hap signatures search --screen "npm install" force   # phrase AND term
hap signatures search --screen npm install force     # three independent terms
```

Pre-existing limitation, unchanged: a term starting with `-` hits the flag parser.

**Output**

One tab-separated line per result, as today — these listings are parsed by scripts — plus a trailing `match=` field carrying ~100 runes around the first hit, newlines collapsed to spaces:

```console
$ hap signatures search --screen npm install --force
a1b2c3d4…  approval  claude  autonomous  3/3  conf=high  top="2"  09-01 09:12:44  match="…Bash(npm install --force)  1. Yes  2. Yes, and don't ask…"

1 screen match(es) for "npm install --force" (412 screens searched)

Next:
  hap signatures show a1b2c3d4    the original situation, plus recent decisions
  ...
```

The full excerpt already has a home in `hap signatures show`, which the existing hint points at.

`(412 screens searched)` also prints on the empty result, and earns the line: it separates *"nothing matched"* from *"there is nothing to match against."* Empty-result hints cross-suggest the other two modes, matching the existing `--semantic` hint pattern.

Ordering needs no new code — `App.Signatures` already returns newest-`updated_at` first, and `signatures.updated_at` is stamped on confirm/correct, so it is decision recency rather than rule birth date.

**TUI**

The Rules tab dispatches the identical `App.SearchSignatures` (`tui.go:2570` → `semanticSearchCmd`), so `--screen` lands as a third search mode for near-free. The tab already renders `SignatureSnapshot` for the selected rule, so searching that same text is the obvious completion.

**Where it lands**

- `internal/store` — the scan. Only new SQL; must work through `sqlbridge` and under all three `HAP_STORE_TEST_MODE` values. No node-scoping exemption needed: `signature_snapshots` has no `node_id` by design, since a signature is a fleet-unique content hash.
- `internal/ports` — one read method on `ReadStore`.
- `internal/frontend/signaturesearch.go` — `SignatureSearchOpts.Screen`, a `screenSearch` beside `keywordSearch`, `SignatureSearchResult.Match`.
- `internal/cli` — flag, refusal, `match=` column, cross-mode hints.
- `internal/cli/help.go` — `Usage`, one `FlagDoc`, `Details` covering AND-of-terms and quoted phrases, one `Examples` entry.
- `internal/tui` — mode toggle on the Rules tab.

**Follow-up, out of scope here**

The motivating case is the consulted LLM finding what this herd already answered for a similar screen. That needs two further changes, both deliberately excluded:

1. `[llm] prompt` currently says *"Use ONLY get_context and submit_decision"*, so the assistant will not run the CLI regardless of having Bash.
2. `hap` is a plugin binary and not necessarily on PATH — such a prompt would need the `{self}` placeholder.

Also excluded: any change to what an LLM *does* with a past `chosen_action`. If that follow-up happens, the history must surface `source` (`operator`/`rule`/`llm`) and `is_correction` — returning a bare action lets an LLM guess become auto-accepted, become a `decisions` row, and become evidence for the next LLM guess with no human in the loop.

**Open questions**

1. Should `--limit` generalize beyond `--semantic`? Keyword search is unbounded today; a screen search over hundreds of rules wants a cap. Generalizing changes existing behavior for script consumers.
2. Snapshots are unmasked and permanent (no retention path). A screen containing a token stays searchable forever. Already true of `hap signatures show`, but a *searchable* corpus is a wider surface — does it warrant a note in the help text, or a redaction pass?

0xGosu:
Closes #421.

A learned rule's salient is masked before it is stored — every literal path, version, hash and number is already `<path>`/`<num>`/`<hash>` — so an operator who remembers the command cannot find the rule that answers it. The raw screen was already sitting in `signature_snapshots`, with no way to scan it.

Verified against a scratch store seeded with a real-shaped approval capture:

```console
$ hap signatures search "npm install --force"
no rules match the keyword search "npm install --force"

$ hap signatures search --screen npm install force
approval:a1b2c3d…	approval	claude	autonomous	3/1	conf=-	top=""	09-10 06:49:46	match="⏺ Bash(npm install --force) ⎿ Running… Do you want to proceed? 1. Yes 2. Yes, and don't as…"

1 screen match(es) for "npm install force" (2 screens searched)
```

## What the shape of the corpus forced

**AND-of-terms, not the whole query as one substring.** A masked salient is ~35 chars, where contiguous is the natural reading. A screen is thousands of runes with the remembered words scattered across a prompt, a menu and a footer — requiring them contiguous there answers "no" for nearly every real query, which is the bug being fixed. The divergence is deliberate and made visible: the summary reads `N screen match(es)` and the help states it.

**A shell-quoted phrase stays a phrase.** This is the part the issue's spec required but its prose didn't account for: `signaturesSearch` joins its peeled argv words into one string before searching, so `--screen "npm install" force` and `--screen npm install force` arrive identical and the documented distinction cannot exist. `SignatureSearchOpts.Terms` carries the argv boundaries through; nil falls back to whitespace-splitting, which is the honest reading of the TUI's single free-text box. `TestScreenSearchPhraseIsNotThreeTerms` is the discriminating case — same query, opposite answers.

**The denominator counts screens READ.** A rule the structured filter dropped, and one learned before snapshots existed, contribute nothing. Counting stored snapshots instead would report a healthy corpus for a search that examined one row, which inverts the signal the line exists for (`nothing matched` vs `nothing to match against`). It prints on the empty result too.

**`--screen --semantic` is refused**, not resolved to one of them: snapshots carry no vectors, so serving the semantic search would silently ignore `--screen` and answer a different question over a different corpus.

## Decisions on the issue's open questions

1. **`--limit` generalizes to `--screen` only.** Keyword search stays unbounded — it predates the flag and scripts parse its output, so bounding it is a behavior change for existing consumers. `--screen` is new, so a cap costs nothing.
2. **Unmasked-and-permanent snapshots get a note in the help, not a redaction pass.** A redaction pass would diverge from what `hap signatures show` already prints for the same row, so the searchable surface and the displayable one would disagree — worse than saying plainly that a secret which appeared on a pane stays searchable.

## Notes for review

- `match=` is **appended** as a trailing field, so the eight columns scripts already parse keep their positions. Its window is cut by **rune**, not byte — captured screens are full of box-drawing glyphs and the `…` truncation marker.
- Ordering is `App.Signatures`' newest-`updated_at`-first, untouched. Ranking by match position instead would bury a rule an operator just taught under an older one that happens to mention the term earlier.
- The TUI's ctrl+g deliberately does **not** share the semantic result set's SEM column: a screen result carries no cosine, and rendering it there prints `0.00` on every row. `semanticSigSearch.screen` gates it.
- The store scan needs no node-scoping and no exemption — `signature_snapshots` has no `node_id` by design, since a signature is a fleet-unique content hash. It is proved under all three `HAP_STORE_TEST_MODE` values (sqlite / proxy / turso).
- `SearchSignatures` keeps its signature as a wrapper; `SearchSignaturesWithStats` is the new one. Both callers (CLI, TUI) are updated and there is no third.

The core tests were proved by mutation: pointing `screenSearch` at the masked-salient corpus, ignoring `opts.Terms`, and counting all stored snapshots each fail exactly the case written for them.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,1782 +1,1039 @@
 # CLAUDE.md
 
-Herd Auto Prompter (**hap**) — a Go plugin for the herdr terminal multiplexer
-that watches every agent pane, auto-answers when a learned rule is confident,
-and escalates to the operator (or a local LLM CLI) when not. `CONTRIBUTING.md`
-has the full ground rules; this file is the day-to-day working reference.
+Herd Auto Prompter (**hap**) — a Go plugin for the herdr terminal multiplexer that
+watches every agent pane, auto-answers when a learned rule is confident, and escalates
+to the operator (or a local LLM CLI) when not. `CONTRIBUTING.md` has the full ground
+rules; this file is the day-to-day working reference.
+
+**How to read the architecture rules.** Each names the identifier that implements it and
+states which way it must fail. The full rationale — mechanism, measured numbers, the
+incident that produced it — is the doc comment on that identifier, which is usually
+richer than what is here; go read it before changing anything the rule covers. What this
+file adds is reach: the hazard is normally in a file you had no reason to open, and the
+**test traps** are the reason each of these regressions shipped green. Find the guarding
+tests with `grep -rn "func Test<Topic>" --include=*_test.go`.
 
 ## Skills (`.claude/skills/`)
 
-Prefer these for how-to detail — this file keeps only what must stay in view.
+Prefer these for how-to detail.
 - **`herdr`** — drive herdr from inside it (workspaces, tabs, panes, agents, waits).
-- **`hap`** — operate the plugin via its CLI `hap`: agent's status, agent's tasks, escalations, config, safety rules, task
-  sources.
+- **`hap`** — operate the plugin via its CLI: status, tasks, escalations, config, safety
+  rules, task sources.
 - **`hap-development-local`** — the local dev loop: link the working tree, rebuild,
-  hot-swap the daemon (`hap daemon --ensure`), and live-test against a real agent.
+  hot-swap the daemon (`hap daemon --ensure`), live-test against a real agent.
 
-The hap skill also ships inside the binary: `hap --skill` prints it, and
-`hap skill install <claude|codex|agents>...` (or the TUI Config-tab shortcut)
-installs it into `~/.claude/skills/hap/`, `~/.codex/skills/hap/`, or
-`~/.agents/skills/hap/`.
+The hap skill also ships in the binary: `hap --skill` prints it,
+`hap skill install <claude|codex|agents>...` (or the TUI Config tab) installs it.
 
 ## Build, test, lint
 
-The semantic matcher links native code (llama.cpp via CGO, FAISS behind bleve's
-`vectors` tag), so **the native deps are needed once** and the `vectors cpu` tags
-always — a build without both fails to link.
+The semantic matcher links native code (llama.cpp via CGO, FAISS behind bleve's `vectors`
+tag), so **the native deps are needed once** and the `vectors cpu` tags always — a build
+without both fails to link.
 
 ```sh
-bash scripts/check-submodule-gitlink.sh        # seconds: submodule must be a gitlink, not a symlink (#265)
+bash scripts/check-submodule-gitlink.sh        # submodule must be a gitlink, not a symlink (#265)
 bash scripts/setup-native.sh                   # one-time: submodules + llama-go libs + FAISS → /usr/local/lib
 go build -tags "vectors cpu" ./...             # CGO; needs a C/C++ toolchain
-go test -tags "vectors cpu" ./... -count=1     # full unit/golden/safety/semantic suite (what CI runs)
+go test -tags "vectors cpu" ./... -count=1     # what CI runs
 gofmt -l . | grep -v submodule && go vet -tags "vectors cpu" ./...
-golangci-lint run --build-tags "vectors,cpu"   # CI runs this too
+golangci-lint run --build-tags "vectors,cpu"
 ```
 
-- The real-model embedder test skips unless `models/all-minilm-l6-v2-q8_0.gguf`
-  exists (download once from the HF repo in `release.yml`, or set `HAP_TEST_EMBED_MODEL`).
+- The real-model embedder test skips unless `models/all-minilm-l6-v2-q8_0.gguf` exists
+  (download once from the HF repo in `release.yml`, or set `HAP_TEST_EMBED_MODEL`).
 - Golden classifier fixtures: `internal/classify/testdata/`; regenerate with
   `UPDATE_GOLDEN=1 go test ./internal/classify/` and review the diff.
 - Run the full suite before every commit that touches Go code.
-- Full pipeline smoke test (fake herdr → real daemon → real LLM CLI):
-  `go build -o /tmp/e2e ./e2e_harness && /tmp/e2e <short-dir> <hap-bin> <config-dir> <state-dir>`,
-  then inspect with `hap audit` / replay `get_context` via `hap mcp`.
-- Iterating on the plugin against a live herdr (link the working tree, rebuild,
-  hot-swap the daemon): see the **`hap-development-local`** skill.
+- Pipeline smoke test (fake herdr → real daemon → real LLM CLI):
+  `go build -o /tmp/e2e ./e2e_harness && /tmp/e2e <short-dir> <hap-bin> <config-dir> <state-dir>`.
 
 ## Local integration suite (real herdr + claude)
 
-`test/integration/` drives an **actual running herdr** (and, when enabled, a
-**real Claude Code CLI**), gated by the `integration` build tag so `go test ./...`
-and CI never run them. Each test **skips** (never fails) when its dependency is
-absent, so these are safe to run anywhere:
+`test/integration/` drives an **actual running herdr** (and, with `HAP_ITEST_CLAUDE=1` /
+`HAP_ITEST_CODEX=1`, a real agent CLI), gated by the `integration` build tag so
+`go test ./...` and CI never run them. Each case **skips** (never fails) when its
+dependency is absent.
 
 ```sh
 go test -tags integration ./test/integration/ -v                    # from inside herdr, or set HERDR_BIN_PATH
-HAP_ITEST_CLAUDE=1 go test -tags integration ./test/integration/ -v -timeout 20m # also drive a real claude (spends tokens; several real-claude cases can exceed the 10m default)
-go test -tags "integration vectors cpu" ./test/integration/ -v      # include the real-model semantic case
+HAP_ITEST_CLAUDE=1 go test -tags integration ./test/integration/ -v -timeout 20m  # spends tokens
+go test -tags "integration vectors cpu" ./test/integration/ -v      # + the real-model semantic case
 ```
 
-- Loads `test/integration/testdata/config.toml` (the Claude Code recipe) — edit it
-  to match the CLI you want to exercise.
-- **Anything asserting a CONFIRM runs its own daemon, and an App with no `DaemonInfo` is
-  the trap.** Since 0.8.0 `App.Confirm` queues an `agent_actions` row for the owning node's
-  daemon instead of sending inline, so the obvious construction —
-  `&frontend.App{Store: st, Herdr: cli}` over a throwaway store — cannot deliver anything:
-  `AssessDaemonHealth` derives `Running` from `DaemonInfo`, nil reads as "no daemon"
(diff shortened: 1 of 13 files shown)


0xGosu:
### Local test-suite stability — findings so far (not yet settled)

Recording this openly because my first read was too generous and the numbers moved against it.

Full `go test -tags "vectors cpu" ./... -count=1` in this devcontainer:

| Run | Branch | Result |
|---|---|---|
| 1 | this PR | `TestALockedMigrationStepLosesTheLeaseAndFailsClosed` (turso) |
| 2 | this PR | that + `TestFSPImmediateAcceptOnEscalate` (daemon) |
| 3 | this PR | that + `TestAutoActConfidenceThresholdGate` (daemon) |
| A | `main`, unmodified | green (39 pkgs) |
| B | `main`, unmodified | green (39 pkgs) |

So `main` is 2/2 green and this branch is 0/3. I initially called it pre-existing flakiness on the strength of one green baseline; three-versus-two is not enough to keep saying that, and the turso lease case failing on **every** branch run is a consistent difference that deserves an explanation rather than a shrug.

**What is established:**

- Every failing test passes in isolation on this branch (`-count=5`, and `-race` clean for the daemon one).
- The daemon test is a **different one each run** — the signature of a race, not a regression.
- This branch touches **no file** under `internal/daemon` or `internal/store/turso` (`git diff --name-only main...HEAD` confirms).
- `internal/store/turso` depends on `internal/ports` at compile time only, and never references the new `ListSignatureSnapshots`. There is no logical channel from this diff to that test.
- The lease test tunes sub-second constants (1500ms TTL, 300ms renew/poll) against a real `tursodb` subprocess over TCP, and the assertion that fails is A's migration *completing* before B can take the lease — i.e. it fails when the machine is slow enough to delay B, not when logic changed.

**Working hypothesis:** this PR adds ~450 lines of tests across `store`, `frontend`, `cli` and `tui`, which lengthens those packages and shifts which packages are co-scheduled with the timing-sensitive ones under `go test ./...`. That would make it a pre-existing fragility this PR *exposes* rather than causes — but I have not proved it yet.

**In flight:** the full suite on this branch with `-p 1` (packages serialized), and CI's own test jobs on ubuntu/macos, which are the better signal since they are clean machines. I will post both results here. If the serialized run is green and CI is green, that confirms load-sensitivity and the lease test's timing constants are worth a separate issue. If either is red, the problem is mine and I will fix it before asking for a merge.

0xGosu:
### Resolved: load-sensitivity, not a regression

Both experiments came back, and together with CI they settle it.

| Run | Config | Result |
|---|---|---|
| this PR × 3 | `./...` parallel, local devcontainer | red (turso lease every time; a *different* daemon test 2 of 3) |
| `main` × 2 | `./...` parallel, local devcontainer | green |
| **this PR** | **`./... -p 1`** (packages serialized) | **green, 39 pkgs** |
| **this PR** | **CI ubuntu-latest** | **green (9m37s)** |
| **this PR** | **CI macos-14** | **green (8m6s)** |

Serializing the packages turns this branch green on the same machine that had failed it three times running, and both CI platforms pass the identical suite. Nothing about the code changed between those runs — only how many test binaries were competing for the box.

That matches the mechanism: this PR adds ~450 lines of tests across `store`, `frontend`, `cli` and `tui`, lengthening four packages and shifting which ones are co-scheduled with the timing-sensitive ones. `internal/store/turso` is reachable from this diff only as a compile-time `internal/ports` dependency and never calls the new `ListSignatureSnapshots`, so there is no path by which the code could change that test's outcome — only the clock can.

**Conclusion:** pre-existing timing fragility that this PR's extra test weight makes easier to hit locally. Not introduced here, and not visible in CI.

**Worth a separate issue** (happy to file it): `TestALockedMigrationStepLosesTheLeaseAndFailsClosed` tunes `schemaLeaseTTL=1500ms / renew=300ms / poll=300ms` against a real `tursodb` subprocess over TCP, and asserts that A's migration loses the lease to B. Under load B's takeover is delayed and A simply finishes first, so the test fails in the *safe* direction — but it fails. The daemon-side ones (`TestFSPImmediateAcceptOnEscalate`, `TestAutoActConfidenceThresholdGate`, a different one each run) look like a genuine race worth its own look; `-race` was clean on the ones I could reproduce in isolation.

0xGosu:
@CharlieHelps please review

0xGosu:
Code review (`/code-review low --fix`): one pass over the non-test hunks, **no findings** — nothing to fix, working tree unchanged.

Filed #431 for the load-sensitive test failures described above. One correction to what I wrote earlier: I called the turso case a timing artifact, but the failing assertion is that **A completed its migration while B held the lease**, and A's own renewal logs the takeover before proceeding. That is either a fixture racing itself or a real gap in the between-step re-prove — I could not tell which from the output, so #431 says so rather than filing it as "flaky test". Not a blocker for this PR (which touches neither package), but worth someone's eyes.

0xGosu:
Both review findings were valid and are fixed in 8663e6e.

**1. `match=` detached from its hit** — confirmed worse than the "approximate centring" my comment claimed. The offset came from the folded screen but sliced the original, and `İ`/`ẞ` lose bytes when lower-cased (2→1, 3→2), so the window drifts one byte per such rune. `foldRunewise` maps rune-by-rune through `unicode.ToLower`, which preserves rune *count*, so a rune index means the same position in both strings; `matchExcerpt` now takes a rune index directly. Both sides fold the same way — folding only one makes exactly those terms unsearchable, which `TestScreenSearchFoldsBothSidesTheSameWay` covers.

Regression test as requested (`TestScreenMatchSurvivesWidthChangingCaseFolds`, 200×`İ` + 200×`ẞ` ahead of the term). Reverting to the byte offset makes it fail with a `match=` of solid `ẞ` and no `npm` in it — the exact symptom you described.

**2. Hint widened a quoted phrase** — correct, and it applied to every cross-mode hint, not just the screen one, since all six were built from the joined query. `shellJoin` re-quotes the original argv words for all of them, so a suggestion always reproduces the query it claims to re-run:

```
$ hap signatures search --screen "npm install" force
- `hap signatures search 'npm install' force` — the rule's own fields and its masked salient instead
```

A plain word stays unquoted; a word carrying a shell metacharacter gets quoted too. `TestSignaturesSearchHintPreservesPhraseBoundaries` covers the emitted footer and fails on the old joined-query form.

gofmt, go vet, golangci-lint and the `cli`/`frontend`/`store`/`tui` packages all pass locally; pushing for CI.

0xGosu:
@CharlieHelps please re-review — both findings addressed in 8663e6e (rune-count-preserving fold so `match=` stays on its hit; `shellJoin` so the hints keep argv phrase boundaries), each with the regression test you asked for and each proved by mutation.

PR #461, opened 2026-09-12
#0 fix: re-prove agy's composer before a hand-out and verify it landed

0xGosu:
Findings 1 and 2 from `/tmp/agy-findings.md` (live agy run, hap v0.9.21).

## What was actually wrong

The finding proposed "extend PR #454's empty-composer proof to the delivery path". That proof **already** gated every typed-text path on main, and it refuses correctly on all 31 agy fixtures — so the stated fix was already shipped. The real defect is a **TOCTOU window**, not a missing gate:

`deliverAutonomousClaimed` takes the proof, then writes the audit row and reserves the checklist item — each a remote round trip under turso or a `github_gist` source — and only then sends. Seconds can pass in which agy draws an approval over the composer that was empty when we looked.

That matters because agy commits a menu choice on the **bare digit** and does not honour bracketed paste at a modal, so a hand-out landing there selects an option and spills its remainder into the composer. Observed live: the hand-out boilerplate's "index `2`" took the conversation-wide *always allow* row. **Option 3 on that menu writes `settings.json`.**

## The fix

**(a) A last look before the keystrokes.** The proof is re-taken immediately before `SendToAgent`, the way `claimBlockedBy` re-asks its controls before an auto-accept claim. A refusal releases the reservation and retires the row as `ignored` — nothing was typed, so it is not a delivery failure and the operator is not notified.

**(b) Verify the hand-out landed.** herdr accepting the keystrokes is not agy starting the task: text typed mid-turn is *queued* in the composer and fires whenever that turn ends, out of order. Such an item is left `[-]` and escalated, and deliberately gets **no ledger row** — `reclaimStrandedTasks` would otherwise return it to `[ ]` while the queued copy is still on its way to the same agent. Nothing can clear an agy composer from outside (herdr rejects every spelling of `C-u`), so that would be a double send that cannot be called back.

`domain.AgyComposerDraft` reads the composer block back *between its rules*, so a wrapped multi-line hand-out is seen. `AgyDraftMatches` attributes it on a **prefix**, never containment, so a short operator draft is not mistaken for hap's text. Attribution is biased to refuse when unsure: a draft wrongly attributed to hap strands a task for a human, while the other way just falls back to the ledger path that already existed.

## Tests

- `TestAgyHandoutRefusedWhenAMenuAppearsAfterTheComposerProof` — a new fake **pane script** serves a different screen to each successive read. This is the trap: a fake answering every read identically passes with or without the second proof. It asserts the `ignored` audit row, which proves the *opening* proof passed and only the last look refused.
- `TestAgyQueuedHandoutIsLeftInProgressAndEscalated` — asserts no ledger row, no rollback, and an escalation.
- Domain cases for `AgyComposerDraft` (wrapped multi-line, mode placeholder, standing approval) and `AgyDraftMatches` (a mid-text fragment must NOT match).

## Gate

`go build`, `gofmt`, `go vet` clean. **No new `golangci-lint` findings** (baseline on this machine is 2 pre-existing stdlib typecheck errors — the installed linter is built with go1.26.2 and cannot typecheck the go1.27 stdlib).

Full-suite note: `internal/daemon` is timing-sensitive and this devbox is loaded by another agent — five consecutive whole-package runs each failed on a *different* test, including two runs on untouched `main`. Every test named here passes deterministically in isolation.

## Note on the findings doc

Finding 7 ("agy reports `done`, not `idle`, so anything keyed on idle misses it") is **wrong**: `autoSendParked`, `domain.AgentBusy`, `sessionRenameParked`, `autoAcceptParked` and the classifier all accept `done` beside `idle`, and hand-out confirmation keys on the `working` transition, never on the parked string. It is untested for `done`, not broken — carried to the later PR.

---

## Review round 2 (0feaa30, cc483e3)

Three of the four findings were right; the verification half was where all of them landed.

**461-A — the read-back had no settle.** `ports.SendToAgent` returns as soon as herdr has written the request (`retrySubmit` covers codex and claude only), so the composer was read while an *accepted* hand-out was still standing in it. A successful delivery was then recorded as queued: `[-]`, no ledger row, no counted attempt — unreachable by both `reclaimStrandedTasks` and the `maxTaskHandouts` ceiling, on the happy path. The wait (`agyHandoutSettleDelay`, the second `scheduleUnblockCheck` already uses for the same question) and both outcomes now live in `verifyAgyHandout`, **off the loop** — `deliverAutonomousClaimed` runs in the select loop's own goroutine.

**461-B — the scan window was a fixed 20 lines**, smaller than a wrapped hand-out, so the composer read as empty for exactly the *large* hand-outs and took the ledger path: the double send this feature exists to prevent. The window now scales to what was sent (`AgyDraftScanLimit`), and running out of it is a third verdict, `AgyDraftUnreadable`, treated as **queued** — an empty composer finds its opening rule on the first step of the scan, so exhausting the window means something large is in there.

**461-C — `agyDraftMatchRunes` was only a cap.** It is a floor too now, except when the draft IS the whole of what was sent (a flat floor would refuse a short hand-out that really was queued — the costly direction). And the comparison drops whitespace on both sides, so a wrap falling mid-word no longer stops a queued hand-out from being recognised.

**461-D — pushed back, with a test.** A standing approval and an unreadable pane are both refused by the *opening* proof, above the audit write and `reserveDeclaredTask`, so neither costs a task-list write or an audit row per sweep; the last look is only reached when the screen changed between the two proofs. `TestAgyApprovalIsRefusedBeforeAnythingIsReserved` pins it. One genuine exception named on the thread: the pre-delivery task review reserves before `deliverAutonomousClaimed`, so with `enable_llm_review_before_auto_send` on the churn does exist — above the last look, pre-existing, and not reached by either suggested remedy.

### Tests

The old case could not exhibit 461-A: the fake repaints synchronously inside `Send`. `fakeHerdr` now records **when** each send and read happened, and `agyHandoutRepaint` repaints on a delay.

- `TestAgyHandoutIsNotReadBackBeforeAgyCanRepaint` — asserts the ledger row, no queued escalation, **and** that the deciding read came at least a settle after the send. Content alone can never discriminate.
- `TestTheAgyHandoutSettleIsNotTakenOnTheMainLoop`, `TestATallQueuedAgyHandoutIsStillLeftInProgress`, `TestAgyApprovalIsRefusedBeforeAnythingIsReserved`, `TestATallComposerBlockIsUnreadableRatherThanAbsent` (with a scaled-window control), `TestAgyDraftScanLimitScalesWithWhatWasSent`, plus new `AgyDraftMatches` cases.

Each verified by mutation: forcing the delay to a nanosecond, and restoring the fixed 20-line window, fail the cases that name them.

### Gate

`go build`, `gofmt`, `go vet` clean; **`golangci-lint` 0 issues** (baseline on this machine is also 0 once the native submodule is initialised, so the bar is exact rather than "no new"). Full unit suite green under `-p 2` apart from `internal/store/turso`'s `TestALockedMigrationStepLosesTheLeaseAndFailsClosed`, which passes alone and is untouched by this diff. Integration suite green; `HAP_ITEST_AGY=1` real-agy cases all pass (`TestRealAgyChoiceConfirm` is flaky against a live Gemini — 2 failures then a pass at this commit, and it also passes at the previous tip; the diff does not touch `mcqdeliver.Agy` or the confirm path).


0xGosu:
**Flagging a transport-level constraint that affects part (b) of this PR — worth reading before merge.**

The findings doc grew an addendum after this branch was cut. **Finding 10** reports that a partially delivered hand-out silently *corrupted* the task text: roughly 200 characters vanished mid-sentence, and what reached the agent still read as valid English but had lost its object. That raises delivery verification from bookkeeping to correctness.

Two things follow that this PR does **not** cover:

1. **This PR's check runs after the send, so a mangled-but-submitted hand-out is invisible to it.** `agyHandoutQueued` reads the composer back to catch text that was *queued* rather than acted on. If a truncated hand-out actually submitted, the composer is empty, and the delivery is recorded as clean. The corruption case and the queued case are different failures; only the second is fixed here.

2. **Finding 10's prescribed cure — "read back the full text before pressing Enter" — is not implementable on the current transport.** A multi-line hand-out routes through `agent prompt`, which writes the text *and* its Enter in one bracketed-paste-aware request (`internal/herdr/cli.go`, `submitPasted`), so there is no pre-submit moment at which the composer can be inspected. The obvious alternative — `pane send-text`, verify, then `pane send-keys enter` — is not available either: `pane send-text` is not paste-aware, so every embedded newline arrives as a literal Enter and the hand-out submits its first line while typing the rest into the next prompt. That split is exactly what `submitText` exists to prevent.

So closing finding 10 needs a transport change, not just a verification step — for example a herdr primitive that writes without submitting (or the composer-clearing primitive the findings doc also asks for, which `send-keys C-u` cannot provide today). Worth deciding before finding 10 is scoped into its own PR.

Also note the addendum's **correction to finding 1**: findings 11 and 12 show hap scoring an agy pane as `idle` while an "Accept this file edit?" modal stood on screen, which means the residual risk is in *classification*, not in this gate. This PR closes the TOCTOU window and is still worth having, but it is not on its own sufficient to stop a hand-out reaching a modal — the classifier has to stop reporting idle first. That work is tracked separately.

internal/daemon/daemon.go
```diff
@@ -3323,7 +3347,21 @@ func (d *Daemon) deliverAutonomousClaimed(ctx context.Context, s domain.Situatio
...
+	// agy is the one agent that can accept the keystrokes and still not start:
+	// it QUEUES text typed during a turn in its composer. Reading the composer
+	// back is what tells that apart from a turn that began, and a queued item
+	// must NOT get a ledger row — reclaimStrandedTasks would return it to "[ ]"
+	// while the queued copy is still on its way to this same agent, and that
+	// double send cannot be called back (herdr rejects C-u, so nothing can clear
+	// an agy composer). It stays "[-]" and the operator is asked instead.
+	switch {
+	case reservedIndex > 0 && domain.IsAgy(s.AgentType) &&
+		d.agyHandoutQueued(ctx, s.PaneID, del.sendText):
```
0xGosu: **The read-back has no settle delay, so the normal success path can be misread as "queued".**

`ports.SendToAgent` returns as soon as herdr has written the request — for agy nothing waits for the agent to repaint (`retrySubmit` in `internal/herdr/cli.go` is `kind == "codex" || kind == "claude"` only). `agyHandoutQueued` then issues its visible read immediately. Every other "did our keystrokes land" path in this repo interposes a delay: `scheduleUnblockCheck` (fixed 1s), `mcqdeliver` (`sweepKeyDelay` between press and re-read), `recaptureAfterAgyAnswer` (goes through `scheduleCapture`).

Scenario: an idle agy accepts the hand-out normally, but the capture lands in the window before agy has cleared its composer and painted the new turn. `AgyComposerDraft` returns the hand-out text, `AgyDraftMatches` confirms it, and a *successfully delivered* item is left `[-]` with **no `task_reservations` row and no `RecordTaskHandoutAttempt`** — so neither `reclaimStrandedTasks` nor the `maxTaskHandouts` ceiling can ever reach it. Unlike `escalateNeverStartedTask`/`escalateUndeliverableTask`, which also leave `[-]` but only at a bounded ceiling, this is operator-only forever, on the happy path.

`TestAgyQueuedHandoutIsLeftInProgressAndEscalated` cannot exhibit this: `onSend` repaints the fake pane *synchronously inside `Send`*, so the queued screen is guaranteed present before the read. It proves the branch fires when the screen says queued; it says nothing about whether the screen says queued when the hand-out actually succeeded.

internal/domain/agyanswer.go
```diff
@@ -189,3 +189,111 @@ func AgyComposerReady(pane string) bool {
 	_, held := AgyHeldForm(pane)
 	return !held
 }
+
+// agyComposerScanLimit bounds the search for the composer's opening rule. A
+// queued multi-line hand-out renders as several lines between the rules, so the
+// block cannot be read at a fixed offset the way AgyComposerReady reads an empty
+// one — but scanning the whole capture would happily pair the closing rule with
+// one agy drew between two earlier turns.
+const agyComposerScanLimit = 20
```
0xGosu, line 202: **A 20-line scan cap is smaller than a realistic hand-out, and the miss direction is the dangerous one.**

The opening rule is searched only `agyComposerScanLimit` lines above the closing rule. A hand-out is a rendered task prompt (boilerplate + task text) and wraps at terminal width — well past 20 lines for anything but a one-liner. When it does, the loop at :232 finds no rule, `open < 0`, `AgyComposerDraft` returns false, and the caller takes the ledger branch: `RecordTaskReservation` is written, and a few minutes later `reclaimStrandedTasks` returns the item to `[ ]` — while the queued copy is still on its way to this same agent. That is exactly the uncallable-back double send this PR exists to prevent, and it fires for the *large* hand-outs, i.e. the common case.

The doc comment on `agyHandoutQueued` says the miss direction "merely falls back to the ordinary ledger path that already existed" — but that is only safe when the item really was taken up. For a genuinely queued hand-out the ledger path is the unsafe one, so the claimed bias is inverted for precisely the case this function exists to catch.

Either scale the limit to the sent text's expected line count, or treat "closing rule present but no opening rule within the window" as unknown and fall back to a stricter answer than "delivered".

internal/domain/agyanswer.go
```diff
@@ -189,3 +189,111 @@ func AgyComposerReady(pane string) bool {
...
+// ledger path that already existed. So this refuses whenever it cannot tell.
+func AgyDraftMatches(draft, sent string) bool {
+	d := strings.Join(strings.Fields(draft), " ")
+	s := strings.Join(strings.Fields(sent), " ")
+	if d == "" || s == "" {
+		return false
+	}
+	head := []rune(d)
+	if len(head) > agyDraftMatchRunes {
+		head = head[:agyDraftMatchRunes]
```
0xGosu, line 406: **`agyDraftMatchRunes` is a cap, not a floor — so "refuses whenever it cannot tell" is not what the code does.**

The doc says the window is "long enough that an operator's own sentence cannot collide by accident", but the only use of the constant is truncation: a draft *shorter* than 24 runes is compared in full. If the operator has typed one or two characters in the moment after the send and they happen to open the hand-out text (`"N"` against `"Next task: …"`), `strings.HasPrefix(sent, head)` is true, the operator's draft is attributed to hap, and the task is stranded at `[-]` with no ledger row — the outcome the comment calls the costly one. A minimum length (refuse when `len(head) < agyDraftMatchRunes`) is what the comment already describes.

Related, same function: the `" "` join at :250 inserts a separator that is not in `sent`. On a narrow pane, where the composer's first line is shorter than 24 runes, the compared head crosses the wrap (`"…run the su" + " " + "ite…"`) and never prefixes `sent`, so a genuinely queued hand-out is mis-attributed the other way, into the ledger/reclaim path above.

internal/daemon/daemon.go
```diff
@@ -3272,6 +3272,30 @@ func (d *Daemon) deliverAutonomousClaimed(ctx context.Context, s domain.Situatio
...
+	// was empty when we looked. agy commits a menu choice on the BARE DIGIT and
+	// does not honour bracketed paste at a modal, so a hand-out landing there
+	// selects an option and spills its remainder into the composer: observed live
+	// 2026-09-12, where the hand-out boilerplate's "index 2" took the
+	// conversation-wide "always allow" row. Option 3 on that same menu writes
+	// settings.json, so nothing here may be sent on a stale proof. Same shape as
+	// claimBlockedBy's last look before an auto-accept claim.
+	if domain.IsAgy(s.AgentType) {
+		if err := d.agyComposerRefusal(ctx, s.PaneID); err != nil {
+			slog.Info("agy's composer stopped being ready between the proof and the send; nothing was typed",
```
0xGosu, line 3292: **The last look sits below the reservation, so a pane parked in a modal now churns the checklist once per sweep, unbounded.**

The opening proof is deliberately above `reserveDeclaredTask` so a not-ready pane costs nothing. This one runs after it, and its refusal path is `rollback()` + an `ignored` audit row with no `RecordTaskHandoutAttempt`. An agy agent sitting on a standing approval (the exact state this gate is for) therefore does reserve → refuse → release every idle poll: two task-list writes per minute, which under a `github_gist` source are two remote read-modify-writes, plus an `ignored` audit row each time, forever — and nothing counts the attempts, so the `maxTaskHandouts` ceiling can never end it. Also note an unreadable pane takes this same branch (`agyComposerRefusal` fails closed on a read error), so a transient herdr failure produces the same loop.

Cheapest fix is to take the last look before the reservation as well, or to count the refusal so the existing ceiling bounds it.

0xGosu (commented):
Reviewed the agy delivery gate. The last-look gate itself is well placed and the "never roll a queued hand-out back to `[ ]`" reasoning holds. Four findings, all on the new verification half (`agyHandoutQueued` / `AgyComposerDraft` / `AgyDraftMatches`) — the top one is that the read-back has no settle delay, so a successful hand-out can be misread as queued and stranded with no ledger row and no attempt counter to recover it.

internal/daemon/daemon.go
0xGosu: Fixed in 0feaa30 — you were right, and it was on the happy path.

The read-back now waits `agyHandoutSettleDelay` (1s, the same constant `scheduleUnblockCheck` uses for the same question) before it looks. It runs **off the loop**: `deliverAutonomousClaimed` is reached from `handleAttention` and from the idle sweep, both in the select loop's own goroutine, so waiting inline would stall every other agent for a second per hand-out. The wait and both outcomes now live in `verifyAgyHandout` (`internal/daemon/keyedform.go`), scheduled through the tracked `afterFunc` so shutdown drains it; the ledger write is factored out as `recordHandoutReservation` and is what the daemon takes inline if it is already shutting down and nothing will run the read-back — an item at `[-]` with no row is unreachable by every sweep, so that case takes the same ordinary path an unreadable pane takes.

On the test: agreed the old one proves nothing about this. `fakeHerdr` now records **when** each send and each read happened, and `agyHandoutRepaint` models what a real pane does — the hand-out is briefly still in the composer, then agy clears it tens of milliseconds later. `TestAgyHandoutIsNotReadBackBeforeAgyCanRepaint` asserts both halves: the ledger row exists and no queued escalation was raised, **and** the gap between the send and the read that decided it is at least the settle. The gap assertion is there because content alone can never discriminate — a fake can always be made to repaint fast enough. Verified by mutation: forcing the delay to a nanosecond fails it with exactly the reported symptom (`agy queued the hand-out in its composer instead of starting it`). `TestTheAgyHandoutSettleIsNotTakenOnTheMainLoop` pins the other half.

One thing I considered and dropped: a second look before calling it queued. 461-B and 461-C already cut the false-positive rate, and a second knob is a second failure mode in a change that is already moving four things.

internal/domain/agyanswer.go
0xGosu: Fixed in 0feaa30, both halves you named.

**Scaled window.** `AgyComposerDraft` now takes the scan limit, and `agyHandoutQueued` passes `domain.AgyDraftScanLimit(sent)`: the sent text's line count at a pessimistically narrow 40-column wrap, floored at the old 20 and capped at 120 (`agyComposerScanCeiling`) so it can still never wander the whole transcript.

**Unknown is a third verdict.** Running out of window is `AgyDraftUnreadable`, never `AgyDraftNone`, and `agyHandoutQueued` maps it to **queued**. The reasoning, which is now in the doc comment because it is the step a reader will doubt: an empty composer finds its opening rule at `n-4` on the first step of the scan, always — and so does a one-line draft — so exhausting the window means something *large* is in the composer, and the largest thing that could just have arrived there is the hand-out. Reading that as "delivered" is precisely the ledger-then-reclaim double send.

You are also right that the doc comment's claimed bias was inverted, and I have rewritten it rather than just widening the window: the ledger path is only safe when the item really was taken up, so neither direction is free and the function says so now.

The widened scan keeps the guard that made the narrow one safe — the block's first line must still be a `>` caret line, which is what stops a far rule agy drew between two turns from being paired with the composer's closing one.

Tests: `TestATallComposerBlockIsUnreadableRatherThanAbsent` (with the same pane and a scaled window as its control, so it cannot pass by the parser simply ceasing to recognize the screen), `TestAgyDraftScanLimitScalesWithWhatWasSent`, and at the daemon level `TestATallQueuedAgyHandoutIsStillLeftInProgress` — a 12x-repeated hand-out wrapped every four words, which must get no ledger row. Restoring the fixed 20-line window fails that last one.

internal/domain/agyanswer.go
0xGosu: Both fixed in 0feaa30, with one deliberate departure from the literal suggestion.

**The floor.** `AgyDraftMatches` now refuses a short head — but only when it is *not* the whole of what was sent. A flat `len(head) < agyDraftMatchRunes` refusal, as written, would refuse a genuinely queued **short** hand-out, and that is the costly direction: it gets a ledger row and `reclaimStrandedTasks` re-offers it while the queued copy is still on its way. So the rule is "shorter than the window is evidence only when it IS everything we sent". `TestAgyDraftMatchesOnlyTheTextHapSent` covers `"N"` and `"Next t"` as refusals and keeps a `"Next task: fix #12"` full-length short hand-out as a match.

The floor is measured on the draft as it *reads* (whitespace collapsed to single spaces) while the comparison is on whitespace-*stripped* text, because the two ask different questions: how much did somebody have to type, versus where did the wrap fall.

**The join separator.** Agreed, and the fix is one level up from removing the join: both sides now have **all** whitespace dropped before comparing, so the answer cannot depend on where the wrap fell at all. A terminal breaks the line wherever the column runs out — mid-word as readily as at a space — so no separator is correct at every width, which is the same reason this repo does not stitch wrapped MCQ option labels. Two new cases cover it: a wrap that fell mid-word, and a narrow pane wrapping every few words. 24 runes of non-whitespace is still far more than an operator's opening words can collide with.

I checked the agy fixtures and `docs/designer/agy-support.md` for a wrap-indicator glyph (`↵`, a trailing `…`, a soft hyphen) before relying on this — agy renders none, it hard-wraps.

internal/daemon/daemon.go
0xGosu: Pushing back on this one, with a test rather than an argument — but I have added the test, because the claim deserves to be enforced rather than reasoned about.

The scenario does not hold: **an agy agent sitting on a standing approval never reaches the reservation.** The opening proof (`deliverAutonomousClaimed`, `daemon.go:3230`) is above *both* the audit write and `reserveDeclaredTask`, and `AgyComposerReady` refuses an approval outright — the composer sandwich is not at the bottom of that capture and `AgyHeldForm` matches. `abandon()` runs: no task-list write, no audit row, no `ignored` row. An unreadable pane is the same story — `agyComposerRefusal` fails closed at *both* proofs, and the first one is the one it hits.

The last look is therefore only ever reached when the screen **changed between the two proofs**, which is by definition not a steady state — that is the race it exists for, and it is why its refusal path can afford to be `rollback()` plus one `ignored` row.

0feaa30 adds `TestAgyApprovalIsRefusedBeforeAnythingIsReserved`: both your cases (a standing approval, and a pane that cannot be read), with no caller-owned claim so the delivery would do its own reserving, asserting zero `MutateTaskFile` calls and zero audit rows. Deleting the opening proof fails it, so the placement of the last look stays affordable by CI rather than by my reading.

On the suggested remedy: I would not take `RecordTaskHandoutAttempt` on this refusal even if the loop were real. Nothing was typed and the item is perfectly deliverable — counting it against `maxTaskHandouts` ends in `escalateUndeliverableTask` leaving `[-]` for a task no agent ever refused, just one that was busy five times. The attempt counter means "this could not be delivered", and a not-ready composer is "not yet".

If you would still rather pay a third read to close the residual window, say so and I will move a proof above the reservation as well — but it cannot remove the last look, since the reservation still happens between it and the send.

0xGosu: Narrowing my own claim before you find it: "never reaches the reservation" is true of the ordinary hand-out path, not of **every** path.

The pre-delivery task review is the exception. `deliverDeclared` → `startTaskListReview` → `taskfile.ApplyReview` reserves the item *before* `deliverAutonomousClaimed` runs, so `del.rollback` arrives already set and the opening proof's `abandon()` releases it. For an agy agent parked in a modal that really is reserve → refuse → release per sweep — the churn shape you described, just located above the last look instead of below it.

Three reasons I have left it alone here rather than widening the change:

- it is reachable only with `enable_llm_review_before_auto_send` on for the source, which is a per-source opt-in and off by default;
- it predates this PR — the opening proof is #454's, and this branch only added the proof *below* the reservation;
- neither remedy in your comment reaches it. Moving the last look above the reservation does nothing when the reservation happened in a different function, and counting the refusal is wrong for the reason above: nothing was typed and the item is still deliverable.

The fix, if you want one, is in the review path — it should take the composer proof before `ApplyReview`, not after. Happy to do that here or as a follow-up; say which.

0xGosu: Filed as #472 so it is not lost, and not held against this PR: it is pre-existing (the opening proof is #454's; this branch only added the one below the reservation), it needs `enable_llm_review_before_auto_send` on for the source, and the fix belongs in the review path — taking the composer proof before `taskfile.ApplyReview` rather than after.

PR #460, opened 2026-09-12
#0 refactor: remove unreferenced helpers and consolidate signature/cli formatting

0xGosu:
## Summary
This PR applies the safe subset identified during codebase cleanup analysis:

### 1. Dead Code Removal
- Removed unreferenced `NodeWatching` in `internal/domain/node.go`.
- Removed unreferenced `SeedRuleDisabled` in `internal/frontend/frontend.go`.
- Removed unreferenced `TaskFilePath` in `internal/frontend/frontend.go` (leftover after CLI migrated to `TaskListFor`).
- Removed unreferenced `GetTask` in `internal/frontend/frontend.go` (superseded by `ResolveTaskRef`).

### 2. Duplication Consolidation
- Extracted shared 16-character signature truncation logic into `domain.ShortSignature` in `internal/domain/signature.go` (with unit test `TestShortSignature`).
- Delegated CLI `shortSignature` and TUI `shortSig` to `domain.ShortSignature`.
- Eliminated redundant `orDashCLI` in `internal/cli/cli.go` in favor of existing `orDash`.

### 3. Changelog
- Added changelog fragment `changelog.d/chore-codebase-cleanup.md`.

### Verification Gate
- `bash scripts/check-submodule-gitlink.sh`: PASS
- `go build -tags "vectors cpu" ./...`: PASS
- `go test -tags "vectors cpu" ./... -count=1 -p 2`: PASS (all suites pass, flake-prone turso/daemon tests re-verified in isolation)
- `gofmt -l . | grep -v submodule`: PASS (clean)
- `go vet -tags "vectors cpu" ./...`: PASS (0 diagnostics)
- `golangci-lint`: PASS (no new issues; 0 issues under Go 1.26 toolchain)
- `go test -tags integration ./test/integration/ -v`: PASS
- `TestRealAgyDetection` with `HAP_ITEST_AGY=1`: PASS

PR #462, opened 2026-09-12
#0 fix: stop the daemon panicking on a bounded wait for turso sync ops

0xGosu:
The hap daemon panics under concurrent load and silently auto-restarts. Reported from `daemon.stderr.log`; one stack carries trimmed module paths — the shape of the **installed v0.9.21 release** — so this is shipped code, not a dev-only artifact.

```
panic: sync: WaitGroup is reused before previous Wait has returned
sync.(*WaitGroup).Wait(0xc000334500)
turso.waitBounded.func1()   internal/store/turso/turso.go:276
created by turso.waitBounded in goroutine 1
```

## Root cause

`waitBounded` spawned a goroutine to call `wg.Wait()` and abandoned it on timeout — leaving it parked in `Wait` **forever**. `WaitGroup`'s contract forbids an `Add` that races a `Wait` which has not returned, so the next time the counter fell to zero while an operation called `Add`, the runtime panicked and killed the daemon.

This is not a tuning problem: **a `WaitGroup` cannot express a bounded wait at all**, because every timeout strands a `Wait`. One slow `Close` armed the hazard permanently.

Replaced with `opGate` — a counter plus an `idle` channel recreated on each 0→1 transition. A waiter merely *selects* on a snapshot, so giving up leaves nothing behind.

## Removing the panic is not sufficient on its own

This is the trap, and it is why the change is bigger than deleting a goroutine: **the panic was the only thing stopping an operation that began after `Close` from running against a freed handle.** The stacks show precisely that — `Add` arrived *after* the wait had reached zero.

The path is real and reachable twice over. `daemon.fleetRun` abandons an in-flight sync op at shutdown ("leaving it to the adapter") but that goroutine keeps running: `fleetPush` calls `sync.Push()` (its `Done` drops the counter to zero, waking `Close`) and *then* calls `fleetRefreshStats` → `Stats()` → `Add(1)`. `fleetFinalPush` leaves a `Push` behind its own budget the same way.

So the gate **latches**: after `beginClose`, `add()` refuses and the operation returns `ErrClosing` instead of touching the native handle. Trading a loud crash for a silent use-after-close against native memory would have been the worse bug — and it would have been invisible.

## Tests

The panic cannot be asserted directly: it is inherently racy, and a panic raised in another goroutine *aborts the test binary* rather than failing a case. The deterministic signature is the **goroutine leak**, so that is what the regression test measures — across 50 waits rather than one, because a ±1 goroutine assertion would be exactly as load-flaky as this package's existing cases.

Verified empirically against the old implementation: **50 timed-out waits stranded exactly 50 goroutines** (the new test's threshold is 10), and the panic reproduced with the same stack shape. So the assertion genuinely discriminates rather than passing vacuously.

- `TestATimedOutWaitStrandsNoWaiter` — the regression test.
- `TestOpGateRefusesOperationsOnceClosing` — the latch.
- `TestOpGateUnderConcurrentOpsAndWaits` — the production shape, under `-race`.
- `TestOpGateGivesUpOnAHungOperation` — the hung-native-call invariant that documents `closeWait`, **ported rather than deleted**.

All four are DB-free, so they run without `tursodb` on PATH.

## The restart / lock race — diagnosed, deliberately not "fixed"

Each panic is followed by `another daemon already holds the lock: resource temporarily unavailable`. That is the restarting daemon racing the dying process's `flock`, which the kernel releases asynchronously as the process tears down — not a defect in `daemonlock.Acquire`, whose fail-closed `LOCK_EX|LOCK_NB` semantics `EnsureFresh` and the `--ensure`/`--restart` distinction both depend on. Loosening it there would weaken a load-bearing guarantee to paper over a symptom.

With the panic gone the restart storm goes with it, since nothing is crash-looping. If a retry is still wanted it belongs in the *restart* path as a short bounded retry on `EWOULDBLOCK`, not in `Acquire` — worth its own change, with its own reasoning about a daemon that is genuinely still alive.

## Surfacing an unclean restart in `hap status` — deferred

`daemon.started` with no preceding `daemon.stopped` is the right signal, but nothing persists that today, so it needs a marker written at clean shutdown and checked at start. Additive and independent of this fix; better argued on its own merits than smuggled into the crash fix.

## Gate

`go build`, `gofmt`, `go vet` clean; **full `go test -tags "vectors cpu" ./... -count=1` green**; no new `golangci-lint` findings (the 2 pre-existing stdlib typecheck errors are the linter being built with go1.26.2 against the go1.27 stdlib).


0xGosu:
**CI is green on both platforms after a rebase onto current main.** Run [34674568065](https://github.com/0xGosu/herdr-auto-pilot/actions/runs/34674568065): Tests (ubuntu-latest) ✓ 14m51s, Tests (macos-14) ✓ 13m32s, plus lint/gofmt/changelog/allowlist.

Classifying the earlier red run explicitly, as asked.

## (a) A real regression from the opGate latch — ruled out

Not on balance-of-evidence but structurally: **`internal/daemon` does not import `internal/store/turso` at all** (the only occurrence of that path in the package is a comment in `fleetsync.go`), and the daemon tests open a plain SQLite store via `store.Open(dbPath)`. There is no code path by which `opGate`, `add()`, `beginClose()` or `ErrClosing` can be reached from `TestLLMFallbackStagingRegateAndPromotion`. The staged-decision path it exercises is store + select-loop work that never touches the sync engine.

Corroborating: the same case failed for me **locally on `feat/never-auto-actions`**, a branch that does not contain this change at all.

## (b) Load-sensitive flakiness — the best-supported explanation

The failure is `daemon_test.go:3729` "staged decision should be accepted" with the decision still `Status:pending` — a timing assertion on an asynchronous promotion, exactly the shape that loses under CPU contention.

Locally on **untouched main (20a5a6e)**:
- that case in isolation: **0/5 runs failed**;
- the **whole `internal/daemon` package** failed under load with `TestFSPImmediateAcceptOnEscalate` and `TestLLMAutoRowCarriesSessionID` — different cases each run, which is the signature of contention rather than a defect;
- quiet, the same package passes 2/2.

Across five consecutive whole-package runs on this devbox (my branch and untouched main), five *different* tests failed. `internal/store/turso`'s schema-lease case behaves the same way, failing once and passing once in two consecutive quiet runs.

I could not satisfy the "same case on untouched main under the same CI conditions" bar directly, because CI only runs on `pull_request` — there is no CI run for a main push to compare against, and I did not want to open a throwaway PR to manufacture one. I am flagging that as the one gap in this argument rather than papering over it.

## (c) Interaction with #460 — not excluded, but unlikely to be the cause

The rebase is what I changed, so I cannot rule it out from one green run. But it argues against itself on timing: the red run started 04:38:13 and #460 merged to main at 04:41:04, so **#460 was not in the merge commit that failed** — it landed afterwards. The delta between red and green is therefore a new base plus a re-roll of the same nondeterminism, not the removal of an interaction with #460.

## One thing I could not check

You asked whether macOS failed the same case. **I could not retrieve it** — `gh run view --log-failed`, `--log`, and the jobs-logs API all returned empty for job 103499203877 on the old run, so I only ever had the ubuntu detail. The macOS job is green on the rebased run, so I did not pursue it further; say the word if you want it dug out of the expired logs.

## Bottom line

Nothing in this PR can reach the failing code path, the failing package is demonstrably nondeterministic on this hardware, and the rebased branch is green on both platforms. I'd merge on that basis — but the flaky daemon suite is worth its own issue, since it will keep costing red runs on unrelated PRs.

PR #464, opened 2026-09-12
#0 feat: never-auto rules that match the answer, not the screen

0xGosu:
Finding 3 from `/tmp/agy-findings.md`.

## The bug

A never-auto rule matches **situation** text, so it cannot express "never pick *that* option". The orchestrator tried, to protect an undecided policy on agy's "(Persist to settings.json)" approval row:

```
hap config rules add --agent-type agy '(?i)[Pp]ersist to settings\.json'
```

That row is printed on **every** agy approval, so the rule matched every prompt, escalated all of them, and the agent could not progress. Removed four minutes later.

Note this is not simply "the rules only screen the pane" — never-auto is *already* matched against outbound text at several sites. The defect is that **one pattern list serves both meanings**, so a rule written to constrain the answer also constrains the screen.

## The change

A second list, `[[safety.never_auto_actions]]`, matched only against the action hap is about to send. Same agent-type scoping; one list and one index space, since the flat/scoped split above exists only for backward compatibility and is what forced two removal verbs.

Applied at the **action** sites only:

| Screens the action (new check added) | Screens the screen (untouched) |
|---|---|
| `act`, `startActionReview` | `decideAndActResolved` |
| `handleActionReviewOutcome` (×2) | `handleActionReviewOutcome` (scan) |
| `handleLLMOutcome` | `handleLLMOutcome` (scan) |
| `actionScreen` (orchestrator) | `sweep` |

Three exclusions are deliberate: a generated **task prompt** (`tasklistreview`, `generatedTaskUnsafe`) — these rules name menu options, and matching them against prose an agent is being *asked to do* would refuse work for containing a phrase; and an operator's own reply in `deliverreply` — a human choosing to widen a permission is the human's call, the same reasoning that leaves the operator path unscreened. The orchestrator is included precisely because an LLM's confirm is not a human's.

`ReasonNeverAutoMatch` is **reused** rather than a new reason minted: a new `EscalateReason` must be classified against `autoAcceptExcludedReasons` and `AutoDismissReasons`, and getting that wrong is a failure mode this repo documents twice. The diagnostic text distinguishes the two lists instead.

`NeverAutoActionList` is a distinct type, not an alias, so it exposes `Match` and nothing else — the matcher underneath also carries `SuspectedIrreversible`, and an action rule reaching that heuristic would judge action text against indicators written for pane content. `Kind` is pinned strict in three places for the same reason.

## The seeds ship OFF, and that is the load-bearing decision

Shipped rules for the scope-widening options (`and always allow`, `Persist to settings.json`, `allow all`) are included but inert unless `safety.enable_never_auto_action_seeds = true`.

Refusing an option only **escalates** it — it does not pick a narrower one. So with these on and nothing else changed, every approval whose best answer is a widening row becomes an escalation, and since `ReasonNeverAutoMatch` is in `autoAcceptExcludedReasons`, each one is permanently operator-only. That is the original queue-blocking failure reproduced on the action side, and it would also feed finding 5's escalation flood.

They become safe to enable in the same change that biases agy answer selection toward the narrowest satisfying option (finding 4), which is where the default should flip. **Until then this PR ships the mechanism and the operator surface, and changes no behaviour unless a rule is added.**

## Tests

The one that matters is the **control**: same rule, same pane — which *contains* "always allow", because the menu prints it every time — with the narrow answer chosen. It must be **delivered**. Without that case the suite passes whether the pattern is matched against the answer or against the screen, which is the entire bug.

- `TestNeverAutoActionRuleIgnoresThePaneItIsAnswering` — the control.
- `TestNeverAutoActionRuleEscalatesTheWideningAnswer` — the refusal.
- `TestSituationRulesStillFireBesideTheActionList` — its mirror; the existing rules keep working.
- Domain cases for scoping, seeds-off-by-default, per-pattern silencing, and strict-kind pinning; a CLI case covering `add --action`, its own listing heading, `remove-action`'s separate index space, and a dash-leading regex; and the `Save`/`Load` round trip, since a new key `Save` drops would be protection silently absent.
- `safety.never_auto_actions` added to `configListCommands` and the scalar to `ConfigFields` (both registry tests fail on a stale entry too).

## Gate

`go build`, `gofmt`, `go vet` clean; no new `golangci-lint` findings vs baseline.

Full suite: one failure, `TestALockedMigrationStepLosesTheLeaseAndFailsClosed` in `internal/store/turso` — a package this branch does not touch. It is timing-flaky on this machine: across two consecutive *quiet* runs it failed once and passed once. (`golangci-lint` also panicked once under load with "file requires newer Go version go1.27"; re-run quiet it produced the usual baseline output. Both are the known go1.26.2-linter / loaded-devbox artefacts, not signals about this change.)


internal/daemon/daemon.go
```diff
@@ -1186,6 +1197,31 @@ func (d *Daemon) snapshot() (config.Config, *domain.NeverAutoList, *classify.Cla
...
+// an action screen by accident. The call sites that do want it are the ones
+// that hold a chosen answer — act, the action review, and the LLM promotion —
+// plus the orchestrator's actionScreen, whose "operator" is itself an LLM.
+//
+// Deliberately NOT applied where the outbound text is a generated TASK PROMPT
+// (tasklistreview, generatedTaskUnsafe): these rules describe menu options one
+// must never pick, and matching them against prose an agent is being asked to
+// do would refuse work for containing a phrase. Nor to an operator's own reply
+// in deliverreply — a human choosing to widen a permission is the human's call,
+// which is the same reasoning that leaves the operator path unscreened.
```
0xGosu, line 1214: Auto-accept delivery is not screened by the action rules.

This doc comment enumerates the call sites that want an action screen, and the main unattended send is not among them: `autoAcceptOne` (`internal/daemon/autoaccept.go:427`) calls `autoAcceptDeliver` -> `deliver.Deliver` -> `mcqdeliver.*` without ever calling `actionRefused`.

Scenario: operator adds `[[safety.never_auto_actions]] pattern = "(?i)always allow"`. An agy approval escalates for a non-action reason (`llm_low_confidence`, `below_threshold`, ...) carrying the suggestion `Yes, and always allow for commands that start with 'go' (Persist to settings.json)`. The sweep auto-accepts it and the digit is pressed. `claimBlockedBy` re-asks only the kill switch, the mode and `accept_generated_task`; `ReasonNeverAutoMatch` being in `autoAcceptExcludedReasons` protects only rows escalated *by* an action rule, not rows whose *suggestion* is a widening option. So the exact incident this feature cites still goes through.

The PR body classifies `sweep` as "screens the screen (untouched)", but the sweep's situation screen and the auto-accept delivery are different moments.

Fix is small and uses machinery already in that file: call `d.actionRefused(rec.AgentType, domain.MaterializeForSend(suggestion, rec))` in `autoAcceptDeliver` and return it wrapped in `errOutboundRefused`, which `autoAcceptDeliveryFailed` already treats as a verdict (revert claim, leave pending) rather than a fault that burns attempts.

internal/cli/cli.go
```diff
@@ -1920,7 +1920,36 @@ func rules(ctx context.Context, app *frontend.App, out io.Writer, args []string)
...
+			disabled := make(map[string]bool, len(cfg.Safety.DisabledSeedPatterns))
+			for _, p := range cfg.Safety.DisabledSeedPatterns {
+				disabled[p] = true
+			}
+			for _, r := range domain.SeedNeverAutoActionRules {
+				state := ""
+				if disabled[r.Pattern] {
+					state = " [disabled]"
+				}
+				fmt.Fprintf(out, "seed action %s\t%s%s\t%s\n", domain.SeedRuleID(r.Pattern), r.Kind, state, r.Pattern)
```
0xGosu, line 1938: These `seed action <id>` ids cannot be resolved by any command, and the `[disabled]` column can never become true.

`rules disable-seed` / `enable-seed` go through `domain.SeedRuleByID`, and `App.DisableSeedRule` guards on `domain.IsSeedPattern` -- both iterate `SeedNeverAutoRules()`, which is built from `SeedNeverAutoPatterns` + `SeedHeuristicNeverAutoRules` only. `SeedNeverAutoActionRules` is not in that registry.

So: operator sets `enable_never_auto_action_seeds = true`, runs `hap config rules list`, copies the id printed for `(?i)\bpersist\s+to\s+settings\.json\b`, and `hap config rules disable-seed <id>` fails with `no seed rule "<id>"`. The domain side supports it (`filterDisabled` honours `disabled_seed_patterns`), but nothing can write the entry.

Same root cause: `domain.SeedRuleForcedEscalation` -> `SeedRuleForRationale` also searches only `SeedNeverAutoRules()`, so an escalation forced by an action seed gets no "silence this builtin rule" hint in `hap escalations` (cli.go:1312).

Either extend `SeedNeverAutoRules()`/`SeedRuleByID`/`IsSeedPattern` to cover the action seeds, or don't print ids and a `[disabled]` column for rules that have no toggle.

internal/cli/help.go
```diff
@@ -932,7 +934,21 @@ func buildCommands() {
...
+				"\"operator action\" with their own index space, so they are dropped with\n" +
+				"`remove-action`. --agent-type scopes them the same way, and is optional here.\n\n" +
+				"hap also ships action rules for the scope-WIDENING options (\"and always allow\",\n" +
+				"\"Persist to settings.json\"), which pre-authorise a whole command prefix so later\n" +
+				"commands never raise a prompt at all. They are OFF unless you set\n" +
+				"safety.enable_never_auto_action_seeds=true: refusing an option only escalates, it\n" +
+				"does not pick a narrower one, so turning them on before hap prefers the narrowest\n" +
+				"option would turn most approvals into escalations.\n" +
 				"The pattern may start with a dash (`--force`, `-rf /`) and is taken literally —\n" +
 				"only `--agent-type` is read as a flag, in either position.",
```
0xGosu, line 953: Stale now that `--action` exists: this sentence still says `--agent-type` is the only argument read as a flag and everything else is taken literally.

An operator who wants the literal pattern `--action` will run `hap config rules add --action`, `splitActionFlag` will consume it, and the command fails with a usage error. Worth naming `--action` here and mentioning the `--` escape (`rules add --action -- --action`) that `splitActionFlag`'s own doc comment documents.

internal/domain/neverautoaction_test.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+	}
+	if persist == "" {
+		t.Fatal("the settings.json seed rule is missing")
+	}
+	list, _ := NewNeverAutoActionList(true, []string{persist}, nil)
+	if _, matched := list.Match("agy", "Yes, and always allow (Persist to settings.json)"); !matched {
+		t.Error("control: the other seeds must still fire on this answer")
+	}
+	only, _ := NewNeverAutoActionList(true, []string{persist}, nil)
```
0xGosu, line 100: The "control" here is vacuous: `list` and `only` are constructed with identical arguments, and the first assertion passes through the `(?i)\band\s+always\s+allow\b` seed rather than the persist rule the test claims to be exercising. It would still pass if `filterDisabled` dropped the wrong rule, or every rule but that one.

To discriminate, compare an enabled list against a disabled one on the *same* text that only the persist seed matches, e.g. `"(Persist to settings.json)"` with no "and always allow" in it: matched with seeds on, not matched with the persist pattern disabled.

Minor, same file (~line 46): the `if hit, matched := list.Match("agy", agyApproval); !matched { t.Log(...) }` branch is inverted relative to its comment and asserts nothing on either outcome.

internal/frontend/configcoverage_test.go
```diff
@@ -24,6 +24,7 @@ import (
 var configListCommands = map[string]string{
 	"safety.never_auto_patterns":    "hap config rules add / hap config rules remove",
 	"safety.never_auto_rules":       "hap config rules add --agent-type / hap config rules remove-scoped",
+	"safety.never_auto_actions":     "hap config rules add --action / hap config rules remove-action",
```
0xGosu, line 27: The new key is registered here but not in the shipped operator reference: `.claude/skills/hap/SKILL.md` lists `safety.never_auto_patterns` and `[[safety.never_auto_rules]]` in its list-key table (lines 501-502) with no `[[safety.never_auto_actions]]` row, and its settings table (~line 536) has no `safety.enable_never_auto_action_seeds`.

That file ships inside the binary (`hap --skill`, `hap skill install`), so an agent or operator reading it to find "how do I stop hap picking that option" still finds only the situation lists -- and writes exactly the rule that escalated every agy prompt. Nothing enforces this file in CI, which is why it is easy to miss.

internal/daemon/daemon.go
0xGosu: Fixed in #474 (this PR merged before the findings were applied, so the work is on a new branch off main).

You were right, including about the PR body's classification: the sweep's situation screen and the auto-accept delivery are different moments, and only the first had one. `d.actionRefused(rec.AgentType, outbound)` now runs in `autoAcceptDeliver` inside `WithAgentAutomation`, immediately before the keystrokes, on the **materialized** outbound rather than the stored suggestion so what is judged is what is typed. A hit returns `fmt.Errorf("%w: …", errOutboundRefused, why)` exactly as you suggested — `autoAcceptDeliveryFailed` already reverts the claim and leaves the row pending for it, so a permanent refusal cannot burn attempts and dismiss the escalation at the ceiling.

One ordering detail worth naming: the new `case refused != ""` sits **above** `case !sent`, because a refusal leaves `sent` false and `deliverErr` nil — the generic branch would have returned success.

`TestAutoAcceptScreensTheAnswerItWouldSend` reproduces your scenario: the row is escalated for an ordinary reason (shadow mode) and its suggestion is the widening option. It sweeps `maxAutoAcceptAttempts+1` times and asserts the row is still queued, which is what pins the "verdict, not fault" half. Removing the screen fails it with `sent [2]` — the incident, in a unit test.

Its control (`TestAutoAcceptStillDeliversTheNarrowAnswerBesideAnActionRule`) is deliberate: same rule, same config, same pane, narrow answer, must deliver as the digit. Without it every "nothing was sent" assertion passes on a daemon whose auto-accept is broken outright.

internal/cli/cli.go
0xGosu: Fixed in #474 — and it was both halves you named.

`domain.AllSeedRules` is the union of the situation and action seed sets, and `SeedRuleByID`, `IsSeedPattern` and `SeedRuleForRationale` now resolve through it. So `rules disable-seed <id from the listing>` works, `App.DisableSeedRule`'s `IsSeedPattern` guard lets the entry be written, and an escalation an action seed forced gets the same "silence this builtin rule" hint a situation seed's does (`SeedRuleForcedEscalation` needed no change — `ReasonNeverAutoMatch` is already in its switch).

I took "extend the registry" rather than "stop printing the ids", because the domain side already supported it: both lists read the same `safety.disabled_seed_patterns`, so the only thing missing was a command able to write the entry.

What I did **not** extend is `SeedNeverAutoRules()` itself, and the union is deliberately not what anything compiles from. The situation matcher also carries `SuspectedIrreversible`, and an action rule reaching it would judge a chosen menu option against indicators written for pane content — the exact separation `NeverAutoActionList` exists to keep. `cli.go`'s listing still iterates the situation set alone, so nothing prints twice, and a new test asserts the two sets stay disjoint.

`TestActionSeedIDsFromTheListingResolve` takes the ids from `rules list` **output** rather than from the domain package — the point of the finding is that the two disagreed — then disables every one, checks the `[disabled]` column flips, and enables them again so the column is not a one-way door.

internal/cli/help.go
0xGosu: Fixed in #474. The sentence now reads:

> The pattern may start with a dash (`--force`, `-rf /`) and is taken literally.
> `--agent-type` and `--action` are the only arguments read as flags, in either
> position; to add either of those two words as the literal pattern, put it after
> `--` (`hap config rules add --action -- --action`).

Also added `hap config rules add --action '(?i)always allow'` to the examples, since the help block taught every other verb by example and not that one.

internal/domain/neverautoaction_test.go
0xGosu: Fixed in #474, using your discriminator — after checking it against the real seed set rather than eyeballing it. `"Yes, allow this command (Persist to settings.json)"` is matched by the persist pattern and by none of the other three (`and always allow`, `don'?t ask again`, `allow all|yes to all`), so it discriminates; the text the test used before was matched by two.

The comparison is now enabled-vs-disabled on that one string, with the enabled half as a `t.Fatal` control — without it the "must not match" assertion passes on a list that refuses nothing. Two assertions added beside it: silencing one seed leaves the rest armed (otherwise it is `enable_never_auto_action_seeds` by another name), and an operator rule survives a `disabled_seed_patterns` entry naming the same pattern.

On the `t.Log` branch: agreed it asserts nothing, and I removed it rather than inverting it. Handed the whole screen the list *does* match, because the screen contains the answer — it is content-blind and cannot tell the two apart, so neither outcome is assertable there. What makes the feature work is that it is never HANDED pane text, which only a call-site test can prove; the comment now says that and names `TestAutoAcceptScreensTheAnswerItWouldSend` as the second one.

internal/frontend/configcoverage_test.go
0xGosu: Fixed in #474, and with a guard — your last sentence ("nothing enforces this file in CI, which is why it is easy to miss") seemed like the more useful half of the finding.

SKILL.md now has the `[[safety.never_auto_actions]]` list-key row and the `safety.enable_never_auto_action_seeds` settings row. It also gets a short section on **which kind of rule to reach for**, because the table rows alone would not have prevented the outcome you describe: a reader looking for "how do I stop hap picking that option" needs to be told that an ordinary rule matches the menu that PRINTS the option and escalates every prompt. The auto-accept section says the answer is screened too, since that changed in 464-A.

`TestEveryConfigListIsNamedInTheShippedReference` sits beside the existing coverage tests in this file: every list section in `config.Config` must be named in SKILL.md, with an exemption map that must stay live (three `[llm.*_env]` tables are covered by a wildcard row there). Deleting the new row fails it by name.

Scoped to list sections on purpose — they are the ones an operator cannot discover by guessing a dotted key for `hap config set`. Extending it to every scalar would be a bigger change than this finding, and I have not measured how many are currently missing.

PR #466, opened 2026-09-12
#0 feat: answer an agy approval with the narrowest option, not a widening one

0xGosu:
Finding 4 from `/tmp/agy-findings.md`.

> **Stacked on #464** (the action rules), because the two are the pair the operator asked for: this PR is what makes #464's shipped seeds safe to switch on. Review/merge #464 first; the base will retarget to `main` once it lands.

## The problem

Across this run's audit rows, hap's LLM chose agy's *"Yes, and always allow in this conversation for commands that start with X"* at **confidence 98-99 on nearly every approval**.

That does not approve a command — it pre-authorises a command **prefix** for the rest of the conversation. Every later command matching it runs with no prompt at all, so it never reaches a pane, and therefore never reaches the classifier, the never-auto screen, or the operator. A short prefix (`ln`, `git`, `pwd`) switches the safety net off for a whole class of commands, silently.

## Why a preference and not a refusal

`domain.NarrowestAgyApproval` **substitutes** the narrowest option that still grants the request.

Refusing would only escalate — hap does not then pick a narrower option — so refusal alone turns most approvals into escalations, which is precisely the queue-flooding failure (finding 5) this same run suffered, and `ReasonNeverAutoMatch` is auto-accept-excluded so each one would be permanently operator-only. Substituting answers the prompt and keeps the agent moving. #464's hard refusal stays opt-in beside it, and this is what makes enabling those seeds sane.

## Three bounds

- **Exactly one** granting non-widening option must be offered. With none there is nothing narrower to prefer; with several, which one "satisfies the request" is a judgement labels cannot support, and guessing would answer the prompt differently from anything a human reviewed. Both cases leave the original answer to the ordinary controls.
- **It can never turn an approval into a refusal.** The replacement must be a *granting* option, tested separately because "allow" appears inside denial labels too (`No, deny access to this file`).
- **Widening is anchored on the conjunction** (`and always allow`), never a bare "allow" — a file prompt's plain `Yes, allow access` *is* the narrow answer and must not read as widening.

The chosen answer is resolved through the form's own options first, so it works whether the caller holds a label or a digit — the LLM answers an approval with `select_options`, which arrives as a number.

## Where it runs

Substituted **before** the answer is verified and **before** the audit row is written inside `deliverKeyedFormLLM`, so what is recorded and learned is what was actually typed, with the rationale stating hap made the substitution. Otherwise the audit would read as the LLM having chosen the narrow option itself.

The consult context also tells the LLM to prefer the narrowest option — but that is guidance. A prompt cannot be relied on for a safety property, which is why the deterministic substitution is the actual guard.

## Tests

- `TestAgyLLMWideningChoiceIsAnsweredWithTheNarrowOption` — end to end through the promotion path: the LLM picks option 2 at confidence 99, and the digit that reaches the pane is **1**, with the substitution recorded in the audit.
- `TestAgyLLMNarrowChoiceIsLeftAlone` — the control: an already-narrow answer is delivered unchanged and is *not* reported as substituted.
- Domain cases for label-vs-digit input, the never-turn-an-approval-into-a-refusal rule, the exactly-one-narrow-grant bound, approvals-only, and the widening/granting label predicates (including `Yes, allow access to this file` staying narrow).

## Gate

`go build`, `gofmt`, `go vet` clean; **full `go test -tags "vectors cpu" ./... -count=1` green with zero failures**; no new `golangci-lint` findings vs baseline.


internal/daemon/daemon.go
```diff
@@ -5269,6 +5278,23 @@ func (d *Daemon) handleLLMOutcome(ctx context.Context, res llmOutcome) {
...
+		// picks agy's "and always allow…" row at confidence 98-99 on nearly
+		// every approval, which pre-authorises a command PREFIX for the rest of
+		// the conversation — every later command matching it then runs with no
+		// prompt, so it never reaches a pane, a classifier or this screen again.
+		//
+		// Substituted BEFORE the answer is verified and before the audit row is
+		// written inside deliverKeyedFormLLM, so what is recorded and learned is
+		// what is actually typed. The rationale says so, or the audit would show
+		// the LLM choosing an option it did not choose.
+		if narrow, swapped := domain.NarrowestAgyApproval(form, llmDec.Action); swapped {
```
0xGosu, line 5372: The substitution can never run once `#464`'s seeds are enabled, which is the PR's stated purpose.

`handleLLMOutcome` screens the LLM action through `d.actionRefused(s.AgentType, llmDec.Action)` at line 4964 — far above this hunk. The shipped seed `(?i)\band\s+always\s+allow\b` matches exactly the label this code exists to replace, so with `safety.enable_never_auto_action_seeds = true` the widening answer is rejected with `ReasonNeverAutoMatch` and `return`s long before line 5291 is reached. `ReasonNeverAutoMatch` is in `autoAcceptExcludedReasons`, so each such row is permanently operator-only — the queue-flooding failure the PR body says this change prevents.

The reorder is not free: `form` comes from the post-consult visible re-read (~5080), which itself runs after the screens, so someone has to decide which control wins for an agy approval (e.g. let the substitution run and then re-screen the substituted label). Worth resolving before the seeds are ever switched on. Note also that neither new daemon test enables the seeds, so nothing in the suite covers this interaction.

0xGosu, line 5372: The substitution works exactly once per signature, then turns into an escalation — and the escalation carries the *widening* suggestion.

`deliverKeyedFormLLM` records `ChosenAction: llmDec.Action`, i.e. the substituted narrow label. `domain.Confidence` returns a non-empty `TopAction` from a single decision (`confidence.go:139`), so on the next encounter of the same signature the variance guard at ~4998 sees `conf.TopAction == "Yes, run command"` while the LLM again answers the widening row at 98-99 (the measured behaviour this PR is built on) and rejects with `ReasonVarianceGuard` before line 5291.

Two consequences, both at defaults (no flags):
- every encounter after the first escalates, which is the queue flooding the PR is trying to avoid, arriving by a different door;
- `ReasonVarianceGuard` is *not* in `autoAcceptExcludedReasons`, so under FSP auto-accept the row is claimed and delivered through `deliver.Deliver`, which does no substitution — hap types the widening option after all, unattended.

Worth deciding whether the substituted answer should be recorded as the learned decision, or whether the variance guard should compare post-substitution.

0xGosu, line 5372: The substituted text reaches the pane unscreened.

All three outbound screens — `allow.Match(s.AgentType, llmDec.Action)`, `d.actionRefused(...)` and `allow.SuspectedIrreversible(...)` — ran at lines ~4895-4975 against the *pre-substitution* action. `llmDec.Action` is then replaced here, and nothing re-screens it before `AgyAnswerKey` and `deliverAgyFormLLM`. FR-015's invariant ("an LLM can never smuggle an action past the never-auto screen") is satisfied only for text hap decided not to send.

Scenario: an operator has a never-auto (or never-auto-action) rule naming the narrow grant for a given agent type — e.g. they never want hap answering a file-access approval unattended. The LLM picks the widening row, the screens pass on that text, the substitution swaps in `Yes, allow access`, and the rule that was written precisely to stop it is never consulted.

The fix is local: re-run the three screens on `narrow` immediately after the swap and `reject` on a hit.

internal/domain/agynarrow.go
```diff
@@ -0,0 +1,95 @@
...
+// Substituting the narrow option answers the prompt and keeps the agent moving,
+// which is why this is on by default while the hard refusal
+// (safety.enable_never_auto_action_seeds) stays opt-in beside it.
+
+var (
+	// agyWideningRE matches an option that grants MORE than this one request.
+	// Anchored on the conjunction ("and always allow"), never on a bare
+	// "allow": a file prompt's plain "Yes, allow access" is the narrow answer
+	// and must not read as widening.
+	agyWideningRE = regexp.MustCompile(`(?i)\band\s+always\s+allow\b|\bdon'?t\s+ask\s+(me\s+)?again\b|\bpersist\s+to\s+settings\.json\b|\ballow\s+all\b|\byes\s+to\s+all\b`)
```
0xGosu, line 29: `agyWideningRE` is matched against the RAW label while `AgyGrantingOption` folds through `FoldMenuText` — the asymmetry makes the widening test fail open on typographic apostrophes.

`\bdon'?t\s+ask\s+(me\s+)?again\b` only matches the ASCII spelling. CLAUDE.md records that agents render this row with U+2019 (`don’t ask again`), which is the entire reason `FoldMenuText`/`typographicFolds` exist. A label rendered that way reads as non-widening, so `NarrowestAgyApproval` returns it untouched and hap answers the standing-permission row.

Latent for agy today (its fixtures use `and always allow` with ASCII quotes), but the predicate is exported and reads as general. Folding the label the same way the granting test does costs nothing:

```suggestion
	agyWideningRE = regexp.MustCompile(`(?i)\band\s+always\s+allow\b|\bdon'?t\s+ask\s+(me\s+)?again\b|\bpersist\s+to\s+settings\.json\b|\ballow\s+all\b|\byes\s+to\s+all\b`)
```
(and change `AgyWideningOption` to `return agyWideningRE.MatchString(FoldMenuText(label))`)

internal/domain/agynarrow.go
```diff
@@ -0,0 +1,95 @@
...
+// the caller holds the option's label or its digit — the LLM answers an
+// approval with select_options (a number), while a learned rule holds a label.
+//
+// It substitutes only when exactly ONE granting non-widening option is offered.
+// With none there is nothing narrower to prefer; with several, which one
+// "satisfies the request" is a judgement this cannot make from labels alone, and
+// guessing would answer a prompt differently from what any human reviewed. In
+// both cases the original answer stands and the ordinary safety controls — the
+// action rules among them — decide what happens to it.
+func NarrowestAgyApproval(form AgyForm, chosen string) (string, bool) {
```
0xGosu, line 65: Only one of the agy send paths applies this.

CLAUDE.md states the convention for this shape: "All four send paths route agy forms to the keyed deliverer — `daemon.act`, the LLM promotion, `deliver.Deliver` … A new send path must ask `AgyFormSituation` too." `NarrowestAgyApproval` is called from the LLM promotion alone. `daemon.act` (line 2932 onwards, `deliverAgyForm`) and `deliver.Deliver` → `deliverAgy` both send `dec.Input` / `req.Outbound` verbatim.

Concrete scenario: the live run that motivated this PR already recorded widening choices as decisions at 98-99 confidence. Once such a signature graduates, `act` delivers the widening label directly and this guard never sees it. Likewise, any escalation whose suggestion is a widening row is auto-accepted under FSP through `deliver.Deliver` with no substitution — precisely the mode whose premise is that nobody reads the queue.

If the substitution is a safety property rather than an LLM-output cleanup, it belongs at the shared delivery seam (or at least in `act` and `deliver.Deliver` as well), not only in `handleLLMOutcome`.

PR #468, opened 2026-09-12
#0 fix: stop flooding the escalation queue with hand-out proposals

0xGosu:
Finding 5 from the agy field report: driving a real agy under 0.9.21, the operator's queue filled with the same `send next declared task` proposal over and over. Two independent causes, both fixed here.

## An in-progress list still proposed

The `noop_vs_pending_tasks` escalation fired whenever the source had a pending `[ ]` item — even with another item already `[-]`. `NextDeclaredTask` skips `[-]`, so a list being actively worked still resolves a next task and the conflict escalated, asking the operator to start a *second* task. `DeclaredTask.InProgress` now gates that one branch.

It deliberately does **not** gate the unattended hand-out above it: the idle poll pairs distinct agents with distinct items, so a `[-]` there is another agent's reservation, and suppressing on it would stop a second agent ever being handed work. There is a test for that.

## Repeats were never collapsed

`duplicatePendingEscalation` keys on the pane excerpt, and agy repaints between every background command — so each event presented a new excerpt and slipped straight past the dedup. This reason now latches to one proposal per parked episode (`noopVsPendingRaised`), cleared wherever `idleSince` is and pruned against the live listing.

An **episode latch rather than a cooldown**: no interval to justify, and the happy path self-clears — confirming the escalation delivers the task, the agent works, the latch goes with the episode.

Accepted behaviour change, stated in the doc comment: after the operator *dismisses* the row, nothing re-raises it until the agent works again.

Bounds, each load-bearing:
- Scoped to this one reason — every other escalation on the same parked agent still reaches the operator.
- Below the auto-dismiss guard, so a lifecycle dismissal stays visible in history rather than becoming an opaque suppression.
- Stamped only **after** the audit row exists — a failed insert must not silence the retry for the whole episode.
- No audit row per repeat (`audit_log` is never swept), unlike the excerpt dedup's `ignoreDuplicate`.
- Names itself **once per episode**, not once per event, or the flood just moves into the log.
- Not hung on `idleSince`'s `idleMark`, tempting though that is: the mark is written only by the 60s sweep, so an attention-path proposal in the first minute of a park would have nothing to stamp and the next event would re-raise.

## Deliberately not done

The task also suggested debouncing on parked duration. `idleSince` is *parked-since* and monotonic across a park, so `idleLongEnough` is an age check, not a rate limiter — it cannot debounce a sustained flood, and would delay the first legitimate proposal of every park by 1–2 minutes. The episode latch is what actually collapses repeats.

## Test plan

- `go build -tags "vectors cpu" ./...`, `gofmt`, `go vet` — clean
- Full suite `go test -tags "vectors cpu" ./... -count=1` — 41 packages ok, 0 failures
- `GOTOOLCHAIN=go1.26.2 golangci-lint run --build-tags "vectors,cpu"` — 0 issues
- New tests verified by mutation: reverting each half makes its test fail
  - `TestDecideInProgressListSuppressesTheHandoutProposal` — both halves (in-progress suppresses; the same input without it still escalates, so the #175 guard is not silently deleted)
  - `TestDecideInProgressNeverGatesTheUnattendedHandout`
  - `TestHandoutProposalIsRaisedOncePerParkedEpisode` — repaint suppressed, no audit row, and a proposal raised again after a working→park cycle
  - `TestHandoutLatchNeverSuppressesAnotherReason`

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -345,6 +345,30 @@ type Daemon struct {
 	lastAutoSend map[string]time.Time
 	lastAutoNoop map[string]time.Time
 
+	// noopVsPendingRaised latches the noop-vs-pending proposal to ONE per
+	// parked episode, per agent. duplicatePendingEscalation keys on the pane
+	// excerpt, and an agy repaints between every background command, so a
+	// parked agy mints a new excerpt on every event and the same proposal is
+	// re-raised until the operator's queue is nothing else (finding 5).
+	//
+	// An episode latch rather than a cooldown duration: there is no interval to
+	// justify, and the happy path self-clears — confirming the escalation
+	// delivers the task, the agent goes working, and the latch is dropped with
+	// the episode. Accepted behavior change: after the operator DISMISSES the
+	// row, nothing re-raises it until the agent works again.
+	//
+	// Deliberately not hung on idleSince's idleMark, which would be
+	// episode-scoped for free: that mark is written only by the 60s sweep, so
+	// an attention-path proposal in the first minute of a park would have
+	// nothing to stamp and the next event would re-raise. Cleared wherever
+	// idleSince is, and pruned against the live listing in noteIdleAgents.
+	//
+	// Counts rather than latches so the suppression can name itself exactly
+	// ONCE per episode (notePending's rule): 0 = not raised, 1 = raised, n>1 =
+	// n-1 repeats ignored. A log line per suppressed event would move the very
+	// flood this fixes into the log.
+	noopVsPendingRaised map[string]int
+
 	// preDeliveryReviewInFlight tracks the one live pre-delivery review per
 	// agent; the token lets the outcome handler drop superseded results.
 	// Guarded by mu alongside preDeliveryReviewSeq.
@@ -813,6 +837,7 @@ func New(opt Options) (*Daemon, error) {
 		learnInFlight:             map[string]bool{},
 		lastAutoSend:              map[string]time.Time{},
 		lastAutoNoop:              map[string]time.Time{},
+		noopVsPendingRaised:       map[string]int{},
 		preDeliveryReviewInFlight: map[string]preDeliveryReviewFlight{},
 		rerankInFlight:            map[string]rerankFlight{},
 		rerankCache:               map[string][]domain.RerankResult{},
@@ -1863,7 +1888,10 @@ func (d *Daemon) handleTransition(ctx context.Context, tr domain.AgentTransition
 		d.mu.Lock()
 		delete(d.episodeHandled, tr.PaneID)
 		delete(d.idleSince, tr.AgentID)
+		delete(d.noopVsPendingRaised, tr.AgentID)
 		delete(d.autoTaskClaim, tr.AgentID)
+		// The parked episode is over, so the next one may propose its own
+		// hand-out: confirming the escalation is what set the agent working.
 		// A genuinely new episode: the session-name sync gets its full patience
 		// budget back, and its settle window starts again from the next park.
 		delete(d.sessionSyncDeferred, tr.AgentID)
@@ -2102,6 +2130,7 @@ func (d *Daemon) resetRecycledPaneState(ctx context.Context, a domain.AgentTrans
 	delete(d.lastAutoSend, a.AgentID)
 	delete(d.lastAutoNoop, a.AgentID)
 	delete(d.idleSince, a.AgentID)
+	delete(d.noopVsPendingRaised, a.AgentID)
 	delete(d.autoTaskClaim, a.AgentID)
 	d.forgetSessionRenamePushesLocked(a.AgentID)
 	d.mu.Unlock()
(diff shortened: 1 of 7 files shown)


internal/domain/decide.go
```diff
@@ -416,8 +416,14 @@ func resolveSituation(in DecideInput, conf ConfidenceResult) (candidate, suggest
...
 		// @next_task:declared.
+		//
+		// An item already marked "[-]" is the one exception: the list IS being
+		// worked, so asking the operator to start another task is noise rather
+		// than a conflict. See DeclaredTask.InProgress — it gates this
+		// escalation alone, never the unattended hand-out above.
 		if conf.TopAction == ActionNoop &&
-			in.DeclaredTask != nil && in.DeclaredTask.Task != NoTaskContent {
+			in.DeclaredTask != nil && in.DeclaredTask.Task != NoTaskContent &&
+			!in.DeclaredTask.InProgress {
```
0xGosu, line 426: **High — this doesn't suppress the proposal, it converts it into an autonomous send.**

Adding `!in.DeclaredTask.InProgress` doesn't make this branch return quietly; it makes control fall *past* the #175 guard into the two-tier resolution ~40 lines below, which returns `ActionNextDeclaredTask`. The next branch only catches the fall-through when the noop plurality is operator-backed.

Probed against this branch (`Decide`, idle, `ModeAutonomous`, noop history, `DeclaredTask{Task:"build the parser"}`):

- `sourcedHistory(SourceOperator, …)`, `InProgress:true` → `action="noop"` (intended)
- `sourcedHistory(SourceLLM, …)`, `InProgress:true` → **`action="send"`, input = `"Your next task is build the parser…"`**
- same input with `InProgress:false` → `action="escalate" reason="noop_vs_pending_tasks"`

So on a list that *already has an item in progress* and an LLM-learned noop rule, hap now types a **second** task into the agent unattended, where before it asked the operator. That is strictly worse than the escalation being removed, and it directly inverts the #175 rationale in the comment immediately above ("an autonomous noop rule would then swallow the list with zero escalations") — except now it doesn't swallow, it double-sends.

It ships green because `TestDecideInProgressListSuppressesTheHandoutProposal` builds both halves with `sourcedHistory(SourceOperator, …)`, the one provenance where the next branch happens to absorb the fall-through. The LLM-backed case is untested.

The fix is a design choice rather than a one-liner — either return `ActionNoop` explicitly for the in-progress case, or gate the whole idle resolution — so no suggestion block.

internal/daemon/daemon.go
```diff
@@ -5969,6 +6050,11 @@ func (d *Daemon) declaredTask(ctx context.Context, cfg config.Config, tr domain.
 		// is handed the very same "[ ]" line. Composes with LLMReview above:
 		// the review runs at delivery and reserves the task it actually chose.
 		Reserve: m.src.EnableAutoSendTaskWhenIdle,
+		// Gates the noop-vs-pending ESCALATION only (see the field's doc): a
+		// list already being worked does not need the operator asked to start
+		// another task. Read from the same content NextDeclaredTask parsed, so
+		// the two can never disagree about the list they describe.
+		InProgress: len(domain.InProgressDeclaredTasks(string(m.data))) > 0,
```
0xGosu, line 6057: **Medium — the gate is list-wide and `[-]` is sticky, so this silences proposals it shouldn't.**

`InProgress` is `len(InProgressDeclaredTasks(whole list)) > 0` — any `[-]` anywhere in the source, regardless of which agent holds it or how old it is. Two concrete failures:

1. **Multi-agent.** Agent A is confirmed a hand-out, its item goes `[-]`. Agent B then parks with `[ ]` items still pending on the same source, and the operator is never asked to feed B. This is exactly the reasoning the PR body uses to justify *not* gating the unattended hand-out ("a `[-]` there is another agent's reservation, and suppressing on it would stop a second agent ever being handed work") — it applies unchanged to the escalation path, which is the only path a source **without** `enable_auto_send_task_when_idle` has.

2. **Stale `[-]`.** A confirmed hand-out marks `[-]` and only the agent itself ever moves it to `[x]` (`hap task done`). `reclaimStrandedTasks` returns only *unconfirmed* items, so once the agent went `working` nothing clears it. An agent that does the work and forgets to check the item off leaves the source permanently silenced: `InProgress` stays true forever and the noop-vs-pending proposal never fires again for that list.

If the intent is "this agent is already working an item", the check needs to be agent-scoped (the item reserved to *this* agent, via `task_reservations`/`autoTaskClaim`), not "the list has any `[-]`".

internal/daemon/daemon.go
```diff
@@ -3538,6 +3567,40 @@ func (d *Daemon) deliverActionReviewNoop(ctx context.Context, res actionReviewOu
 }
 
 // escalate records and surfaces an escalation: no input is sent (FR-018).
+// noteNoopVsPendingRaised records that this agent's parked episode now carries
```
0xGosu, line 3570: **Low — `escalate` lost its doc comment to the new helper.**

`noteNoopVsPendingRaised` was inserted between `// escalate records and surfaces an escalation: no input is sent (FR-018).` and `func (d *Daemon) escalate`, so that line is now the *first* line of the new helper's doc comment and `escalate` — the function the whole FR-018 rule hangs off — is undocumented. In a repo where the doc comment is the rationale-of-record, this is worth fixing: move the new helper (and `noopVsPendingSuppressed`) above the `// escalate records…` line, or restore it directly on `func (d *Daemon) escalate`.

PR #474, opened 2026-09-12
#0 fix: screen the unattended send, and make the never-auto action seeds addressable

0xGosu:
Review findings 464-A through 464-E on #464, which merged before they were applied. Same fixes, new branch off `main` — the PR was closed by the merge, so there was nothing left to push to.

Each is verified by mutation: reverting the fix fails the test that names it.

## 464-A — the unattended send was never screened by the action rules

`autoAcceptOne` → `autoAcceptDeliver` → `deliver.Deliver` never called `actionRefused`, and `claimBlockedBy` re-asks only the kill switch, the mode and `accept_generated_task` — none of which looks at the text. Every *other* `actionRefused` call site is a decision made while a human can still see the queue; this is the one that types with nobody watching.

`ReasonNeverAutoMatch` being in `autoAcceptExcludedReasons` does not cover it. That protects rows escalated **by** an action rule, not a row escalated for an ordinary reason — `llm_low_confidence`, `below_threshold` — whose **suggestion** is the widening option. Which is the shape the seeds exist for: hap's LLM chose agy's `Yes, and always allow …` at confidence 98-99 on nearly every approval. So the incident this feature cites still went through.

The screen runs inside `WithAgentAutomation`, immediately before the keystrokes, on the materialized outbound rather than the stored suggestion. A hit is wrapped in `errOutboundRefused`, which `autoAcceptDeliveryFailed` already treats as a verdict: claim reverted, row left pending (FR-015), no attempt burned — so a permanent refusal can never dismiss the escalation at the ceiling.

Removing the screen makes `TestAutoAcceptScreensTheAnswerItWouldSend` fail with `sent [2]` — the incident, reproduced in a unit test. Its control (`…StillDeliversTheNarrowAnswerBesideAnActionRule`) is not optional: without it, every "nothing was sent" assertion passes on a daemon whose auto-accept is broken outright.

## 464-B — the action seed ids resolved to nothing

`rules list` printed a stable id and a `[disabled]` column for each shipped action seed, and neither worked: `SeedRuleByID` and `IsSeedPattern` both searched the **situation** registry, so `hap config rules disable-seed <id copied from the listing>` failed with `no seed rule "<id>"`, and nothing could write the entry that column reads. Same root cause left `SeedRuleForRationale` unable to attribute an action-forced escalation, so those rows got no "silence this builtin rule" hint — the rows that need it most, since the rule refused an *answer* the operator cannot rephrase past.

`domain.AllSeedRules` is the union the three resolvers now use. It is deliberately **not** what anything compiles from, and a test pins that the two sets stay disjoint: the situation matcher also carries the irreversibility heuristic, and an action rule reaching it would judge a chosen menu option against indicators written for pane content. The listing keeps iterating `SeedNeverAutoRules()` alone, so nothing prints twice.

The CLI case takes the ids from `rules list` **output**, not from the domain package — the whole point is that the two agreed.

## 464-C — `rules --help` was stale

It still said `--agent-type` was the only argument read as a flag. Now names `--action` too and the `--` escape `splitActionFlag`'s own doc documents, with an example.

## 464-D — the "control" was vacuous

`TestNeverAutoActionSeedsCanBeSilencedIndividually` built `list` and `only` with identical arguments and asserted on text the `and always allow` seed matched, so it passed whether `filterDisabled` dropped the right rule, the wrong one, or every rule. It now compares **enabled against disabled** on `"Yes, allow this command (Persist to settings.json)"` — checked against all four shipped patterns, it is matched by the persist seed and nothing else — plus an assertion that silencing one seed leaves the rest armed, and that an operator rule survives a `disabled_seed_patterns` entry naming the same pattern.

The inverted `t.Log` branch is gone rather than fixed: the list is content-blind and matches the whole screen because the screen contains the answer, so neither outcome is assertable there. What makes the feature work is that it is never *handed* pane text, which only a call-site test can prove — now named in the comment.

## 464-E — the shipped reference, and a guard for it

`.claude/skills/hap/SKILL.md` ships inside the binary (`hap --skill`, `hap skill install`) and had no `[[safety.never_auto_actions]]` row and no `safety.enable_never_auto_action_seeds`. It now carries both, plus a short section on **which kind of rule to reach for** — written as an ordinary rule, a pattern aimed at a menu option matches the menu that prints it and escalates every prompt, which is the mistake the file was otherwise steering a reader into. The auto-accept section says the answer is screened too.

`TestEveryConfigListIsNamedInTheShippedReference` is the new guard, beside the existing config-coverage tests: every list section in `config.Config` must be named in SKILL.md, with an exemption map that must stay live (three `[llm.*_env]` tables are covered by a wildcard row). Nothing checked that file before — which, as the finding says, is why this was easy to miss. Only list sections are checked: they are the ones an operator cannot discover by guessing a dotted key for `hap config set`.

## Gate

`go build`, `gofmt`, `go vet` clean; **`golangci-lint` 0 issues**. Full unit suite green under `-p 2` except `internal/store/turso`'s `TestALockedMigrationStepLosesTheLeaseAndFailsClosed`, which **also fails at untouched `origin/main`** on this machine and which this branch cannot affect — it changes no file under `internal/store`. Integration suite green, including the `HAP_ITEST_AGY=1` real-agy cases.

---

## Review round 2 (f34b9cb, rebased onto main)

All three findings were right, and the first one I had got actively wrong rather than merely incompletely.

**The screen was in the shared deliverer**, so it also screened the OPERATOR's confirmed answer — which `actionRefused`'s own doc comment rules out. Worse, `MarkAgentActionSideEffect` runs *before* `deliverToPane`, so the refusal left the row side-effect **and** failed with nothing typed: failed at next start rather than replayable, and no way to carry out the decision at all. `autoAcceptDeliver` now takes a `screen func(string) error` called inside the barrier immediately before the keystrokes; the operator seam passes **nil**. A parameter rather than the suggested call-site closure, because the closure moves the last look outside `WithAgentAutomation`, and because `nil` at the seam is the shape `ports.TaskSendHost` and the orchestrator's `actionScreen` already use for this exact distinction.

**A materialized task prompt is not an answer.** `MaterializeForSend` expands the two next-task sentinels into the task text, so the screen was judging prose an agent is being asked to do — and since a refusal never burns an attempt, such a row would revert to pending on every sweep forever. Skipped when `outbound != suggestion`, which is exactly the expansion signal: `MaterializeForSend` returns its argument verbatim otherwise, and `SuggestedAction` has already peeled the source prefix. Same argument #478 made one hop further while this was in review.

**The TUI read the wrong master switch.** The two sides share `disabled_seed_patterns` but not their switch, so an operator with `disable_never_auto_seed_patterns` and the action seeds both set was told "every builtin rule is already off" on an escalation an action seed had just forced. `domain.IsActionSeedPattern` settles the side first. One addition flagged on the thread: an action seed whose own key is off now reads as inactive too, the mirror of the situation side.

### Tests

Each verified by mutation. Restoring the screen on the operator path fails with the reported damage (failed + side-effect + nothing sent); the task-prompt and answer cases are a **pair** over the same shipped `allow all` seed, so neither passes on a screen that refuses everything or nothing; the TUI case reproduces the reported message verbatim, with the situation-seed row as its control. The TUI affordance had no test before.

`SKILL.md` also updates the action seeds' rationale now that #466 substitutes the narrowest agy option, and says which sends they screen.

### Gate (re-run on the rebased base)

`go build`, `gofmt`, `go vet` clean; **`golangci-lint` 0 issues**. Integration suite green, including the `HAP_ITEST_AGY=1` real-agy cases. `internal/domain`, `internal/tui`, `internal/cli`, `internal/frontend` green.

`internal/daemon` is green per-test but fails one *different* test on each whole-package run under load (three runs, three different tests: `TestConsultCarriesTheAgentsCwd`, `TestSessionSyncPushRefusesAWorkingAgent`, `TestIdleTaskSourceMatchesAgentShortName` — none in a file this branch touches, each passing alone and on repeat). `internal/store/turso`'s `TestALockedMigrationStepLosesTheLeaseAndFailsClosed` also fails locally and does so at untouched `origin/main`; this branch changes no file under `internal/store`.


internal/daemon/autoaccept.go
```diff
@@ -576,10 +576,36 @@ func (d *Daemon) claimBlockedBy(ctx context.Context, rec *domain.AuditRecord,
...
+// which autoAcceptDeliveryFailed already reverts the claim for and leaves
+// pending (FR-015 — a never-auto match always reaches a human), rather than
+// burning an attempt per sweep until the row is dismissed.
 func (d *Daemon) autoAcceptDeliver(ctx context.Context, rec *domain.AuditRecord, suggestion string) error {
 	var deliverErr error
+	refused := ""
 	sent := false
 	disabled, err := d.opt.Store.WithAgentAutomation(ctx, rec.AgentID, func() {
+		outbound := domain.MaterializeForSend(suggestion, rec)
+		if why := d.actionRefused(rec.AgentType, outbound); why != "" {
```
0xGosu: **The operator's own confirmed reply now goes through this screen too.** `autoAcceptDeliver` has a second caller: `deliverreply.go:117` (`deliverToPane` ← `deliverReply`), which is the path an operator's `hap confirm --send` / TUI confirm takes. Putting `actionRefused` inside this function therefore screens a human's explicit answer, which both `actionRefused`'s own doc comment and CLAUDE.md say must never happen ("Nor to an operator's own reply in deliverreply — a human choosing to widen a permission is the human's call"; "The operator path is deliberately NOT screened").

Scenario: with `safety.enable_never_auto_action_seeds = true` (or any operator `--action` rule), an operator looks at an agy approval, decides deliberately to pick "Yes, and always allow commands starting with 'go'", and confirms it. The queued `deliver_reply` action is refused with `errOutboundRefused` and fails — and because `MarkAgentActionSideEffect` is already written at `deliverreply.go:81` before `deliverToPane`, the row is marked side-effect AND failed although nothing was typed, so it is failed-at-next-start rather than replayable. The operator has no way to carry out their own decision.

The fix is to screen at the auto-accept call site rather than inside the shared deliverer — e.g. do it in `autoAcceptOne`'s `deliver := func()` closure (or pass a flag), so `deliverReply`'s deliberate `screenOutbound`-only gate is preserved.

internal/daemon/autoaccept.go
```diff
@@ -576,10 +576,36 @@ func (d *Daemon) claimBlockedBy(ctx context.Context, rec *domain.AuditRecord,
...
+// what is typed. A hit is a VERDICT, not a fault: wrapped in errOutboundRefused,
+// which autoAcceptDeliveryFailed already reverts the claim for and leaves
+// pending (FR-015 — a never-auto match always reaches a human), rather than
+// burning an attempt per sweep until the row is dismissed.
 func (d *Daemon) autoAcceptDeliver(ctx context.Context, rec *domain.AuditRecord, suggestion string) error {
 	var deliverErr error
+	refused := ""
 	sent := false
 	disabled, err := d.opt.Store.WithAgentAutomation(ctx, rec.AgentID, func() {
+		outbound := domain.MaterializeForSend(suggestion, rec)
```
0xGosu, line 642: **The screened `outbound` can be a generated TASK PROMPT, which `actionRefused` documents as deliberately out of scope.** `MaterializeForSend` expands `ActionNextDeclaredTask` / `ActionNextInferredTask` into the task text itself (`domain/suggestion.go:84-93`), so this screen is applied to prose an agent is being asked to do — exactly what `actionRefused`'s doc says it must not be ("Deliberately NOT applied where the outbound text is a generated TASK PROMPT … matching them against prose an agent is being asked to do would refuse work for containing a phrase").

Scenario: with the action seeds armed, a checklist item reading "Update the CORS config so we allow all origins in dev" materializes into an outbound matching the shipped seed `(?i)\ballow\s+all\b`. Since `errOutboundRefused` is a verdict that never burns an attempt, the row is reverted to pending on every sweep forever — never delivered, never dismissed, just an INFO line per (row, reason).

The code already has the signal to discriminate: skip the action screen when `outbound != suggestion` (i.e. a sentinel was materialized into prose).

internal/domain/safety.go
```diff
@@ -225,7 +246,12 @@ func SeedRuleForRationale(rationale string) (NeverAutoRule, bool) {
 	if !ok || source != string(NeverAutoSeed) {
 		return NeverAutoRule{}, false
 	}
-	for _, r := range SeedNeverAutoRules() {
+	// Both sides: an escalation forced by an ACTION seed carries the same
+	// [never_auto_match] reason and the same diagnostic shape, so searching the
+	// situation set alone left those rows with no "silence this builtin rule"
+	// hint at all — the one case where the operator most needs it, since the
+	// rule refused an answer rather than a screen.
+	for _, r := range AllSeedRules() {
```
0xGosu, line 274: **Widening this resolver makes the TUI's "silence this builtin rule" affordance reachable for action seeds — where it now tells the operator a falsehood and refuses to act.**

`SeedRuleForRationale` is what `tui.go:5893` and `SeedRuleForcedEscalation` → `disableMatchedSeedRulePrompt` (`tui.go:7095`) resolve through. With this change an escalation whose rationale is `[never_auto_match] outbound action: pattern … (source=seed …)` (written at `daemon.go:2966`) resolves to an action seed. But `seedRuleDisabled` (`tui.go:7062`) short-circuits on `safety.disable_never_auto_seed_patterns` for *any* pattern, while the action seeds are compiled from `EnableNeverAutoActionSeeds` alone (`daemon.go:1128`) and that master switch does not touch them.

Scenario: an operator has `disable_never_auto_seed_patterns = true` and `enable_never_auto_action_seeds = true`. An action seed blocks an answer; in the TUI they press `b` on the escalation and get "builtin rule <id>: every builtin rule is already off (safety.disable_never_auto_seed_patterns)". Nothing is written to `disabled_seed_patterns`, the rule is still armed, and the same escalation repeats — with the UI insisting it is already disabled. This PR's own SKILL.md text says the opposite ("the action seeds are governed by their own key above").

The CLI path is fine (`frontend.DisableSeedRule` guards on `IsSeedPattern` only) and this PR tests it; the TUI affordance the same widening enables has no test, which is why it is easy to miss. `seedRuleDisabled` needs to know which side the rule is on before applying the master switch.

0xGosu (commented):
Careful pass over the non-test changes (the action screen, the three widened seed resolvers, the docs). Three findings, all about reach the change acquires beyond its intent — the shared deliverer serving the operator path, the materialized sentinel being prose, and the widened rationale resolver landing in a TUI affordance written for situation seeds only. Details inline.

internal/daemon/autoaccept.go
0xGosu: Confirmed and fixed in f34b9cb. You are right on both halves, and the second one is the part I had not seen: `MarkAgentActionSideEffect` at `deliverreply.go:81` runs before `deliverToPane`, so the refusal did not merely block the operator's answer — it left the row side-effect AND failed, which is failed-at-next-start rather than replayable. There was no way to carry the decision out at all.

`autoAcceptDeliver` now takes a `screen func(string) error`, called inside the barrier immediately before the keystrokes; `deliverToPane` passes **nil**, with the reasoning in its doc comment so the next person does not re-add it.

**A parameter rather than the closure you suggested**, for two reasons — the alternative you offered ("or pass a flag") in a shape this repo already uses:

- screening in `autoAcceptOne`'s `deliver` closure moves the last look outside `WithAgentAutomation`, and this is a content gate whose whole value is being the final read before the keystrokes;
- `nil` at the seam is the same idiom as `ports.TaskSendHost`'s `screen` and the orchestrator's `actionScreen` returning nil for a non-orchestrator author — "this caller is not screened" stated at the call site rather than inferred from a bool.

`TestTheOperatorsOwnWideningAnswerIsNotScreened` covers it: an operator confirms exactly the widening option, the action must reach `done`, the digit must be typed, and the row must not be left side-effect+failed. Restoring the screen inside `autoAcceptDeliver` fails all three, with your scenario's message verbatim. The discriminating control is `TestAutoAcceptScreensTheAnswerItWouldSend` — same rule, same answer, refused on the unattended path — and asserting the digit rather than only "no error" is the other half, since a daemon that delivers nothing also refuses nothing.

internal/daemon/autoaccept.go
0xGosu: Confirmed and fixed in f34b9cb, using your discriminator.

I verified the inequality is exact before relying on it: `MaterializeForSend` returns its argument **verbatim** except for the two sentinels (`suggestion.go:84-93`), and `SuggestedAction` has already peeled the source prefix (`StripSourcePrefix` at `suggestion.go:22`), so an ordinary `respond: Yes` arrives as `Yes` and materializes to `Yes`. `outbound != suggestion` therefore means exactly "a sentinel became prose" and nothing else — worth stating, because had `SuggestedAction` not peeled, the skip would have silently disabled the screen everywhere while every test stayed green.

I kept the inequality rather than naming the two sentinels: a future sentinel that expands the same way is then covered without touching this, and the comment says that is what it means.

The screen moved into `autoAcceptActionScreen` (it had to become a callable for 464-A anyway), so both exemptions live in one place with the reasoning.

`TestTheActionScreenSkipsAMaterializedTaskPromptButNotAnAnswer` is your scenario as a pair: your CORS/`allow all` checklist item, and an ANSWER carrying the same phrase, both against the armed shipped `\ballow\s+all\b` seed. The pair is the point — either alone passes on a screen that refuses everything, or one that refuses nothing.

Also worth noting the timing: #478 landed the same argument one hop further while this was in review ("It is the same argument `actionRefused` already makes for a generated task prompt"), so this is now consistent with main rather than a local exception.

internal/domain/safety.go
0xGosu: Confirmed and fixed in f34b9cb — and you are right that this was a consequence of the widening rather than a pre-existing bug, so it belongs in this PR.

`seedRuleDisabled` now settles which side the rule is on before reading any master switch, via a new `domain.IsActionSeedPattern`. The two sides share `disabled_seed_patterns` but not their switch, so reading the situation one for an action rule reports the opposite of the truth.

**One addition beyond the finding, flagged rather than slipped in:** an action seed whose own key is OFF now reads as inactive, with the message naming `enable_never_auto_action_seeds`. That is the mirror of the situation side's master switch — it silences by being *on*, this one by being *off* — and it changes `seedRuleStateSuffix` output for a case you did not raise: an OLD escalation raised while the key was still set. That is exactly what `seedRuleStateSuffix`'s own doc comment exists for ("an OLD escalation … does not read as a live block the operator still has to clear"), so leaving it out would have been a second wrong answer in the same function. Say the word if you would rather it stayed narrow.

Two tests, where there were none:

- `TestTheSeedMasterSwitchesAreReadPerSide` is your exact scenario — both keys set, two rows, one forced by an action seed and one by a situation seed. `b` must prompt on the first and must still report the master switch on the second. That situation row is the control: without it the test passes on a `seedRuleDisabled` that has simply stopped reporting anything. Reverting the fix fails it with your message verbatim.
- `TestAnUnarmedActionSeedReadsAsInactive` covers the addition above.

Both build their fixture from the live seed list (`anActionSeedRule`, matching the existing `aSeedRule`), so they keep testing a real rule after the seed set changes rather than silently asserting on nothing.

PR #476, opened 2026-09-12
#472 fix: take agy composer proof before taskfile.ApplyReview

### LINKED ISSUE #472: Pre-delivery task review reserves before the agy composer proof, so a parked agy churns the checklist per sweep
Split out of the #461 review (finding 461-D), where it was raised against the wrong location and is not a blocker for that PR.

## What happens

`deliverDeclared` → `startTaskListReview` → `taskfile.ApplyReview` reserves the checklist item **before** `deliverAutonomousClaimed` runs, so `del.rollback` arrives already set. `deliverAutonomousClaimed` then takes its opening agy composer proof (`internal/daemon/daemon.go`, `agyComposerRefusal`) and `abandon()`s on a refusal, which calls that caller-owned rollback.

For an agy agent parked in a modal — a standing approval, a picker, the survey, or the operator's own draft — that is **reserve → refuse → release on every idle poll**: two task-list writes per minute, which under a `github_gist` source are two remote read-modify-writes. `agyComposerRefusal` fails closed on a read error, so a pane herdr cannot read produces the same loop.

Nothing counts these, deliberately: no `RecordTaskHandoutAttempt`, because nothing was typed and the item is still perfectly deliverable. Burning the `maxTaskHandouts` ceiling on a merely busy pane would end in `escalateUndeliverableTask` leaving `[-]` for a task no agent ever refused.

## Scope

- Reachable only with `enable_llm_review_before_auto_send` set for the source — a per-source opt-in, off by default.
- Pre-existing: the opening proof is #454's. #461 only added the *second* proof, below the reservation, and [its own path is covered](https://github.com/0xGosu/herdr-auto-pilot/pull/461) by `TestAgyApprovalIsRefusedBeforeAnythingIsReserved`, which proves an approval and an unreadable pane cost no task-list write and no audit row when the delivery does its own reserving.

## Likely fix

Take the agy composer proof **before** `taskfile.ApplyReview`, not after — the review is an LLM round trip, so proving the pane first is also cheaper. That cannot remove the last look before the send: the reservation still happens between any earlier proof and the keystrokes, which is what the last look exists for.

## Guard to add with it

A test in the shape of `TestAgyApprovalIsRefusedBeforeAnythingIsReserved` but driving the review path with a caller-owned claim, asserting zero `MutateTaskFile` calls for an agy pane parked in a modal.

0xGosu:
Closes #472

### Summary
In the pre-delivery task-list review path, `handleTaskListReviewOutcome` previously invoked `taskfile.ApplyReview` to reserve the checklist item before handing the delivery down to `deliverAutonomousClaimed`. When the target agy agent was parked in a modal (standing approval, picker, draft) or unreadable, `deliverAutonomousClaimed`'s opening `agyComposerRefusal` proof failed, triggering `abandon()` $\to$ `reviewRollback` to release the item back to `[ ]`. This resulted in a reserve $\to$ refuse $\to$ release loop on every sweep (two task-list mutations per poll).

### Fix
- Proves agy composer readiness (`d.agyComposerRefusal`) in `handleTaskListReviewOutcome` before calling `taskfile.ApplyReview`.
- If refused, updates decision status to rejected, drops the auto task claim, and returns immediately without mutating the checklist.
- Preserves the last-look composer proof in `deliverAutonomousClaimed` before keystrokes are sent.
- Adds `TestAgyReviewApprovalIsRefusedBeforeAnythingIsReserved` in `internal/daemon/agy_test.go` guarding against task file mutations on parked/unreadable agy panes.

### DIFF
--- a/internal/daemon/tasklistreview.go
+++ b/internal/daemon/tasklistreview.go
@@ -376,15 +376,15 @@ func (d *Daemon) handleTaskListReviewOutcome(ctx context.Context, res taskListRe
 	// standing approval, a picker or the operator's draft. Proving the pane
 	// before taskfile.ApplyReview ensures a refusal costs no task-list write
 	// (and no remote read-modify-write under a gist source).
+	//
+	// Unlike out.Noop below — which commits edits because the source is
+	// genuinely exhausted and no delivery was requested — a busy pane is a
+	// transient refusal of a delivery attempt: applying edits here would still
+	// cost remote read-modify-writes on every idle sweep while the modal stands,
+	// and risks drifting the list out from under a human working in that modal.
 	if domain.IsAgy(s.AgentType) {
 		if err := d.agyComposerRefusal(ctx, s.PaneID); err != nil {
-			slog.Info("agy composer not ready; not sending", "agent", s.AgentID, "reason", err)
-			if res.decision != nil {
-				if uerr := d.opt.Store.UpdateLLMDecisionStatus(ctx, res.decision.ID, "rejected"); uerr != nil {
-					slog.Error("llm decision status update failed", "error", uerr)
-				}
-			}
-			d.dropAutoTaskClaim(s.AgentID)
+			d.standDown(ctx, res, "agy_composer_not_ready", err.Error(), llmConf, proposal, now)
 			return
 		}
 	}
@@ -537,13 +537,15 @@ func (d *Daemon) handleTaskListReviewOutcome(ctx context.Context, res taskListRe
 }
 
 // standDown abandons a review without sending anything — not the reviewed task
-// and not the original either. It exists for the kill switch, which is the one
-// outcome where fail-open is wrong: the operator asked the daemon to stop, so
-// "keep the agent working" is exactly what they did not want.
+// and not the original either. It exists for outcomes where fail-open is wrong:
+// the kill switch is active, or the target pane is busy/unready (e.g. an agy
+// composer modal) and sending the unreviewed original would violate the same
+// barrier.
 //
-// It still does not escalate. There is nobody to ask — the operator already
-// said "stop" — so an escalation would only add a row to answer later for
-// something that needs no answer; the audit row is what tells them it happened.
+// It still does not escalate. There is nobody to ask — the operator paused or
+// the pane is temporarily occupied — so an escalation would only add a row to
+// answer later for something that needs no answer; the audit row is what tells
+// them it happened.
 func (d *Daemon) standDown(ctx context.Context, res taskListReviewOutcome,
 	reason, why string, llmConf *int, proposal string, now time.Time) {
 
--- a/internal/daemon/tasklistreview.go
+++ b/internal/daemon/tasklistreview.go
@@ -371,6 +371,24 @@ func (d *Daemon) handleTaskListReviewOutcome(ctx context.Context, res taskListRe
 		return
 	}
 
+	// agy only ever receives typed text at a proven EMPTY composer: herdr reports
+	// its modals idle, so an unattended send could type a hand-out into a
+	// standing approval, a picker or the operator's draft. Proving the pane
+	// before taskfile.ApplyReview ensures a refusal costs no task-list write
+	// (and no remote read-modify-write under a gist source).
+	if domain.IsAgy(s.AgentType) {
+		if err := d.agyComposerRefusal(ctx, s.PaneID); err != nil {
+			slog.Info("agy composer not ready; not sending", "agent", s.AgentID, "reason", err)
+			if res.decision != nil {
+				if uerr := d.opt.Store.UpdateLLMDecisionStatus(ctx, res.decision.ID, "rejected"); uerr != nil {
+					slog.Error("llm decision status update failed", "error", uerr)
+				}
+			}
+			d.dropAutoTaskClaim(s.AgentID)
+			return
+		}
+	}
+
 	// The safety re-gate. The reviewer is an LLM authoring both task text and
 	// the choice of task, so its output is screened exactly like any other
 	// LLM-authored outbound (FR-015). It runs INSIDE the mutation, against the
--- a/changelog.d/fix-review-reserves-before-proof.md
+++ b/changelog.d/fix-review-reserves-before-proof.md
@@ -1,0 +1,1 @@
+- Fixed pre-delivery task review taking the agy composer proof before reserving the checklist item rather than after: an agy agent parked in a modal (or a pane that cannot be read) no longer churns the task list with reserve-refuse-release writes on every sweep.
--- a/internal/daemon/agy_test.go
+++ b/internal/daemon/agy_test.go
@@ -689,7 +689,6 @@ func TestAgyApprovalIsRefusedBeforeAnythingIsReserved(t *testing.T) {
 			ctx := context.Background()
 			h := newHarness(t, "")
 			h.herdr.setPane(tc.pane)
-			h.herdr.setAgents([]domain.AgentTransition{agyAgent()})
 			h.herdr.failRead = tc.failRead
 			mutations := 0
 			h.daemon.opt.MutateTaskFile = func(path string, fn func(string) (string, error)) error {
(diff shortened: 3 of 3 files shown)


0xGosu:
### Review Follow-up: Staged LLM Decision Status & Resource Lifecycle

We investigated the codebase to determine the role and blast radius of marking the staged decision as `"rejected"` via `UpdateLLMDecisionStatus` during the early return when an agy composer is busy/parked in a modal.

---

#### 1. What does `llm_decisions.status = 'rejected'` feed?
**It is strictly inert lifecycle bookkeeping for staging IPC between the MCP subprocess and the daemon.** It does **not** feed:
- **Signature promotion or demotion / graduation counters**: Promoted/demoted signature state lives in the `signatures` table (`internal/domain/graduation.go:59`), calculated strictly from historical rows in the `decisions` table via `Store.DecisionsForSignature` (`internal/store/store.go:1151`). `llm_decisions` is never joined or queried for signature evaluation.
- **Confidence calculation**: `domain.Confidence` evaluates decisions from the `decisions` table (`internal/domain/confidence.go`). It never reads `llm_decisions`.
- **Learning**: Statistical learning is recorded via `Store.RecordDecision` into `decisions` (`internal/daemon/daemon.go:3452`), which only executes *after* delivery has succeeded at the pane. Refused deliveries never record a decision. User correction learning writes to `AUTO.md` via `llm.learn_from_user_command` upon operator escalation resolution, not from staged consult rows.
- **Audit trail**: Operator-facing audit history lives in `audit_log` (`internal/store/store.go:449`). `audit_log` records no reference to `llm_decisions.status`.

**Where `llm_decisions.status` actually matters:**
1. **Retention pruning (`Store.PruneAgedRows`, `internal/store/retention.go:160-166`)**:
   ```sql
   DELETE FROM llm_decisions WHERE node_id = ? AND status != 'pending' AND created_at < ?
   ```
   Rows with `status != 'pending'` are eligible for pruning once aged. Rows left `'pending'` are explicitly exempt from retention pruning to protect in-flight consults.
2. **Stale sweep (`Daemon.expireStaleLLMWork`, `internal/daemon/daemon.go:5931-5939`)**:
   Scans `status = 'pending'` decisions via `Store.PendingLLMDecisions` to expire dangling consults that died mid-flight.

---

#### 2. Candidate Behaviours: Keep `"rejected"` vs. Leave `"pending"` / Re-queue
**Verdict: Keep `"rejected"`.**
- **Consistency**: Marking `"rejected"` on non-delivery is the established invariant across all daemon delivery paths:
  - Normal review delivery non-send (`internal/daemon/tasklistreview.go:527-536`): `if !delivered { status = "rejected" }`
  - Normal review fallback (`internal/daemon/tasklistreview.go:309`): `"rejected"`
  - Action reviews (`internal/daemon/daemon.go:3638`): `"rejected"`
  - Autonomous actions (`internal/daemon/daemon.go:4942`): `"rejected"`
  - Form consults and task generations (`internal/daemon/keyedform.go:435`, `internal/daemon/sweep.go:601`): `"rejected"`
- **Why leaving `"pending"` is harmful**:
  - It would prevent `PruneAgedRows` from pruning the row, leaking SQLite records until `expireStaleLLMWork` eventually touches it (up to `2 * longest`).
  - It violates the single-flight contract of staged consult rows.
- **Why re-queuing is unnecessary and undesirable**:
  - `llm_decisions` is an ephemeral RPC staging row, not a delivery queue.
  - The daemon runs continuous idle sweeps. When the agy modal is dismissed and the composer clears, the next idle sweep captures the live pane situation and evaluates the task list afresh. Trying to replay a stale consult against a future pane state risks executing on drifted context.

---

#### 3. Resource Cleanup Verification
The early return skips nothing and releases all resources cleanly:
- **Flight cancellation**: `fl.cancel()` was called prior to the check (`internal/daemon/tasklistreview.go:290`).
- **Flight registry**: `d.preDeliveryReviewInFlight` entry for `s.AgentID` was removed (line 284).
- **LLM request status**: `Store.UpdateLLMRequestStatus(..., "done")` was already executed (line 291).
- **Pairing claim**: `d.dropAutoTaskClaim(s.AgentID)` is explicitly called (line 387), matching `deliverAutonomous:3264`.
- **Checklist mutation/reservation**: Because the check runs before `taskfile.ApplyReview` and `MutateTaskFile`, the checklist was never modified or reserved `[-]`, so no rollback is required (eliminating the rollback churn from issue #472).
- **Audit**: `deliverAutonomousClaimed` checks `agyComposerRefusal` before `AppendAudit` (`internal/daemon/daemon.go:3308-3321`), so neither path records an audit event for a refused send on a busy pane.
- **Deferred cleanup**: No lingering goroutines, locks, or context leaks exist.

No code modifications to PR #476 are necessary.

internal/daemon/tasklistreview.go
```diff
@@ -371,6 +371,24 @@ func (d *Daemon) handleTaskListReviewOutcome(ctx context.Context, res taskListRe
...
+			slog.Info("agy composer not ready; not sending", "agent", s.AgentID, "reason", err)
+			if res.decision != nil {
+				if uerr := d.opt.Store.UpdateLLMDecisionStatus(ctx, res.decision.ID, "rejected"); uerr != nil {
+					slog.Error("llm decision status update failed", "error", uerr)
+				}
+			}
+			d.dropAutoTaskClaim(s.AgentID)
+			return
+		}
+	}
```
0xGosu, line 390: **The refusal writes no audit row.** This block is `standDown` minus the audit — but the function's own doc says every non-send branch is audited "so a silent fallback never looks like an ordinary send", and the kill-switch sibling 8 lines above goes through `d.standDown`, which writes `auditTaskReview` under a distinct reason.

Scenario: agy sits in a standing approval. A full review consult runs, scores above threshold, and is discarded here with only an `slog.Info` line. `hap audit` shows nothing — the operator sees a consult spent and a proposal thrown away with no record.

`llmConf` and `proposal` are already in scope, so this is a drop-in:

```suggestion
	if domain.IsAgy(s.AgentType) {
		if err := d.agyComposerRefusal(ctx, s.PaneID); err != nil {
			d.standDown(ctx, res, "agy_composer_not_ready", err.Error(), llmConf, proposal, now)
			return
		}
	}
```

internal/daemon/tasklistreview.go
```diff
@@ -371,6 +371,24 @@ func (d *Daemon) handleTaskListReviewOutcome(ctx context.Context, res taskListRe
 		return
 	}
 
+	// agy only ever receives typed text at a proven EMPTY composer: herdr reports
+	// its modals idle, so an unattended send could type a hand-out into a
+	// standing approval, a picker or the operator's draft. Proving the pane
+	// before taskfile.ApplyReview ensures a refusal costs no task-list write
+	// (and no remote read-modify-write under a gist source).
```
0xGosu, line 378: **Gating here discards the review's checklist EDITS, not just the reservation.** Before this change, `taskfile.ApplyReview` applied the accepted `TaskActions` and only the `[-]` reservation was rolled back on a composer refusal. Now a high-confidence review that drops an invalid item or rewords one is thrown away wholesale because the pane happens to be in a picker.

While the agent stays in a modal (minutes to hours), every sweep re-consults the LLM and re-computes the same edits, and the list never converges — even though applying the edits needs no pane at all. The `out.Noop` branch below states the opposite policy explicitly: "The edits stand (dropping an invalid task IS the work)."

Defensible as the intent of the fix (an unready composer costs no remote read-modify-write), but the divergence from the sibling branch is worth a line in the comment — or consider applying the edits with `Reserve` forced off and skipping only the reservation.

internal/daemon/tasklistreview.go
0xGosu: Implemented. Replaced the manual status update with `d.standDown(ctx, res, "agy_composer_not_ready", err.Error(), llmConf, proposal, now)`. This records the review audit row with `domain.AuditActionTaskReviewFailed` and reason `agy_composer_not_ready`, sets `llm_decisions.status = "rejected"`, and drops the auto task claim, matching the kill-switch standDown path. Added test assertion in `TestAgyReviewApprovalIsRefusedBeforeAnythingIsReserved` verifying the audit row is written with `[agy_composer_not_ready]`.

internal/daemon/tasklistreview.go
0xGosu: Documented the divergence in the comment. Unlike `out.Noop` (which commits edits because the source is genuinely exhausted and no delivery was requested), a busy/parked pane represents a transient refusal of a delivery attempt. Gating before `ApplyReview` ensures that unattended idle polls never incur task-list writes or remote read-modify-writes (under gist sources) while an agent is parked in a modal, and avoids mutating the checklist out from under a human working in that modal.

PR #501, opened 2026-09-12
#0 feat: install the hap and hap-orchestrator skills for the orchestrator

0xGosu:
> **Stacked on #500** (`docs/hap-orchestrator-skill`), which is where
> `.claude/skills/hap-orchestrator/` comes from. Review that one first; the base
> retargets to `main` once it merges. Please don't rebase it away.

## The problem

The orchestrator's brief opens with "run `hap --skill` and `herdr --skill`". That
is the one instruction an **automatic context compaction** takes away: once the
brief is gone, the session has no route back to the documentation it was set up
from, and nothing re-sends it.

## What this does

Puts both skills on disk in the orchestrator's working directory, where the
session's own skill loader finds them at any time.

- **`skillbundle.go`** (package `skilldoc`) embeds the `hap-orchestrator` **tree**
  as a separate `embed.FS` — `HapSkill` stays a `string`. The skill is a
  `SKILL.md` plus six reference files it links **by relative name**, so the
  single-document embed the `hap` skill uses would have shipped it broken. The
  `all:` prefix is load-bearing: a plain directory pattern drops `.`/`_` children,
  so a later `.gitkeep` or `_draft.md` would stop shipping silently.
- **`AgentSkillsDir(kind)`** is a pure map — `.claude/skills` for
  `domain.OrchestratorAgentKind`, `.agents/skills` for anything else.
  `Targets()` is untouched: it is *home*-relative and serves `hap skill install`,
  and its own comment records that `~/.agents` is not read globally.
- **`InstallAgentSkills(root, kind)`** writes
  `<root>/<dir>/hap/SKILL.md` plus the whole `hap-orchestrator` tree, returning
  only the paths it actually changed.
- **`daemon.bootstrapOrchestratorSkills`** runs from `ensureOrchestrator`, after
  `orchestratorPermitted` and after `OrchestratorLaunch` resolves the kind, and
  **before the alive/launch branch**. That placement is the point: `orchestratorDir`
  is only ever reached from `if !found {`, so an adopted session, a daemon restart
  and — the case this exists for — **an upgrade** never saw it. `Run`'s startup pass
  passes a nil listing, which is what defeats `startOrchestratorPass`'s
  "healthy, nothing to do" early-out, so every daemon start refreshes them for free.

## Two decisions, stated plainly

**An operator-edited copy IS refreshed, not left alone.** hap owns
`<state>/orchestrator`, it re-creates that directory, this is the same bargain
`hap skill install` has always made (`TestInstallToOverwritesAnExistingInstall`),
and a skill left stale across an upgrade is exactly the failure this prevents.
Nothing is discarded silently: the refresh is logged, and the protected route is
the second decision.

**Nothing is written into a `full_self_prompting.orchestrator_agent_cwd` you
chose.** That is the seam `orchestratorDir` already draws — hap does not create
that directory either ("a typo must not silently make a directory") — and here it
also keeps an *unattended* write out of somebody's repository, where a
`.claude/skills/hap/SKILL.md` could overwrite a project skill of their own and
turn up in their `git status`. hap logs that once, naming `hap skill install`.
That is also where an operator who wants a hand-written orchestrator skill puts it.

## Properties

- **Idempotent by content**: a file whose bytes already match is left alone, so a
  steady state costs reads and no writes (asserted on mtime).
- **Never fatal**: a failure logs once per distinct message and the pass carries
  on to the start and the brief. A herd with no skills on disk is much better off
  than one with no orchestrator.
- **Default kind unchanged.** The `.agents/skills` branch is written and tested but
  **dormant** — `domain.OrchestratorLaunch` still refuses any kind but `claude`.
- **Accepted limit, documented on the function**: an *adopted* session runs in a
  directory hap did not choose and may not write to, so its skills land in
  `<state>/orchestrator` where it will not look. The alternative is writing into an
  operator's directory, which is the one thing this refuses to do.

## Surfaces updated

The built-in brief's step 1, `.claude/skills/hap-orchestrator/SKILL.md`'s setup
block, `.claude/skills/hap/SKILL.md` (orchestrator section + the install section),
`README.md`, and the `hap help skill` page.

## Tests

`skillbundle_test.go` — the embedded tree is compared against the real directory,
so a reference file added to the repo and not to the bundle fails; per-kind dir
mapping; both skills written; idempotence (no rewrite, mtime unchanged); a changed
file refreshed; partial-write reporting.
`internal/daemon/orchestrator_test.go` — skills on disk after a start; a **live**
session's stale skill refreshed (proves the placement before the alive branch);
the adoption case; and no `.claude`/`.agents` written into an operator's cwd.

## Gate

`go build`, `go test ./... -count=1 -p 2`, `gofmt`, `go vet`, `golangci-lint`
(`GOTOOLCHAIN=go1.26.2`) → **0 issues**, and `go test -tags integration
./test/integration/` — all pass.


internal/daemon/orchestrator.go
```diff
@@ -68,7 +69,7 @@ const orchestratorBrief = `You are hap's herd orchestrator on this machine. hap
 If the ` + "`hap`" + ` CLI is not on your PATH, use the binary at {self} in place of ` + "`hap`" + ` in every command below.
 
 Start by setting yourself up:
-1. Run ` + "`hap --skill`" + ` and ` + "`herdr --skill`" + ` and read both: they document the hap CLI (status, escalations, tasks, rules, config) and herdr (workspaces, panes, agents, reading and prompting an agent).
+1. Load the ` + "`hap-orchestrator`" + ` skill: hap has installed it, and the ` + "`hap`" + ` skill, into this working directory, so both are yours to re-read at any time — do that whenever your context is compacted and this brief is gone. ` + "`hap --skill`" + ` prints the hap one if the skill is somehow missing. Read ` + "`herdr --skill`" + ` too: it documents herdr (workspaces, panes, agents, reading and prompting an agent), which hap does not.
```
0xGosu: The brief asserts the skills are on disk unconditionally, but with `full_self_prompting.orchestrator_agent_cwd` set they are deliberately not written.

Scenario: an operator sets `orchestrator_agent_cwd`. `ensureOrchestrator` calls `bootstrapOrchestratorSkills`, which returns at the `cwd != ""` branch after logging once — nothing is installed anywhere. hap then *starts* the session there (this is not the adopted case, so it is briefed), and step 1 tells it "Load the `hap-orchestrator` skill: hap has installed it, and the `hap` skill, into this working directory". There is no such skill in that cwd, and the stated fallback only covers half of it: `hap --skill` prints `HapSkill` (skilldoc.go embeds only `.claude/skills/hap/SKILL.md`), so there is no route at all to the hap-orchestrator document — the very doc the compaction-recovery instruction points at.

`TestOrchestratorBriefShape` only asserts the brief *mentions* the skill, never that the claim holds for a cwd-set install, which is why this passes.

Worth either expanding `{self}`-style at send time (brief text chosen on whether the cwd is hap's), or having the brief name `hap skill install` as the route when the directory is the operator's.

skillbundle.go
```diff
@@ -0,0 +1,127 @@
...
+	base := filepath.Join(root, filepath.FromSlash(AgentSkillsDir(kind)))
+	var written []string
+	changed, err := syncSkillFile(filepath.Join(base, hapSkillDirName, SkillFileName), []byte(HapSkill))
+	if changed {
+		written = append(written, filepath.Join(base, hapSkillDirName, SkillFileName))
+	}
+	if err != nil {
+		return written, err
+	}
+	walkErr := fs.WalkDir(orchestratorSkillFS, orchestratorSkillRoot, func(p string, d fs.DirEntry, err error) error {
```
0xGosu, line 102: One failing file permanently truncates the rest of the tree, and after the first pass it is silent.

The callback returns the `syncSkillFile` error, so `fs.WalkDir` aborts — and `WalkDir` walks in lexical order, so the same file fails on every subsequent pass and every file after it never gets installed. Concretely, this PR's own `TestInstallAgentSkillsReportsPartialWritesOnFailure` sets up exactly that state (a plain file where `hap-orchestrator/` must go); in the daemon the equivalent (say `escalations.md` existing as a directory, or one file unwritable) leaves `incidents.md`, `reviewing-prs.md`, `task-sources.md` and `watchers.md` never installed, so SKILL.md links to references that are not there. And because `bootstrapOrchestratorSkills` notes the error once per distinct message (`skillsErrNoted == msg`), the second pass onwards logs nothing — the half-installed tree is invisible.

Since the goal is "install as much as possible, never fatal", accumulating per-file errors and continuing the walk (return `nil` from the callback after recording the error, join at the end) would install the other six references rather than stopping at the first.

PR #504, opened 2026-09-12
#0 feat: pause turso sync on demand, and migrate between engines

0xGosu:
Two related gaps in running the shared-database engine.

## `database.turso_sync_paused`

Takes a machine off Turso Cloud without unpicking its URL and token — for a metered network, a Turso incident, or a node you do not want publishing its herd right now. The engine stays `turso` and the local replica stays the store; only the pulls, the pushes and the shutdown push stop, and queued writes travel on the first push after you turn it off.

It is the **one `[database]` key the running daemon re-reads**, so `hap config set` applies it on a reload. A pause switch reachable only through `--restart` would cost the herd the in-flight captures and consults it exists to protect. `hap config set` suppresses the section's usual restart note for this key alone, and `hap help daemon` states the exception.

Four things the gate deliberately does *not* do:

- **The shutdown push is gated too** — it is the one cloud call outside the loop's ordinary path, so a paused node would otherwise reach Turso Cloud exactly once, at exit.
- **The pull tick still runs its local half** (`fleetPausedTick`). A sync database never checkpoints itself, so gating the whole pull path grows the replica's WAL for the length of the pause — an engine change, which is what this key promises not to be. Its checkpoint is gated on the WAL bound alone, never `fleetPull`'s pull counter: that counter does not advance while paused and starts at 0, so reusing it would checkpoint on *every* tick of a daemon that started paused.
- **`checkFleetSyncWedged` refuses while paused.** A pause freezes `lastError`, the failure count and both timestamps while the outage clock keeps running, so a node that happened to be failing when it was paused would meet every bound minutes later and restart itself — again every `fleetRecoveryRetryInterval`.
- **Lifting the pause pushes**, via the loop's own `fleetPushNow` nudge. Every write during the pause armed a debounce timer that fired into the gate and was not re-armed, so without it the queued rows wait for an unrelated write.

Reporting is honest: `daemonhealth.FleetSyncHealth.Paused` short-circuits `Degraded`, from which `IsolatedFor`, `Isolated`, `DiagLines` and both banners fall silent. It reads as PAUSED with the unpushed count, never as failing — reporting a deliberate pause as a fault is how an operator learns to ignore the banner that means a real outage. It is still *said*, because a herd off the wire looks exactly like a quiet one.

## `hap migrate --to <sqlite|turso>`

Switching `database.engine` pointed hap at a different database and left the history where it was; the only copier was one-way and once-only (`ImportLegacy`). This closes the door in both directions.

Built on the **existing importer** rather than a second copier — re-allocating ids in ascending old-id order, remapping every cross reference and omitting the in-flight rows is the hard part, and a second implementation would corrupt the database it writes into.

- **The two ends of node scope are asymmetric on purpose**: the source's node selects ("mine to take"), the destination's stamps ("whose they are now"). Filtering on the destination's selects nothing when the two differ; stamping the source's copies a history every operational query then refuses to show. It follows that `--all-nodes` *flattens* the fleet onto this machine rather than archiving it side by side.
- **Knowledge is never scoped** — signatures, embeddings, snapshots and decisions carry no `node_id` by design, so scoping them would silently downgrade what the destination knows.
- **A destination with no allocator still gets explicit ids**, counting from that table's `MAX(id)`: the ascending-order property is why the copy list is walked in old-id order, and SQLite advances `sqlite_sequence` so later inserts continue after the copied rows.
- **The re-run guard is an empty destination, not `legacy_imports`** — that table is keyed by node and records one origin, so gating on it would make the round trip this feature exists for impossible. Merging is not available at all: every id is re-allocated, so `INSERT OR IGNORE` cannot recognize a row it already wrote under a different one.

The command lives in `cmd/hap` because it is the only package that may open both engines — every other process reaches a turso store through the daemon's proxy, and the command's own precondition is that no daemon is running. It backs the destination up *before either handle is opened* (a copy from under an open handle loses whatever the WAL had not folded in) and takes the `-wal`/`-shm` sidecars with it. It honours `turso_sync_paused` by skipping its framing pull/push.

## Status

Verified locally on this machine:

- `go build -tags "vectors cpu" ./...` — links clean.
- `go test -tags "vectors cpu" ./internal/store/... -count=1` — green across all three `HAP_STORE_TEST_MODE` passes. The round-trip test runs in the sqlite and proxy passes and skips only the turso one (it is `skipUnlessSQLite`, which skips turso mode alone).
- `go test -tags "vectors cpu" ./... -count=1 -p 2` — green.
- `GOTOOLCHAIN=go1.26.2 golangci-lint run --build-tags "vectors,cpu"` — 0 issues.


internal/store/migrate.go
```diff
@@ -0,0 +1,196 @@
...
+
+// migrateNodeScoped names the tables in the copy list whose rows belong to ONE
+// node, and which a scoped migration therefore filters.
+//
+// It MIRRORS nodeScopedTables in nodescope_test.go, restricted to the tables
+// the copy list actually touches; TestMigrateScopeListsMatchTheGuard pins the
+// two together, because a table that gained a node_id elsewhere and was not
+// added here would come over whole — the fleet's rows landing in one machine's
+// file, silently.
+var migrateNodeScoped = map[string]bool{
```
0xGosu, line 183: **`task_lists` is missing from the copy list, so `hap migrate` silently drops every db-backed checklist.**

`nodeScopedTables` (nodescope_test.go) has `task_lists`; `migrateNodeScoped` does not, and neither does `importer.copySteps()`. `task_lists` is not in-flight bookkeeping — it is the whole content of the `sqlite` task-source provider (`internal/store/schema.go:202`, `internal/taskstore/dbtask`), i.e. the operator's actual checklists, one markdown blob per `(node_id, name)`.

Scenario: an install on the default `sqlite` task provider (which is what `config.Default()` names, and what a config-less install gets) runs `hap migrate --to sqlite` and then `hap config set database.engine sqlite`. Every `[[task_sources]]` entry still points at `db://<node>/<name>`, but `Store.ReadTaskList` (`internal/store/tasklists.go:29`) now finds no row and returns an `fs.ErrNotExist` wrap — or a create-on-demand path writes a fresh empty list over the top. The migration reported success and the task lists are gone. The same happens on the way up.

`TestMigrateScopeListsMatchTheGuard` cannot catch this: it only iterates tables the copy list already touches (`for table := range copied`), plus the reverse direction (entries naming an untouched table). A node-scoped table absent from *both* lists is invisible to it.

The help text's omission list — "in-flight rows — pending LLM requests and decisions, queued agent actions, and the roster" — does not cover this either, so an operator has no warning.

cmd/hap/migrate.go
```diff
@@ -0,0 +1,283 @@
...
+	if err != nil {
+		return err
+	}
+	tursoStore, err := store.OpenDB(tdb.DB(), store.Options{
+		NodeID: nodeID, Engine: store.EngineTurso,
+		IDs: store.NewTimeOrderedIDs(store.NodeBits(nodeID), nil),
+		// Migrate: false, then PrepareSharedSchema — exactly as the daemon
+		// does. DDL on a shared database is issued only by the schema lease
+		// holder; two nodes running the same ALTER wedge the loser silently,
+		// and a migration is precisely the moment a second machine might be
```
0xGosu, line 103: **`--to turso` writes a whole history into the shared database without the node-bits collision check the daemon refuses to start without.**

The daemon does this immediately after `PrepareSharedSchema` and before anything is written (`cmd/hap/main.go`):

```go
if other, clash, err := st.NodeBitsCollision(ctx); err == nil && clash {
    return fmt.Errorf("turso: this node's id bits collide with node %s (%s) ... start again on
        whichever of the two machines has never written to the shared store", ...)
}
```

Its own comment says why: "Ids carry 12 bits of the node id; two nodes sharing them would share an id space. Refuse now, **before this node writes anything**, rather than collide later."

`runMigrate` opens the same store with the same `NewTimeOrderedIDs(NodeBits(nodeID), nil)` allocator and then copies the machine's entire audit log, decisions, corrections and agent actions into the shared database — skipping that check. Concrete failure: an operator with a colliding node id runs `hap migrate --to turso`, it reports success, and the first `hap daemon --ensure` afterwards refuses to start with the collision error — whose stated remedy ("start again on whichever machine has never written to the shared store") is now impossible, because the migration just made this machine a writer. Colliding ids are already in the fleet's rows.

Ask `NodeBitsCollision` right after `PrepareSharedSchema` and refuse, exactly as the daemon does.

cmd/hap/migrate.go
```diff
@@ -0,0 +1,283 @@
...
+	} else {
+		mo.Src, mo.Dst = sqliteStore, tursoStore
+		mo.SourceLabel = paths.DBPath()
+	}
+	rep, err := store.Migrate(ctx, mo)
+	if err != nil {
+		if errors.Is(err, store.ErrDestinationNotEmpty) {
+			return fmt.Errorf("%w.\nThe copy re-allocates every id, so running it twice would duplicate every row\n"+
+				"rather than merging. Start from an empty destination, or pass --force if you\n"+
+				"have decided the duplication is acceptable (the backup above is the way back)", err)
```
0xGosu: **"the backup above is the way back" is false for `--to turso`, and that is the direction where `--force` does the most damage.**

The backup is taken of `dstPath`, which for `--to turso` is `paths.TursoDBPath()` — this machine's *local replica*. But the destination that matters is Turso Cloud: after the copy, `tdb.Push()` publishes the duplicated rows, and every other node folds them in on its next pull.

Scenario: an operator re-runs `hap migrate --to turso`, hits `ErrDestinationNotEmpty`, reads this message as "`--force` is recoverable", and passes it. Every audit row, decision, correction and kill event is now duplicated in the shared database and on every peer. Restoring `hap-turso.db.pre-migrate-<stamp>.bak` undoes nothing — the rows are already remote, and the next pull brings them straight back.

Either restrict the "way back" wording to `--to sqlite`, or say plainly for `--to turso` that a forced copy is published to the fleet and cannot be rolled back from a local backup.

cmd/hap/migrate.go
```diff
@@ -0,0 +1,283 @@
...
+		// holder; two nodes running the same ALTER wedge the loser silently,
+		// and a migration is precisely the moment a second machine might be
+		// starting up.
+		Migrate: false, AgentLockDir: filepath.Join(paths.StateDir, "agent-automation-locks"),
+	})
+	if err != nil {
+		return err
+	}
+	defer tursoStore.Close()
+	if err := turso.PrepareSharedSchema(ctx, tdb, tursoStore, time.Now); err != nil {
```
0xGosu, line 111: **The command does not actually honour `turso_sync_paused` here.** `turso.PrepareSharedSchema` opens with an unconditional `db.Pull()` (`internal/store/turso/schema.go:70`) and, when this node takes the schema lease, ends in a push.

So a paused node running `hap migrate` still contacts Turso Cloud — contradicting this file's own comment ("Deliberately honours the pause ... it just copies whatever the local replica holds") and CLAUDE.md ("it honours `database.turso_sync_paused` by skipping the framing pull/push"), and the report's closing line ("nothing was pulled or pushed"), which is printed while a pull definitely happened.

Not necessarily wrong to pull for the schema check — but either the schema step should be pause-aware too, or all three claims should be narrowed to "the framing pull/push" excluding the schema handshake, so an operator on a metered link is not told nothing went over the wire.

internal/frontend/daemonhealth.go
```diff
@@ -163,6 +175,7 @@ func (a *App) AssessDaemonHealth() DaemonHealth {
 			h.BinaryReplaced = rec.BinaryReplaced
 			h.FleetSyncLine = rec.FleetSync.Line(now)
 			h.FleetSyncDegraded = rec.FleetSync.Degraded()
+			h.FleetSyncPaused = rec.FleetSync != nil && rec.FleetSync.Paused
```
0xGosu, line 178: **`FleetSyncPaused` never reaches the TUI, so there the pause is silent rather than "said".**

This field is set here and read in exactly one place: `internal/cli/cli.go:1057` (`hap status`). The TUI has no reader for it, and it never renders `FleetSyncLine` at all — its only fleet-sync output is gated on `FleetSyncDegraded` (`internal/tui/tui.go:7375`), which this PR makes `false` whenever `Paused` is true, and `Banner()` is suppressed by the same `Degraded()` short-circuit.

So against the changelog's "Fixed `hap status` **and the TUI** ... it now reads as PAUSED", the TUI actually goes from showing the sync diag lines to showing nothing — which is the exact failure mode the design note warns about ("an operator who has forgotten the key set weeks ago would otherwise read a herd that looks quiet as a quiet herd"), just moved into the surface most operators watch.

Adding a `FleetSyncPaused` line beside the existing banner block in `tui.go` would close it.

internal/store/migrate.go
0xGosu: Fixed in f37727d. `task_lists` is now in `importer.copySteps()` (stamped with the destination's node like every node-owned row, so `db://<node>/<name>` locators still resolve on the machine that owns the config) and in `migrateNodeScoped`. Because `ImportLegacy` shares the copy list, the automatic sqlite→turso import no longer drops these lists either.

`TestMigrateScopeListsMatchTheGuard` now iterates `nodeScopedTables` and `explicitIDTables` AND every table in a freshly opened schema (`sqlite_master`): each must be copied or named with a reason in a new `migrateNotCopied` map, which is itself checked against the schema. `TestMigrateRoundTripPreservesRowsAndReferences` seeds a task list and reads it back through `ReadTaskList` after sqlite → shared → sqlite. Mutation check: deleting the copy step fails both tests. `go test ./internal/store/...` passed (sqlite mode only — proxy/turso modes not run).

**Not yet verified:** the full `go test -tags "vectors cpu" ./...` run and `golangci-lint` have NOT been run on these commits — the build slot on this machine is held by another job, and they will run once it frees up. What did run: the targeted package tests named above, `gofmt` and `go vet` on the touched packages.

Accepted limitation: under `--all-nodes` two nodes' lists with the same name collide and the second is dropped, like agent names.

cmd/hap/migrate.go
0xGosu: Fixed in 5d29394. `runMigrate` now calls `refuseNodeBitsCollision` right after `PrepareSharedSchema`, before anything is written. Two deliberate differences from the daemon:
- It asks only for `--to turso`. Going to sqlite nothing is written to the shared database, and refusing there would block the colliding node's one route home. (A read error passes, as in the daemon.)
- The remedy is not the daemon's verbatim. Moving THIS machine's node-id file aside and re-running would copy nothing: the local file's rows stay stamped with the old id and the copy is scoped to `Src.self`, so it would report success over an empty history. The message says so and points at regenerating the id on a machine with no history to lose.

`TestMigrateToTursoRefusesACollidingNodeID` seeds a twin node and asserts the refusal going up and no refusal going down; mutation-checked both halves. Note the check reads the replica refreshed by `PrepareSharedSchema`'s pull, which is one reason the fix for the `turso_sync_paused` thread keeps that pull.

**Not yet verified:** the full `go test -tags "vectors cpu" ./...` run and `golangci-lint` have NOT been run on these commits — the build slot on this machine is held by another job, and they will run once it frees up. What did run: the targeted package tests named above, `gofmt` and `go vet` on the touched packages.

cmd/hap/migrate.go
0xGosu: Fixed in d1120a9, restricting the way-back wording by direction (`destinationNotEmptyError`). `--to sqlite` still names the backup as the way back. `--to turso` now says plainly there is NO way back: a forced copy is published to Turso Cloud (immediately, or on the first push after `turso_sync_paused` is lifted), every other node pulls the duplicates in, and the backup is only this machine's local replica, so restoring it undoes none of that. The pre-copy "backed up" line going up now says it is the local replica and that Turso Cloud is not backed up, and `hap help migrate` (the `--force` flag and the details) carries the same warning. `TestDestinationNotEmptyErrorOnlyPromisesAWayBackGoingToSQLite` covers both directions and that the sentinel survives. Changelog fragment updated.

**Not yet verified:** the full `go test -tags "vectors cpu" ./...` run and `golangci-lint` have NOT been run on these commits — the build slot on this machine is held by another job, and they will run once it frees up. What did run: the targeted package tests named above, `gofmt` and `go vet` on the touched packages.

cmd/hap/migrate.go
0xGosu: Addressed in e92a7e9 by narrowing the claims rather than making the schema step pause-aware, for two reasons: that pull is what `refuseNodeBitsCollision` (the node-bits collision thread) reads peers from, so skipping it under a pause would let a paused node pass that check against a stale replica; and DDL on a shared database needs the lease, which needs a push, so a pause-aware schema step could only refuse.

All three claims are corrected: the comment in `runMigrate` now says the pause covers only the copy's framing pull/push, and that `PrepareSharedSchema` still pulls (and pushes when it leads a schema migration) and a missing replica bootstraps from the cloud. CLAUDE.md is narrowed the same way. The report's paused note (`pausedMigrateNote`) names the round trip that was skipped for the direction taken and says the command still reached Turso Cloud for its schema check. `TestPausedMigrateNoteDoesNotClaimSilenceOnTheWire` covers both directions.

**Not yet verified:** the full `go test -tags "vectors cpu" ./...` run and `golangci-lint` have NOT been run on these commits — the build slot on this machine is held by another job, and they will run once it frees up. What did run: the targeted package tests named above, `gofmt` and `go vet` on the touched packages.

internal/frontend/daemonhealth.go
0xGosu: Fixed in c115ffe. The TUI now renders a line beside the health banner block whenever `FleetSyncPaused` is set, independent of `FleetSyncDegraded`/`Banner()`:

> fleet sync PAUSED — other nodes' escalations and agents are not shown here, nor this node's there · resume: hap config set database.turso_sync_paused false

It uses the `paused` style rather than `warn`, since the operator asked for this. `TestFleetSyncPausedIsStated` asserts the line and the resume key at 100 columns with no warning glyph, with an unpaused health as the control. This makes the changelog's "hap status and the TUI read as PAUSED" claim true, so the fragment was left as is.

**Not yet verified:** the full `go test -tags "vectors cpu" ./...` run and `golangci-lint` have NOT been run on these commits — the build slot on this machine is held by another job, and they will run once it frees up. What did run: the targeted package tests named above, `gofmt` and `go vet` on the touched packages.

internal/store/migrate.go
0xGosu: **Verification update** (supersedes the "not yet verified" note above), on branch head c115ffe:
- `go test -tags 'vectors cpu' ./... -count=1 -p 2`: every package passed except one test in `internal/daemon`, `TestTaskReviewNoopOnlyForAnExhaustedSource/legal_on_a_genuinely_exhausted_source` ("want a noop audit row, got []"). None of these commits touches `internal/daemon` or anything on that test's path (the copy-list change runs only in `hap migrate` and the legacy import). Run alone, it passed 3 out of 3 times. This package is known to fail one random test under full-suite load, but I have not proven that is the cause here beyond those re-runs.
- `GOTOOLCHAIN=go1.26.2 golangci-lint run --build-tags 'vectors,cpu' --new-from-rev=0a26231`: 0 issues.
- Not run: the store suite's `proxy`/`turso` modes against a real database, and the `integration`-tagged suite.

cmd/hap/migrate.go
0xGosu: **Verification update** (supersedes the "not yet verified" note above), on branch head c115ffe:
- `go test -tags 'vectors cpu' ./... -count=1 -p 2`: every package passed except one test in `internal/daemon`, `TestTaskReviewNoopOnlyForAnExhaustedSource/legal_on_a_genuinely_exhausted_source` ("want a noop audit row, got []"). None of these commits touches `internal/daemon` or anything on that test's path (the copy-list change runs only in `hap migrate` and the legacy import). Run alone, it passed 3 out of 3 times. This package is known to fail one random test under full-suite load, but I have not proven that is the cause here beyond those re-runs.
- `GOTOOLCHAIN=go1.26.2 golangci-lint run --build-tags 'vectors,cpu' --new-from-rev=0a26231`: 0 issues.
- Not run: the store suite's `proxy`/`turso` modes against a real database, and the `integration`-tagged suite.

cmd/hap/migrate.go
0xGosu: **Verification update** (supersedes the "not yet verified" note above), on branch head c115ffe:
- `go test -tags 'vectors cpu' ./... -count=1 -p 2`: every package passed except one test in `internal/daemon`, `TestTaskReviewNoopOnlyForAnExhaustedSource/legal_on_a_genuinely_exhausted_source` ("want a noop audit row, got []"). None of these commits touches `internal/daemon` or anything on that test's path (the copy-list change runs only in `hap migrate` and the legacy import). Run alone, it passed 3 out of 3 times. This package is known to fail one random test under full-suite load, but I have not proven that is the cause here beyond those re-runs.
- `GOTOOLCHAIN=go1.26.2 golangci-lint run --build-tags 'vectors,cpu' --new-from-rev=0a26231`: 0 issues.
- Not run: the store suite's `proxy`/`turso` modes against a real database, and the `integration`-tagged suite.

cmd/hap/migrate.go
0xGosu: **Verification update** (supersedes the "not yet verified" note above), on branch head c115ffe:
- `go test -tags 'vectors cpu' ./... -count=1 -p 2`: every package passed except one test in `internal/daemon`, `TestTaskReviewNoopOnlyForAnExhaustedSource/legal_on_a_genuinely_exhausted_source` ("want a noop audit row, got []"). None of these commits touches `internal/daemon` or anything on that test's path (the copy-list change runs only in `hap migrate` and the legacy import). Run alone, it passed 3 out of 3 times. This package is known to fail one random test under full-suite load, but I have not proven that is the cause here beyond those re-runs.
- `GOTOOLCHAIN=go1.26.2 golangci-lint run --build-tags 'vectors,cpu' --new-from-rev=0a26231`: 0 issues.
- Not run: the store suite's `proxy`/`turso` modes against a real database, and the `integration`-tagged suite.

internal/frontend/daemonhealth.go
0xGosu: **Verification update** (supersedes the "not yet verified" note above), on branch head c115ffe:
- `go test -tags 'vectors cpu' ./... -count=1 -p 2`: every package passed except one test in `internal/daemon`, `TestTaskReviewNoopOnlyForAnExhaustedSource/legal_on_a_genuinely_exhausted_source` ("want a noop audit row, got []"). None of these commits touches `internal/daemon` or anything on that test's path (the copy-list change runs only in `hap migrate` and the legacy import). Run alone, it passed 3 out of 3 times. This package is known to fail one random test under full-suite load, but I have not proven that is the cause here beyond those re-runs.
- `GOTOOLCHAIN=go1.26.2 golangci-lint run --build-tags 'vectors,cpu' --new-from-rev=0a26231`: 0 issues.
- Not run: the store suite's `proxy`/`turso` modes against a real database, and the `integration`-tagged suite.

PR #505, opened 2026-09-12
#0 feat: filter the orchestrator's own events out of hap stream, and fix the agy preset model

0xGosu:
Three changes, one branch.

## `hap stream orchestrator` no longer echoes the reader's own actions

The reader of that stream is normally the orchestrator, and its own actions sat in the log beside everyone else's — so every task it marked done, every escalation it answered and every config key it set came straight back at it as an event it had to recognize and ignore. Events whose author is `domain.OrchestratorAuthor` are now suppressed; `--include-self` prints them for debugging an emitter.

**The sequence accounting is the load-bearing part.** The cursor advances over a suppressed event exactly as over a printed one, so they are never re-delivered on `--resume` and never read as a gap. Everything else in the loop stays on the **raw** batch — `prunedUnder`'s `batch[0].Seq`, `settled`, `lastEventAt`, the full-batch continue — because each answers a question about the log rather than about what was printed. A `printed`-based `settled` would re-query the floor forever on an orchestrator-only log; a `printed`-based batch check would end the no-wait replay mid-stream. The `head`/`floor` banner and the gap and reset notices are unchanged.

Two new tests, and the second is the one that carries the weight:

- `TestStreamResumeSpansARunOfSuppressedEvents` — `--resume 1` over three consecutive self-authored events prints exactly the one visible line, with the banner still counting the suppressed ones and no gap or reset notice.
- `TestStreamNewestEventSuppressedStillAdvancesTheCursor` — when the newest event is the reader's own, nothing is printed, so the output cannot witness the cursor at all; a recording log watches it cross that seq, then retention removes everything beneath it and a visible event arrives. A cursor left behind reports `# gap missed=2..3` for the reader's own work.

Mutation-checked: rewriting the loop to `continue` past suppressed events (skipping the cursor advance) fails the newest-suppressed test. The resume-spanning test passes under that mutation, because a visible event in the same batch drags the cursor forward anyway — which is exactly why the newest-suppressed case needs to exist separately.

## The agy preset moves to `gemini-3.8-flash-high`

`task_generate` and `learn_from_user` named `gemini-3.1-pro-high`, which is the older and weaker model despite the Pro name. The re-ranking judge stays on `gemini-3.8-flash-low` — a deliberate cheap tier, since an agent is parked for that whole run.

## `LLMPresetAgy`'s doc comment gave two false reasons

Both verified live on 2026-09-12:

- hap puts `HAP_REQUEST_ID`, `HAP_DB_PATH` and `HAP_CONTROL_PATH` into the child process environment (`internal/llm/llm.go`), and an MCP server agy spawns inherits them — so an `agy mcp add` entry needs no per-request `env` at all.
- agy *does* have an `--allowedTools` equivalent: `permissions.allow` entries spelled `mcp(server/tool)` in agy's `settings.json`. Granted only `mcp(hap/get_context)` and `mcp(hap/submit_decision)`, the consult ran while shell and file tools stayed denied.

The conclusion is unchanged — there is still no agy preset for `llm.command` — but the reason is now the true one: a preset writes argv only, both mechanisms are machine-wide files no picker keystroke can create, and creating them would expose hap's tools to every agy session on that machine. The comment's "fails opaquely with nothing staged" clause is gone, since it re-asserted the error being removed. The same wrong claim is corrected in `sample/config.toml`, and a stale "fails opaquely" note in `llmpresetsample_test.go`.

## Verification

- `go test -tags 'vectors cpu' ./... -count=1 -p 2` — exit 0, 42 packages, no failures (`internal/daemon` 140s, `internal/frontend` 13s, `internal/cli` 39s).
- `GOTOOLCHAIN=go1.26.2 golangci-lint run --build-tags "vectors,cpu"` — 0 issues.
- `gofmt -l .` and `go vet -tags "vectors cpu" ./...` clean.

Docs updated alongside: `hap help stream orchestrator`, `README.md`, `.claude/skills/hap/SKILL.md`, and the Monitor bullet in `.claude/skills/hap-orchestrator/SKILL.md` (that skill is the direct consumer and would otherwise have its own actions silently missing). Patch-level work, so `herdr-plugin.toml` is deliberately not bumped.


internal/cli/stream.go
```diff
@@ -125,9 +136,22 @@ func streamOrchestrator(ctx context.Context, app *frontend.App, out io.Writer, a
...
+		// The cursor advances over a SUPPRESSED event exactly as it does over a
+		// printed one, so a self-authored event is never re-delivered on the
+		// next resume and never reads as a gap. Everything else in this loop is
+		// accounted on the RAW batch — prunedUnder's batch[0].Seq, settled,
+		// lastEventAt and the full-batch continue above — because each answers a
+		// question about the LOG, not about what was printed: counting only
+		// printed events would re-query the floor forever on a log the
+		// orchestrator alone wrote, and would end the no-wait replay mid-stream
+		// on a full batch it happened to author.
 		for _, ev := range evs {
```
0xGosu, line 167: **Suppressed-only stream never notices its reader is gone.**

Moving the `Fprintln` inside the filter removes the only signal that the consumer went away — and the case where it matters is precisely the case where nothing is printed.

Concrete scenario: the Monitor tool dies while the orchestrator agent keeps running (the case the brief itself anticipates: *“if the monitor stops, restart it with `--resume`”*). The agent keeps authoring events; the orphaned `hap stream orchestrator` reads them, suppresses every one, never calls `Fprintln`, never sees `EPIPE` — and never gets `SIGPIPE`, which only fires on an actual write. It polls SQLite every 500 ms forever, and every Monitor restart leaks another one. Before this change the first event printed would have reaped it.

The comment (“a suppressed event is no evidence either way”) covers the narrow claim, but not the accumulation. A cheap fix is to probe the writer on a suppressed event too — e.g. `fmt.Fprint(out, "")` is a no-op for a buffer but still returns the pipe error — or to write nothing and periodically check `out` liveness some other way.

0xGosu:
**`internal/daemon/orchestrator.go:82` — the in-binary brief was not updated for the suppression** (not in the diff, so noting it here).

`orchestratorBrief` is what a *daemon-started* orchestrator actually receives, and step 3 still reads:

> Start the Monitor tool on `hap stream orchestrator`. It prints a `# … head=N` line, then one line per event: `<seq> <time> <kind> key=value … by=<author>`. Remember the last seq you handled …

README, `.claude/skills/hap/SKILL.md`, `.claude/skills/hap-orchestrator/SKILL.md` and `hap help stream orchestrator` all gained the “your own events are suppressed” note; this one did not. The PR body gives the reason for updating the orchestrator skill — “that skill is the direct consumer and would otherwise have its own actions silently missing” — and it applies with at least as much force here.

Scenario: a fresh orchestrator runs `hap confirm 42 --send`, watches the stream for the `correction`/`escalation.dismissed` line to confirm it landed, and never sees one. The `{skills}` step mitigates it when the skill is installed, but the brief is the only instruction that is always delivered.

internal/cli/stream.go
0xGosu: Fixed in 085d7a7. The stream now counts the events it suppresses since its last write, and once that count is non-zero and `streamProbeAfter` (10s) has passed since the last write, it writes one `# suppressed N self-authored event(s) through seq=S (--include-self shows them)` notice; a failed write ends the stream exactly as a failed event line does. The check runs on every pass — idle ticks after the last suppressed event and the full-batch replay path included — and any printed event clears the count, so a burst of the orchestrator's own work costs one line, not one per poll.

Tests: `TestStreamNoticesAVanishedReaderAcrossSuppressedEvents` closes the writer after the banner, appends only self-authored events, and requires the stream to exit; its control asserts no event line was even attempted and the notice was, so the exit came from the probe. `TestStreamProbeIsOneLinePerBurst` pins exactly one notice per burst for a live reader. Both fail (stream never exits) with the probe disabled.

Scope note, also in the code comment: a genuinely idle stream still cannot notice a vanished reader — that predates this PR, and closing it would mean a heartbeat on every quiet stream. The changelog fragment mentions the notice.

0xGosu:
Fixed in 4263808. Step 3 of `orchestratorBrief` now adds one sentence: events you authored are left out (`--include-self` shows them), and a `# suppressed` line (the reader-liveness notice added for the other finding) only notes that some were — ignore it. `TestOrchestratorBriefShape` now requires both `--include-self` and `# suppressed`, so the brief cannot drift from the binary silently again. The `hap` and `hap-orchestrator` skills (both embedded in the binary) already described the suppression; they now mention the `# suppressed` line too.

PR #512, opened 2026-09-18
#0 feat: show TUI loading progress and parallelize its reads

0xGosu:
## What

The TUI now shows loading progress, and it and the status verbs load much faster over a remote store.

- **Loading screen:** until the first refresh lands, the body reads `loading… 44%  ███████████░░░░░░░░░░░░░  4/9 · learned rules`, and after a second it adds how long the load has taken so far. It used to show `no agents detected`, which over a far server looked like the herd was gone. After that, a refresh that runs past 0.75 s shows `↻ N%` in the header.
- **Reads run concurrently:** `GetStatus` (~20 independent queries), `fillFleet` (now `readFleet` plus an assembly step), the rule listing, and the TUI's `refreshDataProgress` issue independent reads together (`frontend.Concurrently`). The duplicated `CountSignatureEmbeddings` call is gone.
- **One poll in flight:** a 2 s tick used to start a full refresh even while the previous one was still loading. Over a remote server that queued ~6 refreshes behind the proxy's 2-connection pool. It is now latched (`pollStarted`), with a 2-minute backstop.
- **libsql only:** the TUI's proxy pool is 4, and the store socket serves 16 sessions (`libsqlMaxClients`) so the TUI doesn't crowd out CLI verbs. A session over the cap is refused, not queued. turso keeps 8, because its pre-warmed pool costs memory per connection.

Progress is **pulled** (`loadTracker`, repainted by the existing 1 s clock tick) rather than pushed as messages, so a refresh stays exactly one command yielding one `refreshMsg`. A failed refresh still carries only the reads that come before the failing one in the old sequential order, because the TUI replaces its data wholesale.

## Measured (live, libsql engine → Layerbase, ~230 ms/round trip)

| | Before | After |
|---|---|---|
| TUI first herd paint | 10.1 s | 2–4 s |
| `hap status` | 4.2 s | 2.7 s |
| `hap agents` | 4.2 s | 2.1 s |

## Testing

- New tests: progress reaches exactly 100% (`refreshSteps` can't drift from the reads), the loading screen shows before the first refresh and never after it, the header indicator appears only past `slowLoadAfter`, and ticks don't stack polls, with the backstop covered.
- The frontend, tui, cli, cmd/hap and store suites pass. frontend and tui also pass under `-race`, and `golangci-lint` is clean with the pinned toolchain.


0xGosu:
@CharlieHelps please review

PR #514, opened 2026-09-18
#0 feat: let the TUI clear a configured LLM command so a preset can be re-picked

0xGosu:
## Summary
- `x` on the TUI Config tab now clears a configured `llm.command`, `llm.task_generate_command`, `llm.learn_from_user_command`, `llm.reranking_command` or `full_self_prompting.orchestrator_agent_command` (after a y/n confirm). The row returns to `(disabled)`, so `e`/enter re-opens the preset picker.
- Editing a configured template in the TUI stays forbidden (CR-036) — clearing types nothing, so the mangling hazard doesn't apply.
- `frontend.ClearLLMCommand(ctx, key, expected)` refuses unless config.toml still holds exactly the template the operator was shown; the TUI confirm also revalidates against the latest refresh.
- CLI parity already exists: `hap config set <key> ""` clears (SplitCommand("") → nil).
- README + hap skill note, changelog fragment.

## Tests
- New: clear→picker round trip for all 5 keys, cancel keeps template, changed-template refusal, `x` on an ordinary field still refused.
- `go test -tags "vectors cpu" ./internal/tui ./internal/frontend` green; golangci-lint 0 issues.
- Local env lacks FAISS headers, so the `vectors` build of daemon/match/cmd couldn't run locally — relying on CI.

https://claude.ai/code/session_01AdwfkMi8mRgkef6xjfd31u


0xGosu:
@CharlieHelps review

